PR #8, opened 2018-02-14
Create community README page

ewilderj:
Includes details of our regular meetings, link to meeting calendar, and an overview of current discussion forums.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,3 +43,8 @@ race, religion, or sexual identity and orientation.
 The Kubeflow community is guided by our [Code of
 Conduct](https://github.com/kubeflow/community/blob/master/CODE_OF_CONDUCT.md),
 which we encourage everybody to read before participating.
+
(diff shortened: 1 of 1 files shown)


jlewi:
Maybe add the "Who should get involved section" from
https://github.com/kubeflow/kubeflow#who-should-consider-contributing-to-kubeflow

Then we can probably get rid of that info on kubeflow/kubeflow and just link to kubeflow/community.

ewilderj:
I don't think the call to involvement needs to be removed from the main repo: for a lot of people it will be the first thing they see, and it helps to have the "you should contribute" in the main README. It's at the bottom, and I don't think that document is concerningly long or anything.

I see the community repo purpose as "great! you want to be involved, here's all our resources that will help", vs. the call to action.

I do think linking to the community repo from the main README makes sense though, I can add that link in once this is merged.

PR #30, opened 2018-03-05
Add tf-operator design doc for API v1alpha2

ScorpioCPH:
Hi folks,

This is a proposal for tf-operator API `v1alpha2`, PTAL, thanks!

List of issues related to the API
  kubeflow/tf-operator#64
  kubeflow/tf-operator#223

/cc @jlewi @gaocegege @DjangoPeng @ddysher

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/30)

### DIFF
--- a/proposals/tf-operator-design-v1alpha2.md
+++ b/proposals/tf-operator-design-v1alpha2.md
@@ -1,0 +1,440 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [TF-Operator Design (v1alpha2)](#tf-operator-design-v1alpha2)
+  - [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [UI or API](#ui-or-api)
+  - [Design](#design)
+    - [TFController](#tfcontroller)
+    - [Distributed TensorFlow Configuration](#distributed-tensorflow-configuration)
+    - [Event-Driven](#event-driven)
+    - [Reconciler](#reconciler)
+    - [Error Handling](#error-handling)
+    - [Test](#test)
+  - [Alternatives Considered](#alternatives-considered)
+    - [Future Works](#future-works)
+    - [Related Issues](#related-issues)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+_Authors:_
+
+* @ScorpioCPH - Penghao Cen &lt;cenph@caicloud.io&gt;
+
+# TF-Operator Design (v1alpha2)
+
+## Motivation
+
+Kubeflow community currently have [tf-operator](https://github.com/kubeflow/tf-operator) (v1alpha1) for running TensorFlow jobs on Kubernetes. And we have received some refactoring [requests](https://github.com/kubeflow/tf-operator/issues?q=is%3Aissue+is%3Aopen+label%3Akind%2Fapi-change) for API changes.
+Open this file to summarize the design details and move the version of API to `v1alpha2`.
+
+## Goals
+
+- Define the structure of API `v1alpha2`.
+    + Cover most of the refactoring requests we have discussed.
+    + Simplify the API definition.
+- Define an `event-driven` mechanism for TFJob life-cycle management.
+    + And use `reconciler` mechanism as a double check.
+- Clarify the `error handing` logic.
+- Provide a `test` mechanism to verify the design and implementation.
+
+## Non-Goals
+
+- **Notes:** As we make a big change in API v1alpha2, compatibility with v1alpha1 is **NOT** be taken into consideration in this proposal.
+
+## UI or API
+
+The `TFJob` API v1alpha2 object will have the following structure:
+
+**TFJob**:
+```go
+// TFJob represents the configuration of signal TFJob
+type TFJob struct {
+    metav1.TypeMeta `json:",inline"`
+
+    // Standard object's metadata.
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    // Specification of the desired behavior of the TFJob.
+    Spec TFJobSpec `json:"spec,omitempty"`
+
+    // Most recently observed status of the TFJob.
+    // This data may not be up to date.
+    // Populated by the system.
+    // Read-only.
+    Status TFJobStatus `json:"status,omitempty"`
+}
+```
+
+**TFJobSpec**:
+```go
+// TFJobSpec is a desired state description of the TFJob.
+type TFJobSpec struct {
+    // TFReplicaSpecs is map of TFReplicaType and TFReplicaSpec
+    // specifies the TF replicas to run.
+    // For example,
+    //   {
+    //     "PS": TFReplicaSpec,
+    //     "Worker": TFReplicaSpec,
+    //   }
+    TFReplicaSpecs map[TFReplicaType]*TFReplicaSpec `json:"tfReplicaSpecs"`
+}
+```
+
+**TFReplicaSpec**:
+```go
+// TFReplicaSpec is a description of the TFReplica
+type TFReplicaSpec struct {
+    // Replicas is the desired number of replicas of the given template.
+    // If unspecified, defaults to 1.
+    Replicas *int32 `json:"replicas,omitempty"`
+
+    // Template is the object that describes the pod that
+    // will be created for this TFReplica.
+    // We use RestartPolicy in PodTemplateSpec
+    // to describe how the containers within the pod should be restarted.
+    // Please set this restart policy carefully according to your code.
+    Template *v1.PodTemplateSpec `json:"template,omitempty"`
+
+    // Restart policy for all TFReplicas within the TFJob.
+    // One of Always, OnFailure, Never and ExitCode.
+    // Default to Always.
+    RestartPolicy RestartPolicy `json:"restartPolicy,omitempty"`
+}
+
+// RestartPolicy describes how the TFReplicas should be restarted.
+// Only one of the following restart policies may be specified.
+// If none of the following policies is specified, the default one
+// is RestartPolicyAlways.
+type RestartPolicy string
+
+const (
+    RestartPolicyAlways    RestartPolicy = "Always"
+    RestartPolicyOnFailure RestartPolicy = "OnFailure"
+    RestartPolicyNever     RestartPolicy = "Never"
+    RestartPolicyExitCode  RestartPolicy = "ExitCode"
+)
+```
+
+**TFReplicaType**:
+```go
+// TFReplicaType is the type for TFReplica.
+type TFReplicaType string
+
+const (
+    // TFReplicaTypePS is the type for parameter servers of distributed TensorFlow.
+    TFReplicaTypePS TFReplicaType = "PS"
+
+    // TFReplicaTypeWorker is the type for workers of distributed TensorFlow.
+    TFReplicaTypeWorker TFReplicaType = "Worker"
+
+    // TFReplicaTypeChief is the type for chief worker of distributed TensorFlow.
+    // If there is "chief" replica type, it's the "chief worker".
+    // Else, worker:0 is the chief worker.
+    TFReplicaTypeChief TFReplicaType = "Chief"
+
+    // TFReplicaTypeEval is the type for evaluation replica in TensorFlow.
+    TFReplicaTypeEval TFReplicaType = "Eval"
+)
+```
+
+**TFJobStatus**:
+```go
+// TFJobStatus represents the current observed state of the TFJob.
+type TFJobStatus struct {
+    // TFReplicaStatuses is map of TFReplicaType and TFReplicaStatus,
+    // specifies the status of each TFReplica.
+    TFReplicaStatuses map[TFReplicaType]*TFReplicaStatus `json:"tfReplicaStatuses"`
+
+    // Represents time when the TFJob was acknowledged by the TFJob controller.
+    // It is not guaranteed to be set in happens-before order across separate operations.
+    // It is represented in RFC3339 form and is in UTC.
+    StartTime *metav1.Time `json:"startTime,omitempty"`
+
+    // Represents time when the TFJob was completed. It is not guaranteed to
+    // be set in happens-before order across separate operations.
+    // It is represented in RFC3339 form and is in UTC.
+    CompletionTime *metav1.Time `json:"completionTime,omitempty"`
+
+    // Represents last time when the TFJob was reconciled. It is not guaranteed to
+    // be set in happens-before order across separate operations.
+    // It is represented in RFC3339 form and is in UTC.
+    LastReconcileTime *metav1.Time `json:"lastReconcileTime,omitempty"`
+
+    // Represents is an array of current observed TFJob conditions.
+    Conditions []TFJobCondition `json:"conditions"`
+}
+```
+
+**TFReplicaStatus**:
+```go
+// TFReplicaStatus represents the current observed state of the TFReplica.
+type TFReplicaStatus struct {
+    // The number of actively running pods.
+    Active int32 `json:"active,omitempty""`
+
+    // The number of pods which reached phase Succeeded.
+    Succeeded int32 `json:"succeeded,omitempty"`
+
+    // The number of pods which reached phase Failed.
+    Failed int32 `json:"failed,omitempty"`
+}
+```
+
+**TFJobCondition**:
+```go
+// TFJobCondition describes the state of the TFJob at a certain point.
+type TFJobCondition struct {
+    // Type of TFJob condition.
+    Type TFJobConditionType `json:"type"`
+
+    // Status of the condition, one of True, False, Unknown.
+    Status v1.ConditionStatus `json:"status"`
+
+    // The reason for the condition's last transition.
+    Reason string `json:"reason,omitempty"`
+
+    // A human readable message indicating details about the transition.
+    Message string `json:"message,omitempty"`
+
+    // The last time this condition was updated.
+    LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
+
+    // Last time the condition transitioned from one status to another.
+    LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+}
+```
+
+**TFJobConditionType**:
+```go
+// TFJobConditionType defines all kinds of types of TFJobStatus.
+type TFJobConditionType string
+
+const (
+    // TFJobCreated means all sub-resources (e.g. services/pods) of this TFJob
+    // have been successfully created.
+    // But they are waiting to be scheduled and launched.
+    TFJobCreated TFJobConditionType = "Created"
+
+    // TFJobRunning means all sub-resources (e.g. services/pods) of this TFJob
+    // have been successfully scheduled and launched.
+    // The training is running without error.
+    TFJobRunning TFJobConditionType = "Running"
+
+    // TFJobRestarting means one or more sub-resources (e.g. services/pods) of this TFJob
+    // reached phase failed but maybe restarted according to it's restart policy
+    // which specified by user in v1.PodTemplateSpec.
+    // The training is freezing/pending.
+    TFJobRestarting TFJobConditionType = "Restarting"
+
+    // TFJobSucceeded means all sub-resources (e.g. services/pods) of this TFJob
+    // reached phase have terminated in success.
+    // The training is complete without error.
+    TFJobSucceeded TFJobConditionType = "Succeeded"
+
+    // TFJobFailed means one or more sub-resources (e.g. services/pods) of this TFJob
+    // reached phase failed with no restarting.
+    // The training has failed its execution.
+    TFJobFailed TFJobConditionType = "Failed"
+)
+```
+
+## Design
+
+### TFController
+
+The TFJob controller `tf-operator` will process TFJobs and CRUD services/pods according to the spec of TFJob. It is responsible for synchronizing TFJob objects stored in the system with actual running services and pods, continuously strive to make the observed state match the desired state.
+
+Here is the definition of `TFController`:
+
+```go
+type TFController struct {
+    // kubeClientset is a standard kubernetes clientset
+    kubeClientset kubernetes.Interface
+
+    // tfJobClientset is a clientset for CRD TFJob
+    tfJobClientset tfjobclient.Interface
+
+    tfJobLister listers.TFJobLister
+    tfJobSynced cache.InformerSynced
+
+    // for pod/service CRUD
+    podLister     kubelisters.PodLister
+    podControl    controller.PodControlInterface
+    serviceLister kubelisters.ServiceLister
+
+    // workQueue is a rate limited work queue. This is used to queue work to be
+    // processed instead of performing it as soon as a change happens. This
+    // means we can ensure we only process a fixed amount of resources at a
+    // time, and makes it easy to ensure we are never processing the same item
+    // simultaneously in two different workers.
+    workQueue workqueue.RateLimitingInterface
+
+    // recorder is an event recorder for recording Event resources to the
+    // Kubernetes API.
+    recorder record.EventRecorder
+
+    // A TTLCache of pod creates/deletes each TFReplica expects to see
+    expectations controller.ControllerExpectationsInterface
+}
+```
+
+### Distributed TensorFlow Configuration
+
+**Auto-Generated TF_CONFIG:**
+
+To make distributed TensorFlow work, user **should** get the distributed TensorFlow configurations `TF_CONFIG` which generated by `tf-operator`.
+This config looks like this:
+
+```json
+{
+    "cluster": {
+        "ps": ["ps1:2222", "ps2:2222"],
+        "worker": ["worker1:2222", "worker2:2222", "worker3:2222"]
+    },
+    "task": {
+        "type": "ps",
+        "index": 1
+        },
+    }
+}
+```
+
+`tf-operator` will append these auto-generated environment variables into `Env` field.
+Check more details from [here](https://cloud.google.com/ml-engine/docs/trainer-considerations#use_tf_config).
+
+**User-Defined Arguments:**
+
+Other user-defined arguments can also be passed into container by `Args` field in `Container` struct.
+
+### Event-Driven
+
+First, we should follow the `Event-Driven` pattern as other resource controller in kubernetes (e.g. Deployment/Job):
+
+- Start `tfJobInformer` to listen on CRUD events of TFJob.
+    + `tfJobInformer` was automatically generated from API definition by `informer-gen` script. 
+- Create one pair pod/service for each specify TFReplicaType + replica index in TFJob CreateHandler.
+    + For example, as a given TFReplicaSpec:
+      ```
+      {
+        "PS": {
+            Replicas: 2,
+        },
+        "Worker": {
+            Replicas: 3,
+        },
+      }
+      ```
+      We will create:
+      - `two` pair pods/services for PSs:
+        - tf-job-name-ps-1-uid
+        - tf-job-name-ps-2-uid
+      - `three` pair pods/services for Workers:
+        - tf-job-name-worker-1-uid
+        - tf-job-name-worker-2-uid
+        - tf-job-name-worker-3-uid
+    + We use a postfix `uid` to make each object name unique.
+    + Then set these objects' `OwnerReferences` to this TFJob object.
+- Listen on pods/services via `podInformer` and `serviceInformer`.
+    + On pod created/updated/deleted, get TFJob object by parsing `OwnerReferences`, set the `TFJob.Status` as defined above according to the whole TF cluster state.
+    + Update the `TFJob.Status.Condition` if needed.
+- Terminate/Delete the TFJob object if every pod is completed (or leave pod phase as `Succeeded`).
+    + This maybe be lead to logs and model checkpoint files unreachable.
+
+### Reconciler
+
+More than that, we should provide a `Reconciler` mechanism to reconcile observed and desired states and repair discrepancies as a double check for `Event-Driven` mechanism.
+
+Here is `configuration` of tf-operator:
+
+```go
+// TFControllerConfiguration contains configuration of tf-operator.
+// DefaultTimerConfig is the suggested tf-operator configuration for production.
+type TFControllerConfiguration struct {
+    // ReconcilerSyncLoopPeriod is the amount of time the reconciler sync states loop
+    // wait between two reconciler sync.
+    // It is set to 15 sec by default.
+    // TODO(cph): maybe we can let it grows by multiple in the future
+    // and up to 5 minutes to reduce idle loop.
+    // e.g. 15s, 30s, 60s, 120s...
+    ReconcilerSyncLoopPeriod metav1.Duration
+}
+
+// DefaultTFControllerConfiguration is the suggested tf-operator configuration for production.
+var DefaultTFControllerConfiguration TFControllerConfiguration = TFControllerConfiguration{
+    ReconcilerSyncLoopPeriod: 15 * time.Second,
+}
+```
+
+Reconciler use a `ReconcilerSyncLoopPeriod` to determine whether we should call this reconciler or ignore it. We should leave a record `LastReconcileTime` in TFJob object of course:
+
+```go
+    // Represents last time when the TFJob was reconciled. It is not guaranteed to
+    // be set in happens-before order across separate operations.
+    // It is represented in RFC3339 form and is in UTC.
+    LastReconcileTime *metav1.Time `json:"lastReconcileTime,omitempty"`
+```
+
+As `tfJobImformer` provides a forcing resync mechanism by calling `UpdateFunc` which defined in `ResourceEventHandlerFuncs` periodically. We can call the reconciler in this function:
+
+- UpdateFunc return a TFJob object periodically.
+- Check `LastReconcileTime` to determine whether we should trigger a reconciler call.
+- `tf-operator` will list all pods/services which related to this TFJob.
+  + Compare the current state to the spec of this TFJob.
+  + Try to recovery the failed pod/service to make the training healthy.
+      + Error handing is described below.
+- Update the status of this TFJob.
+- TODO: we should call this reconciler with an exponential back-off delay (15s, 30s, 60s …) capped at 5 minutes.
+
+### Error Handling
+
+To make the system robust, the tf-operator should be able to locally and automatically recover from errors. 
+
+We extend kubernetes built-in `RestartPolicy` by adding new policy `ExitCode`:
+
+```go
+    RestartPolicyAlways    RestartPolicy = "Always"
+    RestartPolicyOnFailure RestartPolicy = "OnFailure"
+    RestartPolicyNever     RestartPolicy = "Never"
+    RestartPolicyExitCode  RestartPolicy = "ExitCode"
+```
+
+We let users set this field according to their model code.
+  + If set RestartPolicy to `OnFailure`/`Always`, user should add reloading checkpoint code by themselves.
+  + Otherwise restarting will take no effect.
+
+`ExitCode` policy means that user should add exit code by themselves, `tf-operator` will check these exit codes to determine the behavior when a error occurs:
+- 1-127: permanent error, do not restart.
+- 128-255: retryable error, will restart the pod.
+
+### Test
+
+**Unit Test**
+
+TBD
+
+**E2E Test**
+
+We can use this model from TensorFlow [repo](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/dist_test) for e2e test.
+
+## Alternatives Considered
+
+### Future Works
+
+Apart from the above, we should add these abilities in the future:
+
+- Provide a properly mechanism to store training logs and checkpoint files.
+  + [FYI](https://github.com/kubeflow/tf-operator/issues/128)
+
+### Related Issues
+
+- [Refactor TFJobStatus in CRD API](https://github.com/kubeflow/tf-operator/issues/333)
+- [Deprecate the TfImage field](https://github.com/kubeflow/tf-operator/issues/330)
+- [[discussion] Differences between tensorflow/k8s and caicloud/kubeflow-controller](https://github.com/kubeflow/tf-operator/issues/283)
+- [API: some comments about API changes from PR #215 review](https://github.com/kubeflow/tf-operator/issues/249)
+- [Use conditions instead of phase](https://github.com/kubeflow/tf-operator/issues/223)
+- [API Review](https://github.com/kubeflow/tf-operator/issues/64)
+- [Phase is wrong unexpected TfJob phase: Done](https://github.com/kubeflow/tf-operator/issues/110)


proposals/tf-operator-design-v1alpha2.md
```diff
@@ -0,0 +1,435 @@
...
+
+### Test
+
+**Unit Test**
+
+TBD
+
+**E2E Test**
+
+We can use this model from TensorFlow [repo](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/dist_test) for e2e test.
```
inc0: Can you describe example e2e workflow? From beginning (empty jupyter) to the end (multinode tensorflow serving)?

I think it would be useful to have discussion (separate, dedicated doc?) how do we want to connect jupyter to tf-operator with user-experience in mind.

Another thing to consider which might affect this design is how do we transfer artifacts (code and trained models) between steps in workflow - I (data scientist) write code in jupyter -> somehow export it to place where tf-job (learning) can access it -> tf-job runs same code in distributed and scaled fashion -> tf-job saves model somewhere -> tf-serving runs in scaled fashion and consumes trained model.

proposals/tf-operator-design-v1alpha2.md
```diff
@@ -0,0 +1,435 @@
...
+
+To make distributed TensorFlow work, user **MUST** implement a parser in their code like `ArgumentParser` to get the distributed TensorFlow configurations which generated by `tf-operator`, the common built-in arguments are:
+
+- `ps_hosts`: comma-separated list of hostname:port pairs.
+  + For example, `ps0.example.com:2222,ps1.example.com:2222`
+- `worker_hosts`: comma-separated list of hostname:port pairs.
+  + For example, `worker0.example.com:2222,worker1.example.com:2222,worker2.example.com:2222`
+- `job_name`: job name: worker or ps.
+- `task_index`: task index, should be >= 0.
+  + For example, worker task_index = 0 is the `chief` worker task the performs the variable initialization.
```
inc0: I think it would be useful for these configurations to be env-dependant. For example we could run same job in local (notebook) single node for smoke testing and first metrics and then, ideally without code change, run it distributed to do actual learning.

proposals/tf-operator-design-v1alpha2.md
```diff
@@ -0,0 +1,435 @@
...
+    // A TTLCache of pod creates/deletes each TFReplica expects to see
+    expectations controller.ControllerExpectationsInterface
+}
+```
+
+### Distributed TensorFlow Configuration
+
+**Auto-Generated Arguments:**
+
+To make distributed TensorFlow work, user **MUST** implement a parser in their code like `ArgumentParser` to get the distributed TensorFlow configurations which generated by `tf-operator`, the common built-in arguments are:
```
rongou: I wonder if it's better to pass these in as environment variables instead of command line arguments. Seems it would be more flexible, as the user code can ignore these if they don't care about them, rather than being forced to parse them.

inc0: I'd say we could do both, or even add config file into the mix - CLI overriding ENV overriding config file

jlewi: +1 to Rong's suggestion of using environment variables and more specifically the environment variable TF_CONFIG.

TensorFlow has already adopted the environment variabble TF_CONFIG
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/learn/python/learn/estimators/run_config.py#L84

as a convention for passing runtime information (e.g. # of workers) to libraries like tf.Estimator that can process that information.

Why would we try to define a new convention in the form of a set of specific command line arguments?

Passing command line arguments also seems brittle. What if existing code uses different flag parsing conventions e.g. "-" as opposed to "--"? What if code is using tf.Estimator and therefore can automatically use TF_CONFIG and strict argument parsing. Then the argument parser would raise an exception because of unrecognized command line arguments. So users would have to define arguments specific to TfJob operator even though the user's code was never using these arguments.
new convention in the form of a set of specific command line arguments?

Passing command line arguments also seems brittle. What if existing code uses different flag parsing conventions e.g. "-" as opposed to "--"?

proposals/tf-operator-design-v1alpha2.md
```diff
@@ -0,0 +1,435 @@
...
+- `tf-operator` will list all pods/services which related to this TFJob.
+  + Compare the current state to the spec of this TFJob.
+  + Try to recovery the failed pod/service to make the training healthy.
+      + Error handing is described below.
+- Update the status of this TFJob.
+- TODO: we should call this reconciler with an exponential back-off delay (15s, 30s, 60s …) capped at 5 minutes.
+
+### Error Handling
+
+To make the system robust, the tf-operator should be able to locally and automatically recover from errors.
```
jlewi: What about retryable vs. permanent errors? The current implementation uses exit codes to distinguish between retryable and permanent errors? Should we continue to use that?

proposals/tf-operator-design-v1alpha2.md
```diff
@@ -0,0 +1,435 @@
...
+    TFReplicaTypeLocal TFReplicaType = "Local"
+)
+```
+
+**TFJobStatus**:
+```go
+// TFJobStatus represents the current observed state of the TFJob.
+type TFJobStatus struct {
+    // Phase is the recently observed lifecycle phase of the TFJob.
+    Phase TFJobPhase `json:"phase"`
```
jlewi: Do we need Phase if we have conditions? I thought @gaocegege or someone else suggested that K8s was moving away from using Phase and using conditions instead?

proposals/tf-operator-design-v1alpha2.md
```diff
@@ -0,0 +1,435 @@
...
+
+Kubeflow community currently have [tf-operator](https://github.com/kubeflow/tf-operator) (v1alpha1) for running TensorFlow jobs on Kubernetes. And we have received some refactoring [requests](https://github.com/kubeflow/tf-operator/issues?q=is%3Aissue+is%3Aopen+label%3Akind%2Fapi-change) for API changes.
+Open this file to summarize the design details and move the version of API to `v1alpha2`.
+
+## Goals
+
+- Define the structure of API `v1alpha2`.
+    + Cover most of the refactoring requests we have discussed.
+    + Simplify the API definition.
+- Define an `event-driven` mechanism for TFJob life-cycle management.
```
jlewi: Would it make sense to split the implementation into its own proposal as opposed to using 1 proposal for the implementation and API?

proposals/tf-operator-design-v1alpha2.md
```diff
@@ -0,0 +1,435 @@
...
+// TFJobSpec is a desired state description of the TFJob.
+type TFJobSpec struct {
+    // TFReplicaSpecs is map of TFReplicaType and TFReplicaSpec
+    // specifies the TF replicas to run.
+    // For example,
+    //   {
+    //     "PS": TFReplicaSpec,
+    //     "Worker": TFReplicaSpec,
+    //   }
+    TFReplicaSpecs map[TFReplicaType]*TFReplicaSpec `json:"tfReplicaSpecs"`
```
jlewi: Why not get rid of TFReplicaType and use arbitrary strings for the replica name?
The suggestion in various discussions was  to replace TFReplicaType with various properties on the Replica as opposed to inferring them based on the replica type.

As an example, restart behavior is currently inferred based on TFReplicaType. But that could instead be based on specific properties like RestartPolicy.

jlewi:
Review status: 0 of 1 files reviewed at latest revision, 12 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 76 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbsnJ8BSkzNyI-881:-L6rbsnJ8BSkzNyI-882:bcc0ynt) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L76)):*
> ```Markdown
> // TFJobSpec is a desired state description of the TFJob.
> type TFJobSpec struct {
>     // TFReplicaSpecs is map of TFReplicaType and TFReplicaSpec
> ```

What about TerminationPolicy?
https://github.com/kubeflow/tf-operator/blob/master/pkg/apis/tensorflow/v1alpha1/types.go#L58

How does a user specify the termination criterion for the job?

---

*[proposals/tf-operator-design-v1alpha2.md, line 100 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rc8ClfPBx9EywvX2O:-L6rc8ClfPBx9EywvX2P:bzh9f1o) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L100)):*
> ```Markdown
>     // to describe how the containers within the pod should be restarted.
>     // Please set this restart policy carefully according to your code.
>     Template *v1.PodTemplateSpec `json:"template,omitempty"`
> ```

Per comment above, should we get rid of TFReplicaType and instead add properties to TFRepliaSpec to specify relevant behavior?

I think the only thing we use TFRepliaType for is restart behavior. 
Right now we always restart parameters servers; we don't look at exit code to determine if error is permanent or retryable. 
Should we make this configurable so that for each Replica user can specify whether to always retry? Or only retry certain exit codes?

---

*[proposals/tf-operator-design-v1alpha2.md, line 107 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rcfOgWIxA6F51X6r4:-L6rcfOgWIxA6F51X6r5:b-ctgkjj) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L107)):*
> ```Markdown
> ```go
> // TFReplicaType is the type for TFReplica.
> type TFReplicaType string
> ```

Should we get rid of TFReplicaType?

---

*[proposals/tf-operator-design-v1alpha2.md, line 211 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rdDm4jHsSuI0SIxNy:-L6rdDm4jHsSuI0SIxNz:b3d5sv1) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L211)):*
> ```Markdown
>     // which specified by user in v1.PodTemplateSpec.
>     // The training is freezing/pending.
>     TFJobRestarting TFJobConditionType = "Restarting"
> ```

Can a condition be triggered more than once?

---

*[proposals/tf-operator-design-v1alpha2.md, line 303 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-303:-L6rdR3ABPOYhJYvB9J6:b-cl7euj) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L303)):*
<details><summary><i>Previously, inc0 (Michał Jastrzębski) wrote…</i></summary><blockquote>

I think it would be useful for these configurations to be env-dependant. For example we could run same job in local (notebook) single node for smoke testing and first metrics and then, ideally without code change, run it distributed to do actual learning.
</blockquote></details>

Why do you need this to be env dependent in the API to do this? Right now the way you make this env dependent is in ksonnet your TFJob would have different parameters based on the environment. So you might have parameters
num_workers
num_ps
num_master

The values of these could then be set differently for each environment. So that you could run single worker locally and distributed in the cloud.

---

*[proposals/tf-operator-design-v1alpha2.md, line 347 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6reSMkBvSAc9QI7e77:-L6reSMkBvSAc9QI7e78:bjaiv14) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L347)):*
> ```Markdown
> ### Reconciler
> 
> More than that, we should provide a `Reconciler` mechanism to reconcile observed and desired states and repair discrepancies as a double check for `Event-Driven` mechanism.
> ```

The current implementation does the following for any event run reconcile.
An event could be a specific event (e.g. TFJob created/deleted/updated). We also use the informer to generate a periodic update event to ensure reconcile is called regularly.

With the current implementation though, we don't invoke different logic for different events.  This is less efficient  but perhaps more reliable since we have a single code path that is always invoked. 

Do we have any idea how expensive calling reconcile is and whether we are saving a whole lot of resources by trying to add different logic for different events?

The expensive part is probably calling the APIs to list pods/services. Can we use the informer to cache the status of pods/services and reuse them across different TFJobs so that each call to Reoncile doesn't actually send a list request to the API?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/community/30)*

proposals/tf-operator-design-v1alpha2.md
gaocegege: Phase is required but we should use condition to determine the the state of the TFJob.

proposals/tf-operator-design-v1alpha2.md
gaocegege: We have some discussions in https://github.com/kubeflow/tf-operator/issues/369:

> TF_CONFIG is a TensorFlow convention that TF APIs like the EstimatorAPI use to get information about the runtime environment and configure the job appropriately.
>
> If users want to use command line arguments, they can write a launcher script that parses TF_CONFIG and sets the command line arguments as needed. This is what we do in the TFCNN example; see https://github.com/kubeflow/kubeflow/blob/master/tf-controller-examples/tf-cnn/launcher.py
>
> I think the problem with command line arguments is that everyone will use slightly different conventions.

proposals/tf-operator-design-v1alpha2.md
gaocegege: I think tf-operator does not know the jupyter since it is not in tf-operator's management. The operator only knows the YAML config file and I think the e2e test from jupyter to serving should be done in kubeflow/kubeflow.

proposals/tf-operator-design-v1alpha2.md
gaocegege: Thanks for your suggestion, and I think we already implement it since the Kubernetes could schedule all tasks of the job in one node if we only have one node.

proposals/tf-operator-design-v1alpha2.md
gaocegege: I think they are bundled. If we do not migrate the implementation to the new design we have to implement twice, once for etcd-operator style and once for event-driven style. I think it is unnecessary. :thinking:

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: @inc0 Thanks for your high-level comments, it makes totally sense for me.

For e2e workflow we can re-use [kubernetes/test-infra](https://github.com/kubernetes/test-infra):

- Deploy a local/remote kubernetes cluster.
- Deploy TFJob controller and create CRD.
- Run e2e test case:
  - Load e2e test model from scratch.
  - Create TFJob yaml and deploy it.
  - Wait and check the result.

IMHO, put e2e design into a dedicated doc maybe better.
And i think TensorFlow serving is a very important feature of tf-operator, but i don't take something to think about it now :)

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: Hi, I think this is open to discuss. IMHO, separate retryable and permanent errors maybe not easy, so let user to decide this maybe better.

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: Hi, this proposal follows the [proposal-template](https://github.com/kubeflow/community/blob/master/proposals/TEMPLATE.md), it makes sense for me to put API and design together in one proposal :)

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: @jlewi Hi, any tensorflow official docs about `TF_CONFIG`, I check the example from tensorflow website: https://www.tensorflow.org/deploy/distributed.

And i notice that `Run Config (deprecated, use tf.estimator.RunConfig instead).` in here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/learn/python/learn/estimators/run_config.py#L15

ScorpioCPH: And what it user don not use `estimator` library?

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: Sorry, i'm not very clear about your comment: `get rid of TFReplicaType`. IMO, `TFReplicaType` is a key value for distributed TensorFlow.

ScorpioCPH:
Review status: 0 of 1 files reviewed at latest revision, 12 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 76 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbsnJ8BSkzNyI-881:-L6tFcYa3woCcCAELv8y:bv4rpc2) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L76)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

What about TerminationPolicy?
https://github.com/kubeflow/tf-operator/blob/master/pkg/apis/tensorflow/v1alpha1/types.go#L58

How does a user specify the termination criterion for the job?
</blockquote></details>

Hi, can you show us some user stories or use cases what `TerminationPolicy` is used for?

---

*[proposals/tf-operator-design-v1alpha2.md, line 100 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rc8ClfPBx9EywvX2O:-L6tFx6sRh5i98LJ_bIQ:b-u1lnhy) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L100)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Per comment above, should we get rid of TFReplicaType and instead add properties to TFRepliaSpec to specify relevant behavior?

I think the only thing we use TFRepliaType for is restart behavior. 
Right now we always restart parameters servers; we don't look at exit code to determine if error is permanent or retryable. 
Should we make this configurable so that for each Replica user can specify whether to always retry? Or only retry certain exit codes?
</blockquote></details>

We can use `TFRepliaType` for different handling logic, for example, generate TF cluster spec configuration :)

---

*[proposals/tf-operator-design-v1alpha2.md, line 107 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rcfOgWIxA6F51X6r4:-L6tFWH9e6QblAB2I2ek:bqiaom7) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L107)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Should we get rid of TFReplicaType?
</blockquote></details>

Is there any benefit to use arbitrary strings?

---

*[proposals/tf-operator-design-v1alpha2.md, line 211 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rdDm4jHsSuI0SIxNy:-L6tDyTlownKRpq8GuiH:b-dgx3nq) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L211)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Can a condition be triggered more than once?
</blockquote></details>

Yes, will update the LastUpdateTime field.

---

*[proposals/tf-operator-design-v1alpha2.md, line 347 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6reSMkBvSAc9QI7e77:-L6tEe77LVKTtZU3y6qn:baps359) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L347)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

The current implementation does the following for any event run reconcile.
An event could be a specific event (e.g. TFJob created/deleted/updated). We also use the informer to generate a periodic update event to ensure reconcile is called regularly.

With the current implementation though, we don't invoke different logic for different events.  This is less efficient  but perhaps more reliable since we have a single code path that is always invoked. 

Do we have any idea how expensive calling reconcile is and whether we are saving a whole lot of resources by trying to add different logic for different events?

The expensive part is probably calling the APIs to list pods/services. Can we use the informer to cache the status of pods/services and reuse them across different TFJobs so that each call to Reoncile doesn't actually send a list request to the API?
</blockquote></details>

As described in this proposal, `Event-Driven` is our first choice, we should trust the reported observed state from informer and use `Reconciler` as a double-check (maybe deprecated in the future)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/community/30)*

proposals/tf-operator-design-v1alpha2.md
jlewi: https://www.tensorflow.org/api_docs/python/tf/estimator/RunConfig

The RunConfig docs say that it will automatically be populated based on the environment variable TF_CONFIG if it is set.

My hope is that most high level libraries will follow the convention of using the environment variable TF_CONFIG to automatically configure themselves.

As @gaocegege points out above. If a user is using a library that doesn't use TF_CONFIG then they have two options

1. Modify the code to read and parse TF_CONFIG
1. Create a launcher script (possibly using an init container) to parse TF_CONFIG and set command line arguments appropriately.

I really think we should try to avoid introducing new conventions (e.g. standard command line flags) when the TF community already has an existing convention.

Furthermore, when I look at TF code I don't see broad agreement about what command line flags should be used or how they should be structured.

proposals/tf-operator-design-v1alpha2.md
jlewi: @gaocegege Required by K8s?

proposals/tf-operator-design-v1alpha2.md
jlewi: So I agree user should decide whether an error is retryable or not, but that needs to be communicated to the TFJob operator so that it can determine what to do.

For example, here are two different cases where TF could be terminated

1. The VM running the pod is terminated or becomes unhealthy 
1. TF exits with an exception because user tries to read a file that doesn't exist.

My claim is that in the first case TF operator should restart TF since its a retryable error but in second case the error is most likely permanent (file is unlikely to exist if we keep retrying) so the job should fail.

Exit codes provide a mechanism for distinguishing different types of errors. User has some control over the exit code. For example they can wrap their code in a try/catch block and turn different exceptions into exit codes corresponding to retryable or permanent errors as needed.

This isn't great but its fairly straightforward.

proposals/tf-operator-design-v1alpha2.md
jlewi: @gaocegege @ScorpioCPH  Splitting the proposals doesn't mean implementing them twice. The motivation for splitting the proposals is to narrow the scope of each proposal which should speedup agreement. Actual work doesn't need to begin until both proposals are approved.

I'd expect the API to largely be independent of the implementation. So if we move implementation into its own proposal, that will focus discussion on the API and help us converge.

proposals/tf-operator-design-v1alpha2.md
jlewi: Why not just make it

```
TFReplicaSpecs map[string]*TFReplicaSpec 
```

And let the user pick what keys to use and by extension how many different replicas to have. This would allow the user to easily add a new set of replicas to do something like evaluation.

Or in the case of reinforcement learning, they might use a set of replicas to run simulations to generate training data.

proposals/tf-operator-design-v1alpha2.md
gaocegege: @jlewi By convention, I think

jlewi:
Review status: 0 of 1 files reviewed at latest revision, 12 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 107 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rcfOgWIxA6F51X6r4:-L6w_v9Tc5kIyO9kCi05:b-elc25y) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L107)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

Is there any benefit to use arbitrary strings?
</blockquote></details>

Yes. It allows us to support things like eval workers and other use cases more easily because we don't need to introduce new replica types. Instead we can have properties that control different behavior e.g. restart behavior and user can create however many replicas they need/want.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/community/30)*

jlewi:
Review status: 0 of 1 files reviewed at latest revision, 12 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 347 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6reSMkBvSAc9QI7e77:-L6waUz5fxfRZUQDNVF6:bd7hytg) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L347)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

As described in this proposal, `Event-Driven` is our first choice, we should trust the reported observed state from informer and use `Reconciler` as a double-check (maybe deprecated in the future)
</blockquote></details>

SGTM. As long as Reconcile provides a backup I'm satisfied.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/community/30)*

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: But if let user define this filed, how does tf-operator know how to hand it?

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: @jlewi Thanks, will take a deeper look at this later.

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: Hi, 

- In case 1, AFAIK this case will create a node `NotReady` condition, kubernetes will re-schedule all pods which running on this node to another node after a while. It is handled automatically by kubernetes system, not controlled by any user defined config (e.g. RestartPolicy or exit code).

- In case 2, this is a error in container, so let use handle this error maybe better, by setting RestartPolicy.

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: @jlewi Hi, i don't think it a good idea to split one design doc into two pieces.
Keep this in one doc is better for reader to find the details in on place.

proposals/tf-operator-design-v1alpha2.md
jlewi: In case 1, I don't think K8s will reschedule the pods. Pods aren't durable
https://kubernetes.io/docs/concepts/workloads/pods/pod/#durability-of-pods-or-lack-thereof
I think its the controller that creates a new pod but in our case we are creating/managing the pods ourselves.

In case 2. I don't think we can do this with the built in RestartPolicy because the only options are Never or Always and I don't think that's flexible enough.

proposals/tf-operator-design-v1alpha2.md
jlewi: For reference: kubeflow/tf-operator#223 as quoted in their from the docs

> Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. 

My interpretation of this is that not all V1 resources contain phase. So I don't think we should include phase.

I think the information in phase overlaps with conditions. So by providing phase and conditions are API isn't as clean as it could be because users will need to figure out which one to look at.

proposals/tf-operator-design-v1alpha2.md
jlewi: Sure.

proposals/tf-operator-design-v1alpha2.md
jlewi: Why does TFOperator need to do anything differently based on the name? The TFOperator should be able to handle all replicas the same way. Any differences should be controlled by properties added to the Replica.

For example, in v1alpha1 TFOperator assigns different restart policies based on TFReplicaType.

Instead of depending on the name/Type of the replica to determine the restart behavior we could add a RestartPolicy property to TFReplicaSpec.

kubeflow/tf-operator#61 is an example of an issue that would be fixed by using strings instead of TFReplicaType. If we make it a string, then a user could just call the replica "chief" or "master" based on the TF version.

jlewi:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 76 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbsnJ8BSkzNyI-881:-L6yxnMzSDw5aOK_etaT:b2t36fn) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L76)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

Hi, can you show us some user stories or use cases what `TerminationPolicy` is used for?
</blockquote></details>

Some TF programs have a "chief"  replica. Other TF programs use "worker 0" as the chief.

See: https://github.com/kubeflow/tf-operator/issues/192


I think tf_cnn_benchmarks is an example of one program that doesn't use a chief.

So a field like TerminationPolicy gives the user the ability to configure the operator to match their TF program.

---

*[proposals/tf-operator-design-v1alpha2.md, line 100 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rc8ClfPBx9EywvX2O:-L6yyUSGpFcpkpuKY5hc:b-sp3kt9) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L100)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

We can use `TFRepliaType` for different handling logic, for example, generate TF cluster spec configuration :)
</blockquote></details>

We can and still need to generate the TF cluster spec. Right now we generate the TF cluster spec by converting the TFReplicaType enum to a string. If we make it a map[string]TFReplica  we could just use the string names in the cluster config.

This would help with issues like
https://github.com/kubeflow/tf-operator/issues/61

Because we would no longer have to change the TFReplicaType enum everytime TF changes the naming.

---

*[proposals/tf-operator-design-v1alpha2.md, line 107 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rcfOgWIxA6F51X6r4:-L6yywalolSlGQFh-ZrG:bp6qhyd) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L107)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Yes. It allows us to support things like eval workers and other use cases more easily because we don't need to introduce new replica types. Instead we can have properties that control different behavior e.g. restart behavior and user can create however many replicas they need/want.
</blockquote></details>

See also
https://github.com/kubeflow/tf-operator/issues/61

---

*[proposals/tf-operator-design-v1alpha2.md, line 416 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-416:-L6yzNrCbQD9WeavR_RX:bh2v4u8) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L416)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

@inc0 Thanks for your high-level comments, it makes totally sense for me.

For e2e workflow we can re-use [kubernetes/test-infra](https://github.com/kubernetes/test-infra):

- Deploy a local/remote kubernetes cluster.
- Deploy TFJob controller and create CRD.
- Run e2e test case:
  - Load e2e test model from scratch.
  - Create TFJob yaml and deploy it.
  - Wait and check the result.

IMHO, put e2e design into a dedicated doc maybe better.
And i think TensorFlow serving is a very important feature of tf-operator, but i don't take something to think about it now :)
</blockquote></details>

I think this is covered by our CUJs (critical user journeys)
https://github.com/kubeflow/kubeflow/issues/87

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/community/30)*

ScorpioCPH:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 76 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rbsnJ8BSkzNyI-881:-L6zDT7O4Zy1EXzkPWQE:bw4vzqv) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L76)):*
> configure the operator to match their TF program.

Sorry i'm not very clear about your point, it seems like a parameter passing issue: tell user code which worker is chief.
Is this any relationship with `Termination`?

---

*[proposals/tf-operator-design-v1alpha2.md, line 100 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rc8ClfPBx9EywvX2O:-L6z3E4dAIoE9INhqluB:b-mag18d) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L100)):*
> enumerated

---

*[proposals/tf-operator-design-v1alpha2.md, line 107 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rcfOgWIxA6F51X6r4:-L6z3z_AZVL8TAnu32Mv:b-96he9e) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L107)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

See also
https://github.com/kubeflow/tf-operator/issues/61
</blockquote></details>

Hi, I not a TensorFlow expert :) and just heard the `eval workers` first time.
How about enumerate the all type we support now, and add new types in the next version.
Yes, arbitrary strings are free and flexible, but more flexible means increasing mistakes/errors:
For example, user put a typo `workr` in this filed, it hard to debug this typo.
We can report a type error immediately by using enumeration. 

---

*[proposals/tf-operator-design-v1alpha2.md, line 126 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-126:-L6z1p5LxL43WpKsMBuW:b-kuo71f) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L126)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

For reference: kubeflow/tf-operator#223 as quoted in their from the docs

> Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. 

My interpretation of this is that not all V1 resources contain phase. So I don't think we should include phase.

I think the information in phase overlaps with conditions. So by providing phase and conditions are API isn't as clean as it could be because users will need to figure out which one to look at.
</blockquote></details>

Yes, phases are like state machines with explicit, enumerated states and they aren't extensible.
There are Phases and Conditions in Pod/Node spec now.
And FYI, https://github.com/kubernetes/kubernetes/issues/7856, there is a discussion about `Phases` and `Conditions`.

---

*[proposals/tf-operator-design-v1alpha2.md, line 392 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-392:-L6z3i4n0n6iwDCXGCF1:bgajx8j) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L392)):*
> I think its the controller that creates a new pod but in our case we are creating/managing the pods ourselves.

Yes, you got the point. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubeflow/community/30)*

jlewi:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 76 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rbsnJ8BSkzNyI-881:-L70FkGt3IZRnbu-mvvr:bag20am) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L76)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

> configure the operator to match their TF program.

Sorry i'm not very clear about your point, it seems like a parameter passing issue: tell user code which worker is chief.
Is this any relationship with `Termination`?
</blockquote></details>

The TFJob operator needs to know when a job is done. There are lots of different ways a user might signal that their program is done.

1. A user program might have a chief replica and when that chief exits with exit 0 that should be interpreted as the job completing successfully
2. The user might use worker 0 as the chief  in which case TFJob should be marked completed when the worker 0 exits with exit code zero.
3. The user might want TFJob to run untill all workers have completed with exit code 0.

These are different examples of termination policies. TFJob operator needs to know which termination policy a user wants to use so that it can determine correctly when a job is finished..

---

*[proposals/tf-operator-design-v1alpha2.md, line 100 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rc8ClfPBx9EywvX2O:-L70Geap9Y9SN6x0VrLQ:bqlgv6m) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L100)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

> enumerated
</blockquote></details>

Sorry I don't understand the latest reply?

---

*[proposals/tf-operator-design-v1alpha2.md, line 107 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rcfOgWIxA6F51X6r4:-L70H-zP9ySJfIcqwCzm:b-f0uzzh) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L107)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

Hi, I not a TensorFlow expert :) and just heard the `eval workers` first time.
How about enumerate the all type we support now, and add new types in the next version.
Yes, arbitrary strings are free and flexible, but more flexible means increasing mistakes/errors:
For example, user put a typo `workr` in this filed, it hard to debug this typo.
We can report a type error immediately by using enumeration. 
</blockquote></details>

What do others think? 

I'm on the fence about this. The advantage of making it a string is that the user can add replicas we didn't think about. 

Validation could potentially be done in other places. For example, a user could parse TF_CONFIG and make sure the host names match the expected types.

Arguably this should be done by the API that imposes the convention. For example, if tf.Estimator only allows understands "worker", "eval", "master" then if the TF_CONFIG contains some other name e.g. "workr"  then it should raise an error.

---

*[proposals/tf-operator-design-v1alpha2.md, line 126 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-126:-L70HzAl62_ZXaMCbVRL:bhi4umj) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L126)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

Yes, phases are like state machines with explicit, enumerated states and they aren't extensible.
There are Phases and Conditions in Pod/Node spec now.
And FYI, https://github.com/kubernetes/kubernetes/issues/7856, there is a discussion about `Phases` and `Conditions`.

</blockquote></details>

So can we get rid of Phase from our API? The linked issue indicates Brian would like to get rid of Phase and that its existence leads to users thinking incorrectly about controllers.

So why not remove it from our API now when its easy to do so?

---

*[proposals/tf-operator-design-v1alpha2.md, line 392 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-392:-L70IVlKBQbK1PY2QZy_:bkotx1w) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L392)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

> I think its the controller that creates a new pod but in our case we are creating/managing the pods ourselves.

Yes, you got the point. 
</blockquote></details>

So what does that mean in terms of the API? Does our API need to include a way for users to specify whether container terminations are retryable or not?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubeflow/community/30)*

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: @jlewi After investigating, `TF_CONFIG` is used for `tf.estimators`, which is a high level encapsulation. How about pass the both `TF_CONFIG` and `ClusterSpec` together?

- Generate `TF_CONFIG` JSON string and set it in `Command` field.
- Generate `ClusterSpec` args and set it in `Args` field.

So both estimators model and other tensorflow model can work together.

ScorpioCPH:
@jlewi Reviewable is offline on my laptop. Try to summarize the comments now:

- `TF_CONFIG` & `ClusterSpec`
  - we can generate them both for different use case.
  - pass them both into container.
  - use can pass user-defined arguments by themselves.

- TerminationPolicy & exit code
  - we should get rid of `TerminationPolicy` as it is used for identify which worker is chief in API v1alph1.
    - use worker 0 as chief worker by default and keep it as a convention to make API simple.
    - users who are not a TensorFlow expert should not change the default chief worker (worker 0).
  - I strongly recommend do not use any exit codes to interrupt user's code.
  - restarting will take no effect unless user add specific code (reload checkpoint file by themselves).
    - so let user to make this decision makes more sense.
  - how about add `WorkerRestartPolicy` in TFJob.
    - `RestartAll`: user know that they have wrote the reloading code so restarting will take effect.
      - we will set all Pods' `RestartPolicy` to `OnFailure`.
    - `RestartChief`: only set Pod `worker-0` RestartPolicy to `OnFailure`.
    - `Never`: will do nothing.

- TFReplcaType
  - use enumeration for this filed make it clear and explicit.
  - this proposal is focus on TensorFlow distributed training ("ps" and "worker"), so "eval" in out of the scope. I file an [issue](https://github.com/kubeflow/tf-operator/issues/444) for supporting "eval" worker and we can add this feature in the next version maybe :)

jlewi:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 295 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-295:-L74zNFNRoXBhzPvMm3x:bhdxmy7) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L295)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

@jlewi After investigating, `TF_CONFIG` is used for `tf.estimators`, which is a high level encapsulation. How about pass the both `TF_CONFIG` and `ClusterSpec` together?

- Generate `TF_CONFIG` JSON string and set it in `Command` field.
- Generate `ClusterSpec` args and set it in `Args` field.

So both estimators model and other tensorflow model can work together.
</blockquote></details>

If you pass command line arguments, how do you avoid the brittleness issues mentioned above? For example, suppose my program uses TF_CONFIG and therefore doesn't take the arguments you mentioned. Furthermore, suppose I want to treat unrecognized arguments as errors. Now the extra arguments cause my program to crash

More generally, I think an API is cleaner if we avoid unnecessary redundancy.

I think there are good and flexible solutions for programs that want to take the cluster spec as command line arguments.

I think its better if there is a launcher script that turns TF_CONFIG into whatever environment variables people want. This way people can use whatever conventions they want for command line arguments.

Using an init container it would be very easy to inject one or more automatic launcher scripts. 

So I think using a launcher script with/without an init container is a better pattern then us messing with command line arguments.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/community/30)*

ScorpioCPH:
@jlewi Please take a look at this [model](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/dist_test/python/mnist_replica.py), which don't use tf.estimators, they don't know what TF_CONFIG is.

proposals/tf-operator-design-v1alpha2.md
ScorpioCPH: @jlewi For `TF_CONFIG` i mean `Env` filed, sorry for the typo.

jlewi:
@ScorpioCPH Agreed there are lots of models out there that don't use TF_CONFIG. Our TF_CNN example in Kubeflow. Those models can easily be made to run in Kubeflow just by running a launcher script like we do for TF_CNN
https://github.com/kubeflow/kubeflow/blob/master/tf-controller-examples/tf-cnn/launcher.py

jlewi:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 76 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbsnJ8BSkzNyI-881:-L7BjCio72nVyTe-SKqh:bijrtbl) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L76)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

The TFJob operator needs to know when a job is done. There are lots of different ways a user might signal that their program is done.

1. A user program might have a chief replica and when that chief exits with exit 0 that should be interpreted as the job completing successfully
2. The user might use worker 0 as the chief  in which case TFJob should be marked completed when the worker 0 exits with exit code zero.
3. The user might want TFJob to run untill all workers have completed with exit code 0.

These are different examples of termination policies. TFJob operator needs to know which termination policy a user wants to use so that it can determine correctly when a job is finished..
</blockquote></details>

Can you update per our discussion.

To summarize what we agreed in slack.

There are two use cases we want to support: 
1. some users use worker:0 as the "master worker".
2. Others use use "chief" replica type as the "master worker"
(Although Caicloud folks only use 1, there are many users use 2, so we want to support both)

We can infer which one is the "master worker" by the following logic:
 if there is "chief" replica type, it's the "master worker". Else, worker:0 is the master worker.

So we don't need any fields in the spec. 

Can we update the proposal to document the fact that we will infer whether to use worker:0 or chief automatically.

---

*[proposals/tf-operator-design-v1alpha2.md, line 83 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-83:-L7BjnXU-Z3jMmp27mxF:bgpn8bu) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L83)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Why does TFOperator need to do anything differently based on the name? The TFOperator should be able to handle all replicas the same way. Any differences should be controlled by properties added to the Replica.

For example, in v1alpha1 TFOperator assigns different restart policies based on TFReplicaType.

Instead of depending on the name/Type of the replica to determine the restart behavior we could add a RestartPolicy property to TFReplicaSpec.

kubeflow/tf-operator#61 is an example of an issue that would be fixed by using strings instead of TFReplicaType. If we make it a string, then a user could just call the replica "chief" or "master" based on the TF version.
</blockquote></details>

Per our discussion offline.

We agreed to keep this as enum with allowed values

"chief", "worker", "ps", "eval"

---

*[proposals/tf-operator-design-v1alpha2.md, line 100 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rc8ClfPBx9EywvX2O:-L7BjzY1s8c7sM5_41n7:b-z6azl4) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L100)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Sorry I don't understand the latest reply?
</blockquote></details>

We agreed to keep it as an enum with types
chief, worker, ps eval

Does that match your understanding?

---

*[proposals/tf-operator-design-v1alpha2.md, line 107 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rcfOgWIxA6F51X6r4:-L7BkA1d84rYCnn9-3df:b-rv6unj) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L107)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

What do others think? 

I'm on the fence about this. The advantage of making it a string is that the user can add replicas we didn't think about. 

Validation could potentially be done in other places. For example, a user could parse TF_CONFIG and make sure the host names match the expected types.

Arguably this should be done by the API that imposes the convention. For example, if tf.Estimator only allows understands "worker", "eval", "master" then if the TF_CONFIG contains some other name e.g. "workr"  then it should raise an error.
</blockquote></details>

As noted above we will keep this an enum with types

"chief", "worker", "eval", "ps"

agreed?

---

*[proposals/tf-operator-design-v1alpha2.md, line 126 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-126:-L7BkFxezw2E6087-xOx:b-9affn7) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L126)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

So can we get rid of Phase from our API? The linked issue indicates Brian would like to get rid of Phase and that its existence leads to users thinking incorrectly about controllers.

So why not remove it from our API now when its easy to do so?
</blockquote></details>

I think the agreement in slack was to get rid of phase.

---

*[proposals/tf-operator-design-v1alpha2.md, line 295 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-295:-L7BkNpLTozvolHorpos:bklzmlu) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L295)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

@jlewi For `TF_CONFIG` i mean `Env` filed, sorry for the typo.
</blockquote></details>

I think we agreed to only use TF_CONFIG; is that right?

---

*[proposals/tf-operator-design-v1alpha2.md, line 303 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-303:-L7BkTsocuHpT71czrGa:bztq3a) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L303)):*
<details><summary><i>Previously, gaocegege (Ce Gao) wrote…</i></summary><blockquote>

Thanks for your suggestion, and I think we already implement it since the Kubernetes could schedule all tasks of the job in one node if we only have one node.
</blockquote></details>

@Inc0 can we resolve this thread?

---

*[proposals/tf-operator-design-v1alpha2.md, line 392 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-392:-L7Bk_tPmJ2uikq8Bw2i:b-q0amdn) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L392)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

So what does that mean in terms of the API? Does our API need to include a way for users to specify whether container terminations are retryable or not?
</blockquote></details>

Please update the proposal with your latest thoughts after our discussion.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/community/30)*

ScorpioCPH:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 76 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbsnJ8BSkzNyI-881:-L7NveM9U146JEtQrw9C:bvnlxz3) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L76)):*
> if there is "chief" replica type, it's the "master worker". Else, worker:0 is the master worker.

Update this `TFReplicaType` to:

```go
const (
    // TFReplicaTypePS is the type for parameter servers of distributed TensorFlow.
    TFReplicaTypePS TFReplicaType = "PS"

    // TFReplicaTypeWorker is the type for workers of distributed TensorFlow.
    TFReplicaTypeWorker TFReplicaType = "Worker"

    // TFReplicaTypeChief is the type for chief worker of distributed TensorFlow.
    // If there is "chief" replica type, it's the "chief worker". Else, worker:0 is the chief worker.
    TFReplicaTypeChief TFReplicaType = "Chief"

    // TFReplicaTypeEval is the type for evaluation replica in TensorFlow.
    TFReplicaTypeEval TFReplicaType = "Eval"
)
```

---

*[proposals/tf-operator-design-v1alpha2.md, line 83 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-83:-L7NuVkbktWf5UoKuVx3:byxi9tm) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L83)):*
> We agreed to keep this as enum with allowed values
> "chief", "worker", "ps", "eval"

Agreed.

---

*[proposals/tf-operator-design-v1alpha2.md, line 100 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rc8ClfPBx9EywvX2O:-L7NuctMY44ca3vFrU56:bb77ufg) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L100)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

We agreed to keep it as an enum with types
chief, worker, ps eval

Does that match your understanding?
</blockquote></details>

Yes.

---

*[proposals/tf-operator-design-v1alpha2.md, line 107 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rcfOgWIxA6F51X6r4:-L7Nuf1M6c8-7alZSqb6:bb77ufg) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L107)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

As noted above we will keep this an enum with types

"chief", "worker", "eval", "ps"

agreed?
</blockquote></details>

Yes.

---

*[proposals/tf-operator-design-v1alpha2.md, line 126 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-126:-L7NugX2VwTnXUJHW5Sm:b-80tlqk) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L126)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

I think the agreement in slack was to get rid of phase.
</blockquote></details>

Sure.

---

*[proposals/tf-operator-design-v1alpha2.md, line 295 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-295:-L7Nuijk9dzZZ3-ymBfA:bv3yg46) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L295)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

I think we agreed to only use TF_CONFIG; is that right?
</blockquote></details>

Yes, will update the proposal doc.

---

*[proposals/tf-operator-design-v1alpha2.md, line 392 at r1](https://reviewable.io:443/reviews/kubeflow/community/30#-L6rbdTLWdd8LIZdwopV-r1-392:-L7Nuqj6fndk2n1Sj4Bc:b-80tlqk) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L392)):*
<details><summary><i>Previously, jlewi (Jeremy Lewi) wrote…</i></summary><blockquote>

Please update the proposal with your latest thoughts after our discussion.
</blockquote></details>

Sure.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/community/30)*

ScorpioCPH:
@jlewi @lluunn @DjangoPeng @gaocegege @ddysher 
Update this doc after the off-line discussion, PTAL, thanks!

jlewi:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 76 at r1](https://beta.reviewable.io/reviews/kubeflow/community/30#-L6rbsnJ8BSkzNyI-881:-L7lJ36l3OGRUjy63v0R:b-1usesi) ([raw file](https://github.com/kubeflow/community/blob/7607549a534b527f4254f320a1d1263d01e74d55/proposals/tf-operator-design-v1alpha2.md#L76)):*
<details><summary><i>Previously, ScorpioCPH (Penghao Cen) wrote…</i></summary><blockquote>

> if there is "chief" replica type, it's the "master worker". Else, worker:0 is the master worker.

Update this `TFReplicaType` to:

```go
const (
    // TFReplicaTypePS is the type for parameter servers of distributed TensorFlow.
    TFReplicaTypePS TFReplicaType = "PS"

    // TFReplicaTypeWorker is the type for workers of distributed TensorFlow.
    TFReplicaTypeWorker TFReplicaType = "Worker"

    // TFReplicaTypeChief is the type for chief worker of distributed TensorFlow.
    // If there is "chief" replica type, it's the "chief worker". Else, worker:0 is the chief worker.
    TFReplicaTypeChief TFReplicaType = "Chief"

    // TFReplicaTypeEval is the type for evaluation replica in TensorFlow.
    TFReplicaTypeEval TFReplicaType = "Eval"
)
```
</blockquote></details>

Looks good. I see it in the doc below.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubeflow/community/30)*

jlewi:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[proposals/tf-operator-design-v1alpha2.md, line 93 at r2](https://beta.reviewable.io/reviews/kubeflow/community/30#-L7lKCa89JPGu8eRhW6G:-L7lKCa89JPGu8eRhW6H:b-3x6227) ([raw file](https://github.com/kubeflow/community/blob/57fe667ed89816b6badcdb8794339d63ffcd3b22/proposals/tf-operator-design-v1alpha2.md#L93)):*
> ```Markdown
> // RestartPolicy describes how the TFReplicas should be restarted.
> // Only one of the following restart policies may be specified.
> // If none of the following policies is specified, the default one
> ```

Why isn't RestartPolicy a property of TFReplicaSpecs? I don't think we will want all replica specs to have the same restart policy.

For example, for PS it makes sense to have a restart policy of always because these are just TF server and don't run any user code.

Whereas a restart policy of OnError might only apply to the workers.

---

*[proposals/tf-operator-design-v1alpha2.md, line 153 at r2](https://beta.reviewable.io/reviews/kubeflow/community/30#-L7lKetIDsQ1ZbsUDiW_:-L7lKetIDsQ1ZbsUDiWa:b-wwsxwo) ([raw file](https://github.com/kubeflow/community/blob/57fe667ed89816b6badcdb8794339d63ffcd3b22/proposals/tf-operator-design-v1alpha2.md#L153)):*
> ```Markdown
> 
>     // Represents time when the TFJob was acknowledged by the TFJob controller.
>     // It is not guaranteed to be set in happens-before order across separate operations.
> ```

What does "it is not guaranteed to be set in happens-before order" mean?

---

*[proposals/tf-operator-design-v1alpha2.md, line 333 at r2](https://beta.reviewable.io/reviews/kubeflow/community/30#-L7lLbaFEosA80QEopZJ:-L7lLbaFEosA80QEopZK:bsuo1gj) ([raw file](https://github.com/kubeflow/community/blob/57fe667ed89816b6badcdb8794339d63ffcd3b22/proposals/tf-operator-design-v1alpha2.md#L333)):*
> ```Markdown
>       We will create:
>       - `two` pair pods/services for PSs:
>         - tf-job-name-ps-1-uid
> ```

The UID is per job not per resource? So all the items listed here would use the same value for UID.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubeflow/community/30)*

jlewi:
Thanks for making RestartPolicy a property of the ReplicaSpec.
/lgtm

Issue #24, opened 2018-03-01
Sync common labels to ever repository


jlewi:
#12 defined common labels.

Now we want to automate the application of all those labels across our repositories.

Kubernetes uses this tool.
https://github.com/kubernetes/test-infra/tree/master/label_sync

erikerlandson:
Is the goal to get the full cron automation like kube is running, or just run their bazel manually somewhere?

jlewi:
A one off run would be a huge step but it would be nice to eventually do it regularly.

kunmingg:
@jlewi 
label sync k8s job for kubeflow org is created.
Below is sample action logs from dry run against repo: community.
If looks good I'll schedule cron job to sync all repos.

```
time="2018-04-05T19:26:01Z" level=info msg="Reading repos" org=kubeflow
time="2018-04-05T19:26:01Z" level=info msg="GetRepos(kubeflow, false)" client=github
time="2018-04-05T19:26:02Z" level=info msg="Found 1 repos" org=kubeflow
time="2018-04-05T19:26:02Z" level=info msg="Listing labels for repo" org=kubeflow repo=community
time="2018-04-05T19:26:02Z" level=info msg="GetRepoLabels(kubeflow, community)" client=github
time="2018-04-05T19:26:02Z" level=info msg="Syncing labels for 1 repos" org=kubeflow
time="2018-04-05T19:26:02Z" level=info msg=rename from="do-not-merge/work-in-progress" repo=community to="status/in progress"
time="2018-04-05T19:26:02Z" level=info msg=rename from=question repo=community to="community/question"
time="2018-04-05T19:26:02Z" level=info msg=rename from=bug repo=community to="problems/bug"
time="2018-04-05T19:26:02Z" level=info msg=rename from=enhancement repo=community to="improvement/enhancement"
time="2018-04-05T19:26:02Z" level=info msg=create label="status/lgtm" repo=community
time="2018-04-05T19:26:02Z" level=info msg=create label="area/front-end" repo=community
time="2018-04-05T19:26:02Z" level=info msg=create label="area/lifecycle" repo=community
time="2018-04-05T19:26:02Z" level=info msg=create label="area/operator" repo=community
time="2018-04-05T19:26:02Z" level=info msg=create label="community/discussion" repo=community
time="2018-04-05T19:26:02Z" level=info msg=create label="status/done" repo=community
time="2018-04-05T19:26:02Z" level=info msg=create label=do-not-merge repo=community
[...]

Issue #12, opened 2018-02-16
Define common issue labels for all repositories


jlewi:
See kubeflow/tf-operator#326 for an initial discussion of common labels for issues.

I think most of the labels are uncontroversial e.g. labels for different priorities and furthermore should be applied consistently across all repositories.

Opening this issue to track defining those labels across all our current repositories.

Is there any easy way to enforce a common set of labels across multiple repositories?

jlewi:
We should be able to do this programmatically using the GitHub API.
https://developer.github.com/v3/issues/labels/#create-a-label

Here's a shell script that is similar
https://gist.github.com/rentzsch/4540430

The shell script deletes all labels first; which is probably not what we want to do.

I think I'd prefer a python script so we can do things like check if label already exists and if it does maybe just change its color.

The [requests](http://docs.python-requests.org/en/master/) library should make this fairly straightforward.

chrisheecho:
Hi team, @texasmichelle and I worked on the [first iteration of the labels.](https://docs.google.com/spreadsheets/d/1u1jelEPca6-S8Ozcq4snDOBy4q4--J7Ois2YV59PNS4/edit?usp=sharing) Please take a look and review by next community meeting. would love to get feedback if you are interested

I've added comments access to everyone who have access to kubeflow-discuss@googlegroups.com

gaocegege:
@chrisheecho Thanks for your work, generally LGTM although I added some comments.

jlewi:
LGTM.

Created #24 to actually sync the labels.

texasmichelle:
@chrisheecho have we memorialized this list anywhere yet? Perhaps CONTRIBUTING.md in either [kubeflow/kubeflow](https://github.com/kubeflow/kubeflow/blob/master/CONTRIBUTING.md) or [kubeflow/community](https://github.com/kubeflow/community/blob/master/contributing.md) would be a good place. I'll submit a PR using the Sheets doc if you haven't put them somewhere else.

chrisheecho:
Yes that sounds good. Frankly I didn't know we had to do that. Sgtm do what
you gotta do

texasmichelle:
Better to commit a markdown file to the repo instead of relying on a link to a Sheets doc buried in a closed issue.

chrisheecho:
great. if you are busy with your GH issue stuff I can do this too just lmk

texasmichelle:
It's cool - the PR is ready to merge

texasmichelle:
[...]

PR #37, opened 2018-03-12
Master list of project-wide issue labels

texasmichelle:
From #12, for use in #24.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/37)

### DIFF
--- a/issue-labels.md
+++ b/issue-labels.md
@@ -42,7 +42,7 @@ GitHub Issues are tagged with a standard set of labels across all kubeflow repos
 |             | security
 | status      | icebox
 |             | backlog
-|             | in progress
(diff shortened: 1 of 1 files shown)


issue-labels.md
```diff
@@ -0,0 +1,50 @@
...
+|             | other
+| priorities  | p0
+|             | p1
+|             | p2
+| problems    | bug
+|             | refactor
+|             | security
+| status      | icebox
+|             | backlog
+|             | in progress
```
ankushagarwal: Why use whitespace in `in progress` when hypens are used for other labels

texasmichelle: Good catch

PR #46, opened 2018-03-22
Create a proposal for how we can create a historical record of jobs.

jlewi:
* Related to kubeflow/kubeflow#264

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/46)


jlewi:
@dwhitena How does this fit in with what Pachyderm is doing?

dwhitena:
Yes @jlewi, there is a bunch of overlap with what Pachyderm already does along with how we are trying to integrate with KubeFlow (as further detailed here: https://github.com/kubeflow/kubeflow/issues/151#issuecomment-371628634).

First off, many of the goals of Pachyderm are meant to solve the “big problem” of reproducible ML as discussed in the proposal. We totally agree that users need to be able to tack/version models, resources, processing specifications, etc.

In terms of the Goals outlined in the proposal:

1. “Build an agent to persist resource specs to durable storage” - Pachyderm already does this with it’s own pipeline specifications, such that it can tell users which version of the pipeline spec was used to process what data and generate which results. All of this is versioned an persisted in a way that can independent of the k8s cluster. In fact Pachyderm has a way to export/restore pipeline specifications and versioned data as a cluster is taken down and brought back up.  

    Also, as part of the KubeFlow integration with Pachyderm we are planning to have TFJob or other resource specs versioned as input to Pachyderm pipeline stages. As such, they will be automatically made durable, versioned, and tracked with corresponding results.

2. “Allow user defined ETL pipelines to transform raw data into useful table” - This is Pachyderm’s bread and butter already. Users can define ETL pipelines made of stages written in any language or framework. Each of these stages can be scaled as needed, including doing map/reduce style operations. Along with this users can ingress/egress to/from various systems like Databases, because, at the end of the day, the code running in Pachyderm is just normal code that a user might run outside of pachyderm to, e.g., push atomic changes to a DB (e.g., HBASE/BigTable).

  As an added benefit, Pachyderm provides the non-goal of the proposal, which is to “build a comprehensive set of tolls to surface provenance information.” Because Pachyderm has a unified way of dealing with pipelining and data, this provenance information comes naturally and is revealed in the API.

jlewi:
@dwhitena This seems far more complementary to Pachyderm then overlap with Pachyderm.

> “Build an agent to persist resource specs to durable storage” - Pachyderm already does this with it’s own pipeline specifications, such that it can tell users which version of the pipeline spec was used to process what data and generate which results. 

The goal of this proposal to solve the persistence problem for arbitrary K8s resource specs. Is there a way we can leverage Pachyderm to solve this problem?

IIUC the only way to get this feature of Pachyderm would be to express your computation as a Pachyderm pipeline and I think that's too limiting. 

The proposal is explicitly not doing two of the things Pachyderm does

1. Creating a solution to version datasets
1. Creating an ETL framework.

The proposal is explicitly not suggesting people write their ETL pipelines a particular way. So people could certainly use Pachyderm to convert from the object store to the transactions db.

The intent of kubeflow/kubeflow#151 seems very different  to me. I understand the goals of that proposal to be

1. How can we use Pachyderm as a pipeline tool to launch TFJobs (and possibly other types of jobs created by Kubeflow)
1. How can we use Pachyderm's data versioning capabilities with TFJobs and other resources.

Are there parts of Pachyderm that we could use to build the object and/or transactions DB?

Can we extend Pachyderm to monitor and record arbitrary K8s objects as opposed to just Pachyderm pipelines?

How does Pachyderm persist transactions and/or the data processing graph? Is there a way for us to reuse Pachyderm to make sense of the data?

proposals/provenance-logging.md
```diff
@@ -0,0 +1,128 @@
+## Motivation
+Reproducible ML is a [big problem](https://petewarden.com/2018/03/19/the-machine-learning-reproducibility-crisis/). 
+Kubeflow could add a lot of values to users by making it easy for users to track how datasets and models were produced.
```
aronchick: "In order to address requirements in key scenarios, Kubeflow must provide an easy way to track..."

proposals/provenance-logging.md
```diff
@@ -0,0 +1,128 @@
...
+A lot of valuable information is available in K8s resource specifications. Resources like K8s Jobs, TFJobs, ConfigMaps, and Argo workflows
```
aronchick: "This would go beyond just data provenance and job specification; we should also capture the valuable information in setup, including resources like..."

proposals/provenance-logging.md
```diff
@@ -0,0 +1,128 @@
...
+often contain information (e.g. via command line arguments) about inputs and outputs and the code (docker image) used. There are
+two major problems to using this information to track provenance of models and data today:
+
+1. The information is not durable
+
+	* I should not rely on the K8s APIServer as a persistent, historical data store for resources.
```
aronchick: "... Ideally something immutable with hashing, authentication and version controls would be more appropriate (e.g. Git)."

proposals/provenance-logging.md
```diff
@@ -0,0 +1,128 @@
...
+1. The schema is not convenient
+
+	* People use a variety of conventions/patterns to pass inputs/outputs to their programs
+	* Lack of uniformity is a barrier to building tools that can surface information in useful ways.
+
+The purpose of this proposal is to solve these problems to facilitate tools to track and surface provenance.
+
+## Goals
+
+1. Build an agent to persist resource specs to durable storage
```
aronchick: Are we building an agent? Or just creating a standard to which all components in the system must act? E.g. spit out any changes to /audit. 

Also, we should enforce that auditing has been written via hashing/etc - e.g. the next step (e.g. training) would not proceed without the previous one (e.g. data transformation) being written and signed.

proposals/provenance-logging.md
```diff
@@ -0,0 +1,128 @@
...
+
+1. Allow user defined ETL pipelines to transform raw data into useful tables.
+
+## Non-Goals
+
+1. Building a comprehensive set of tools to surface provenance information.
```
aronchick: Also - providing security components to prevent alteration of logs/provenance (e.g. vault, hardware root of trust, etc)

proposals/provenance-logging.md
```diff
@@ -0,0 +1,128 @@
...
+
+
+## Design
+
+As illustrated by the diagram below the system will consist of
+
+1. A cluster agent that watches the APIServer and records K8s objects to a persistent datastore
+1. A row database (e.g. HBASE/BigTable) to store the raw K8s objects
+1. One or more ETL pipelines to turn the raw data into transaction records
+1. A database to store the transaction records
```
aronchick: I wonder if we can add hooks to let other things plug in here. E.g. could this be the start of the pipeline auditing?

proposals/provenance-logging.md
aronchick: Also - solving data provenance & container build auditing (solved elsewhere)

bhack:
@balajismaniam

bhack:
I don't know if @r2d4 has something to add about the lifecycles of an experiment and how this will interact with reproducibility and tracking. I.e. I am thinking about the relationship of the Dev/Execution-Run/Conditional move in Production phases in  relationship with commit/build/deploy loop phases designed by Skafflod.

bhack:
There was also a related topic that involve [Skaffold in CI](https://github.com/jenkins-x/jx/issues/479#comment-375561805). How we want to track "promoted" experiments? Cause generally is a little bit ambiguous what is "production" for an experiment. I think that the interpretation is different between R&D team and product team. What do you think?

jlewi:
Just to reiterate:

The motivation is to experiment tracking and reproducible data science.

But this proposal is much more narrowly scoped.

In particular, the primary goal of this purpose is to figure out a way to create a historical record of K8s resources e.g. TFJobs that is far more durable than their lifetime in the APIServer.

An alternative approach to accomplish this would be to just have an agent that would watch the APIServer and emit the json representation of K8s objects to stdout. Then we could take advantage of cluster logging to provide a record of these objects.

A related approach might be to leverage K8s Auditing infrastructure.
https://kubernetes.io/docs/tasks/debug-application-cluster/audit/

Using the existing cluster logging infrastructure has the potential advantage that we don't need to setup and maintain any additional datastores provided cluster level logging is configured suitably.

For example,

- On GKE stackdriver can be used as a logging backend
- stackdriver can be configured to dump logs to GCS and/or BQ both of which are suitable as inputs to ETL.

So on GKE 
- if the TFJob operator emitted the TFJob spec as a json formatted log entry
- and Cluster was configured to dump logs to GCS and or BQ
- then it would be possible to write ETL pipelines that would make it easy to extract and surface relevant information.

The questions this raises for me are

1. Do we want to have a single cluster level logging agent that can log an arbitrary set of K8s resources or do we want to leave it up to individual controllers/agents (e.g. TFJob operator)
1. Is the same backend used for log storage well suited to our purposes

I'm strongly in favor of having a single cluster level agent so we can easily add new resources.

For the second point, it might make sense to start just by emitting the objects to stdout and relying on cluster level logging. That should provide a pretty good universal solution. It should also be fairly straightforward to implement since we don't need to support for particular DB's like HBASE or Cloud BigTable.

dwhitena:
@jlewi I agree that the scope of this and the KubeFlow + Pachyderm proposal are quite different. However, I do think there are some general Pachyderm principles that are relevant here. In terms of overlap/complementing Pachyderm:

(1) Regarding creating a durable storage/versioning of KubeFlow/TFJob resource specs - https://github.com/kubeflow/kubeflow/issues/151 proposes that TFJob specs would be input as data in Pachyderm pipelines. As such they will be persisted and versioned over time, such that any TFJob can be connected with it's corresponding spec and results. Additionally or alternatively, even if users didn't want to utilize a Pachyderm pipelines to trigger their TFJob, they could still utilize a Pachyderm data repository and it's API to version the specs. At least that would be one option that is already prepared to fulfill this requirement.

(2) Regarding ETL pipelines, my mistake for misunderstanding the context here. To follow up on your comments, I definitely think that Pachyderm could be one option for users to do ETL. In terms of persisting the input/output/transform information, that would already be stored in Pachyderm pipelines specification (which are already versioned). If you wanted to transform those into another format, you could (such as the format that you are proposing) via some connector or export function I think.

To answer your additional questions:

> Are there parts of Pachyderm that we could use to build the object and/or transactions DB?

Yes and no. I mean you could version and update this metadata via Pachyderm and store it in a data repository. However, Pachyderm isn't a DB and doesn't have any arbitrary query functionality. 

> Can we extend Pachyderm to monitor and record arbitrary K8s objects as opposed to just Pachyderm pipelines?

Yes (see point 1 above). This actually wouldn't be an extension. It would really just require one or more Pachyderm workers that launch K8s objects based on input yaml/json. Then right out of the box you would have the yaml/json versioned with the info about how/when/etc. the object was created.

> How does Pachyderm persist transactions and/or the data processing graph? Is there a way for us to reuse Pachyderm to make sense of the data?

This is what we call "data provenance." In pachyderm each collection of data (a "repo") has a set of commitIDs associated with it that represent states of the data. Then there are versioned pipeline specs that define how to process that data. Thus, what Pachyderm tracks by default is a series of transforms like:

(input data repos and commitIDs) -> (processing stage based on versioned spec) -> (input data repos and commitIDs) -> (processing stage based on versioned spec) -> etc...

For any output or intermediate data you can "inspect" the data to get the above series of upstream events, and for any input you can "flush" the commit to get the downstream series of things it triggered.

proposals/provenance-logging.md
```diff
@@ -0,0 +1,128 @@
...
+- Requires less storage
+- Don't need to maintain ETL pipelines
+
+The disadvantage of this approach is that you have to know ahead of time what information is worth preserving. 
+
+### Use logging
+
+We could treat the objects as text and log them using existing logging support in Kubernetes.
+
+I think
```
Jimexist: :-)

proposals/provenance-logging.md
jlewi: We are building an agent.

There's nothing here that can enforce the next step not starting without the previous one being written and signed. 

That would be the responsibility of an orchestration/pipeline system like Pachyderm. 

/cc @dwhitena

proposals/provenance-logging.md
jlewi: Not sure git is really appropriate in this case but otherwise makes sense and I'll add it to the doc.

proposals/provenance-logging.md
jlewi: Isn't it already inherently extensible? Anyone can an ETL job to produce transaction records or consumer the transaction records.

jlewi:
HeptIO has Ark which snapshots all your K8s resources to a tarball containing the YAML specs to support backup and restore.

jlewi:
Closing this PR for now since there's no movement.

PR #54, opened 2018-04-04
Configurable tooling proposal

kkasravi:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/54)


wbuchwalter:
Awesome!
Couple of remarks/questions:
* What does a `component` mean in this context?
* Would it be possible to include some examples scenarios?
For example, as a data scientist, how can I start a new simple TensorFlow training job with this? I would probably have to use `kf train` but what would the yaml look like?
* There is something I am not sure I understand: You state that the goal is to provide a CLI for data scientists but for the part of the CLI that would be most useful to data scientists (the part specific to Kubeflow), you mention: `not addressed per se within this proposal`, what do you mean?

Looking forward to this, looks really good.

proposals/configurable_tooling.md
```diff
@@ -0,0 +1,111 @@
...
+
+![CLI calls a serverless Function](diagrams/CLI2Function.png)
+## Design
+A data scientist's CLI is a flexible set of commands that are executed against a cluster of DL components. Within kubeflow, this set of components will vary and will use `ks` environments that are parameterized. A CLI with a fixed set of subcommands can not accommodate a configurable set of runtime components provided by kubeflow, of which many may be provided by open source contributors. The kubeflow flexible component design mandates a matching CLI design where subcommands can be provided by component authors or other contributors that provide one or more subcommands for a component. For example different contributors may provide a pytorch CLI that can be used with the pytorch component. The author of the pytorch component may provide their own CLI. Different CLI bindings are possible for any component. Each CLI subcommand invocation will delegate its execution to a serverless Function. This design enables a CLI ecosystem where different functionalities not possible within a fixed CLI deliverable are possible.  
+
+### Implementation
```
inc0: Do we want FaaS? We are as strong as our weakest dependency, and FaaS is still very early in the game (imho). I'd be more inclined to use native k8s as much as possible and try to limit external projects.
Besides stability of external projects, each one also adds complexity to deployment and operations. Look how much issues adding StorageClass requirement caused...

proposals/configurable_tooling.md
```diff
@@ -0,0 +1,111 @@
...
+	Deployment              v1beta1.Deployment              
+	ServiceSpec             v1.ServiceSpec
+	HorizontalPodAutoscaler v2beta1.HorizontalPodAutoscaler
+}
+```
+
+
+## Alternatives Considered
+1. A static set of subcommands that prescribes all available actions a data scientist would do.
+1. Using a different client framework other than spf13/cobra. Given that kubectl, kubeless and ks also use this framework there is an opportunity to leverage their subcommands.
```
inc0: I'd also add #52 approach as an alternative - instead of FaaS we have running API endpoint, but problem itself I think is similar.

proposals/configurable_tooling.md
```diff
@@ -0,0 +1,111 @@
...
+      1. `kf shell [-h] <component> [<container>]`
+      1. `kf tensorboard [-h] <name> <yaml>`
+      1. `kf train [-h] <name> <yaml>`
+1. A CLI that is extensible via a subcommand called `command`    
+  1. `kf command [-h] [add|remove|replace|upgrade|list|describe] <subcommand>`
+1. Subcommands call a Kubeless Function (written in python|bash|golang) which:    
+   1. Runs in a container that includes requisite tooling and dependencies like ks, kubeclt, kubeless, python2.7|python3.6
+
+![CLI calls a serverless Function](diagrams/CLI2Function.png)
+## Design
```
inc0: Besides CLI I'd love to see some API so people at some point could write custom software driving kubeflow - like jupyter plugins, CI systems etc. That will also be useful for automatic hyperparameter tuning work (just call api "run" with different hyperparameters).

kkasravi:
Hi @wbuchwalter 

- What does a component mean in this context?
[kam] we don't have the best labeling right now in our *.libsonnet's - but things  tf-job, tf-serve, tf-hub
Would it be possible to include some examples scenarios?
- For example, as a data scientist, how can I start a new simple TensorFlow training job with this? I would probably have to use kf train but what would the yaml look like?
[kam] `kf train ...` would probably take a few parameters rather than a yaml. See below.
- There is something I am not sure I understand: You state that the goal is to provide a CLI for data scientists but for the part of the CLI that would be most useful to data scientists (the part specific to Kubeflow), you mention: not addressed per se within this proposal, what do you mean?
[kam] A follow up proposal will probably be opened today by a colleague that will add `kf train`, `kf serve` and some others. This proposal is mostly about how the `kf` CLI will allow different subcommands to be added, removed, updated - and these subcommands call a kubeless Function (serverless function). I can revisit the proposal and make this more clear.

proposals/configurable_tooling.md
kkasravi: kubeless has some pretty cool features including a CRD operator and events - take a look https://github.com/kubeless/functions/tree/master/incubator. Also there is an opportunity to rework the function operator to address other needs (and it more like a kubeflow operator) see https://github.com/kubeless/kubeless/tree/master/pkg/controller

jlewi:
Why do we need a CLI for managing Jupyter notebooks? Isn't the whole point of JupyterHub to manage notebooks? What problem is a CLI solving that JupyterHub doesn't solve? Is an API needed and JupyterHub doesn't provide one?

What is the point of `kf train [-h] <name> <yaml>` ? How is this different from using `ks apply` or `kubectl apply`? 

Same question for `kf serve`.

For tensorboard why do we need a CLI? There's an open proposal to make [TB a service](https://github.com/tensorflow/tensorboard/issues/92)? Would that eliminate the need for a CLI because we would no longer need to launch/mange individual instances of TB for jobs?

I 100% agree that different folks will be managing/configuring Kubeflow. We have an existing issue related to that see kubeflow/kubeflow#105. I think that problem is very different though from making things easier for datascientists. I think one of the problems we have with deployment is we need a better way to auto-configure Kubeflow for a user's K8s setup. Its not obvious to me that a CLI will create a substantially different experience from what we have today (compared to perhaps a web flow that walks through configuration and spits out a ksonnet app).

I think the idea in the proposal of creating a very flexible mechanism to implement a CLI (and arguably by extension APIs) is very proposal. 

The proposal mentions a number of different possible.

My suggestion would be to try to pick a very well defined use case where we think a CLI/API would be useful and then use that to test out the solution proposed for building a CLI.

One problem worth considering is fetching logs logs. One of the usability problems with TFJob and other batch jobs is making the logs easily accessible after the job finishes. This problem isn't solved by kubectl logs because that depends on pods persisting which we don't allow. 

So fetching the logs depends on clusters being configured to persist logs to durable storage. This in turn means every K8s distribution has a different solution for how users fetch logs.

I think it would be very useful if we could have a CLI and API that would support fetching logs by labels

```
kf logs -l tf_job=my-job,replica_name=master,replica_index=0
```

To support this we would need to make it very easy for folks to write plugins to fetch logs from different backends (e.g. splunk/stackdriver etc...). So I think this would be a good way to test out the ideas in the proposal.

I think this is a better starting point than `kf train` & `kf serve` because I suspect it will be easier to get consensus on what the API should look like. I think for `kf train` & `kf serve` I think its going to be hard to get consensus on which fields/parameters should be exported and I'm highly skeptical that you will ultimately get an API that's much simpler than the underlying K8s APIs.

kkasravi:
@jlewi sure i will proceed with a logs plugin that leverages labels across resources and use this as a way to further explore 'configurable tooling' concepts

inc0:
@jlewi how about use case like this:

1. I'm a data scientist trying to develop solution for problem X (say detecting cats)
2. I want to be able to try multiple models and compare them easily
3. I want to have minimal setup required locally

So solution (proposed):
1. Create common object that will allow to gather all potential solutions for cat detection, let's call it "model", kf create model "catdetect"
2. I want to write code to try
2.1 I want to use my own editor locally and upload code to be run in k8s, kf create version --file /path/to/cat/detect/v1 --model catdetect
2.2 I want to use notebook that will automatically create model version for me when I want it to, kf create notebook --model catdetect
2.3 notebook will know it's model's context, be allowed to edit existing versions or create new version (via plugin) - that's main difference between our current use of jupyterhub vs what I think we should work towards - having some context awareness in jupyter notebook instance
3. I wrote code, now I want to run it in distributed tensorflow, kf train --replicas 3 --model catdetect --version v1
3.1 2.1 and 3 can be one shortcut CLI call that will first create version and then run training
3.2 I can get logs of training easily by calling kf logs --model catdetect --version v1
4. After training multiple version I want to easily compare their performance, kf tensorboard --model catdetect -> link to tensorboard
5 I want to pick best one and run serving out of it, kf serve --model catdetect --version v3 -> serving url

Now if this workflow looks good, that requires couple things, not all can be modeled with ksonnet
1. Context persistence - we have overarching entity (model) that will separate work. Having this persistent resource allows us to navigate over child resources (versions) and group them together (to compare via tensorboard for example). Model could be CRD. Currently there is no logical connection between notebook instance and tf-job that could run it. Moving code from notebook to tf-job is hard and I can't see how it can be managed for ongoing research, where you want to tweak and try all the time.
2. Model needs to have storage backend (or S3/gs) for:
2.1 code - multiple files for every version
2.2 checkpointed saved model - multiple checkpoints per version (model here means tf.Saver thing)
2.3  logs for tensorboard
That means we need it to know where to save and look for stuff, so it has to have it configured (part of CRD?)
3. Having this mechanism will make things like hyperparameter tuning (even automatic) much easier - change params and create new model version. All that can be programatic so it will be easy for people to for example spawn a notebook instance to look at given code+params and results in tensorboard.

@jlewi is this what you were asking for? This scenario can be implemented in many different ways (API service, kubeless functions), if we agree it's good use case scenario, next step would be figuring out implementation backend.

kkasravi:
interesting @inc0 user story. I agree that establishing a context which could be modeled as a CRD would simplify the data scientist CLI and certainly reduce redundant arguments across the CLI like '--model catdetect'. 

Reasons for bundling subcommands under a CLI as suggested by this proposal is best summarized by Brian Grant in this [slide](https://docs.google.com/presentation/d/10dp4hKciccincnH6pAFf7t31s82iNvtt_mwhlUbeCDw/edit#slide=id.g34527ffc9c_1_9). This is why `ks apply cloud -c job1` is fundamentally different than `kf train job1`. In the former the subcommands {apply, component, delete, diff, env, generate, help, init, param, pkg, prototype, registry, show, upgrade, validate, version} are very different than `kf` which in your example would be {create, train, logs and serve}. If these subcommands share a context such as Model as you suggest then most of the arguments can also be eliminated across these commands. This is why an extensible CLI that composes over a Model is superior to using underlying commands like ks or kubectl even if a given subcommand looks similar. You have, in your user story, described a [monad](https://en.wikipedia.org/wiki/Monad_(functional_programming)) -  "Create functions that use the data type, and compose them together into actions, following the rules defined in the first step." Functionally it would look like kf.create(model).train.logs.serve.

kkasravi:
@jlewi providing a Context as a CRD as @inc0 suggests or defining a monad as I noted is what @inc0 has done in his example then your concern
> I think this is a better starting point than kf train & kf serve because I suspect it will be easier to get consensus on what the API should look like. I think for kf train & kf serve I think its going to be hard to get consensus on which fields/parameters should be exported and I'm highly skeptical that you will ultimately get an API that's much simpler than the underlying K8s APIs.

above is solved because each action operates on the data type (CRD) and there is no need to explicitly export fields/parameters.

jlewi:
@inc0 There's a lot to unpack in your comment. 

Let me start with the more straightforward bits

```
kf train --replicas 3 --model catdetect --version v1
```

What is model in this context? What is the full set of parameters that kf train would take? How does this ultimately lead to a different/better experience than

```
kubectl create -f  my_tfjob.yaml
```
Where my_tfjob.yaml is a TFJob spec?

To ask that another way which fields in TFJob spec aren't you going to expose? 

My contention is that almost all of the fields in TFJob are needed. For any given job you might only modify a couple, but if you take the union of any reasonable subset of jobs and you will end up needing all the fields. 

How are you defining model?

In TensorFlow we generally define a model as graph + weights and maybe some metadata about layers correspond to inputs and outputs. [TensorFlow saved model](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/saved_model/README.md) is the file format TF users.

What do you mean by context awareness for the model and JupyterHub? 

I think what you are getting at is model provenance; e.g. understanding what data, hyperparameters, code was used to produce a model.  That's a big problem and I think well beyond the scope of the original proposal.

Have you taken a look at Pachyderm? Pachyderm does a pretty good job of recording model provenance and then making it accessible to the users.

I really like @kkasravi's idea of using kubeless to add extension points to make it easy to interact with Kubeflow and possibly building a CLI.  

Other areas where a CLI might be useful

- Staging files to/from the cluster

wbuchwalter:
@jlewi :
> What is model in this context

IIUC The model was defined beforehand by:
 * `kf create model "catdetect"` 
* and `kf create version --file /path/to/cat/detect/v1 --model catdetect`.

`kf train --version x` would then basically start `/path/to/cat/detect/v1` in a container.

>  which fields in TFJob spec aren't you going to expose?

Ultimately probably most of them, but I think the goal is to expose them in a more user friendly way.
I.e, if I am a data scientist and I don't know k8s, what is `apiVersion`? How do I know when to update it? What is the `Kind` how do I know if a `kind` is availble for my framework of choice? 
Why is my my job configuration wrapped in `spec`, `replicaSpecs`, and `template` objects? What do they mean? Sure I could copy paste and it would mostly work but that's not exactly a great UX.

Some tools I have seen allow users to do something like `kf train --tensorflow ./my-code/ --replicas 3 ...`. The code is then uploaded on the cluster and mounted directly into a container compatible with the framework specified via flag which I think is a great feature. A CLI would allow us to expose the API in the most user-friendly way possible, not constrained by compatibility with kubectl/ks.

> I think what you are getting at is model provenance; e.g. understanding what data, hyperparameters, code was used to produce a model. That's a big problem and I think well beyond the scope of the original proposal.

Agree. While what @inc0 proposed is very intersting IMHO, I think we should tackle this one step at a time to get it right. And it seems the first step is validating @kkasravi architecture proposal. Logs sound good to me.

inc0:
So @wbuchwalter is spot on:)

> To ask that another way which fields in TFJob spec aren't you going to expose?

if you look at https://github.com/kubeflow/tf-operator/blob/master/examples/tf_job.yaml , only thing we would need to specify here would be number of replicas (we can provide options for specifying number per role (--masters 1 --ps 1 --workers 3). Since code will be mounted as volume (or S3) to node, image can be defaulted to one we (kubeflow) support, and allow people to override. This way simplest cli call would be

`kf train --version 1 --workers 3`

> How are you defining model?

Maybe I made mistake by calling it "model". I agree, probably we should use different name for this to avoid confusion. For sake of discussion, I'll stick to "model" tho. So this is really one overarching resource for other sub-resources, which are what really matter. Think of it like namespace in k8s world.
We need this kind of overarching common namespace to allow things that would compare versions for example (think tensorboard to compare different versions of model). Version of model is piece of code people write, checkpoint is what you called "model" - graph+weights.

> What do you mean by context awareness for the model and JupyterHub?

Think of use case like this - you spawn notebook and have your existing code versions there, you can tweak, edit and create new version to compare with previous. Another thing would be set of controls in jupyter notebook that would spawn TFJob with one click - and since it'll land in same `model` as previous, you'll also have tensorboard for performance comparison.

> I think what you are getting at is model provenance

Model provenance will be easier to handle with this approach, but UX was my main goal here.

As for @kkasravi approach, I believe we're talking about similar thing with few architectural differences, which we can flesh out once we agree that general idea is ok. If you think idea itself is good, I'd suggest writing PoC somewhere to test-drive few approaches to this problem.

wbuchwalter:
It might also be worth including @YujiOshima and @gaocegege to get their feedback on how this would integrate with hp-tuning as there might be overlap with what Vizier/Katib is doing.

jlewi:
> To ask that another way which fields in TFJob spec aren't you going to expose?

> if you look at https://github.com/kubeflow/tf-operator/blob/master/examples/tf_job.yaml , only thing we would need to specify here would be number of replicas (we can provide options for specifying number per role (--masters 1 --ps 1 --workers 3). Since code will be mounted as volume (or S3) to node, image can be defaulted to one we (kubeflow) support, and allow people to override. This way simplest cli call would be

> kf train --version 1 --workers 3

I agree with @wbuchwalter that some of the K8s metadata fields (APIVersion, Kind) are somewhat boilerplate that can be off putting to newcomer. But those are a small subset of the fields you are implicitly eliminating in the example

> kf train --version 1 --workers 3

 https://github.com/kubeflow/tf-operator/blob/master/examples/tf_job.yaml is a contrived example. Its not at all representative of real jobs.

- Where is image specified?
- How do users specify resources (RAM/CPU/GPU)? Are we going to force users to have the same resources for chief, workers, eval, and ps?
- What about command line arguments? 
- What about environment variables e.g. GOOGLE_APPLICATION_CREDENTIALS and S3 environment
  variables?


> IIUC The model was defined beforehand by:
kf create model "catdetect"
and kf create version --file /path/to/cat/detect/v1 --model catdetect.
kf train --version x would then basically start /path/to/cat/detect/v1 in a container.

The use of model is confusing to me in this context. It seems to me you are creating a specification for how to train the model. One might even call this a tensorflow job specification :)

Conceptually, train typically takes a bunch of  arguments that specify things like the following

- Docker image containing the code
- path of input data
- path to write the model
- Number of steps to train
- Model hyperparemeters
- Maybe an initial checkpoint to start from

How does the proposal of having `kf create model` fundamentally change this? I still need to specify all this information.

inc0:
> Docker image containing the code
So lot's of issues above are coming from "image congaing code". In this idea code is saved in S3 and image only contains deps, which means we can support reasonable default image, hence no need to specify it every time (although it needs to be possible).

> How do users specify resources (RAM/CPU/GPU)? Are we going to force users to have the same resources for chief, workers, eval, and ps?

Additional options can be added and CLI won't be only way to create TFJob for this, you can always write yaml for power-user-stuff.

> What about command line arguments?

Again, can be added to CLI easily, I listed simplest example which would just do python << path to file >>

> The use of model is confusing to me in this context. It seems to me you are creating a specification for how to train the model. One might even call this a tensorflow job specification :)

Well, tbh "model" here is really common specifications to train multiple models that solve same issue:) So yeah, bad name. I used nomenclature from ml-engine, maybe incorrectly.
Model version corresponds to code + parameters, so one model-version will be one job spec.
One model version can produce multiple checkpoints, which are actual graphs+weigths from different stages of training.

> 
    Docker image containing the code
    path of input data
    path to write the model
    Number of steps to train
    Model hyperparemeters
    Maybe an initial checkpoint to start from

Image - since code is in S3, image can be our default kubeflow tf image
Path of input data - specified in code, we can experiment in adding it to model definition somewhere
Path to write the model - That's why I say when we create "model" we need to specify storage backend. We specify it once vs every time you run a job
Number of steps to train - code or parameter
Model hyperparameters - as above
Initial checkpoint - as above

jlewi:
@kkasravi What do you think we should do with this PR? Should we just leave it open while you work on the POC?

kkasravi:
@jlewi we have been vetting this internally since a number of groups have each thought along the lines of a CLI. @nqn is our gatekeeper and can probably provide some context and a timeline. Along these lines - we have also been looking at the logging problem (disappearance of pods/logs). I've noted your proposal to the ppl working on logging internally.

jlewi:
Thanks. 

The issue of logs came up recently in kubeflow/tf-operator#536

Is your plan to try to tackle both 

1. Kubeflow Deployment
1. Kubeflow usage (e.g. submitting jobs / models / etc...)

at the same time?

kkasravi:
we would like to work on the deployment phases first since this will open kubeflow up to different cloud-providers and auth-providers. It also best leverages work you started in bootstrapper and Michal is contributing to. I think focused contributions here have a good chance of aligning with your expectations and address the 3 p1's you opened on bootstrapper in the last day or so. For the logging issue, I work with Karthik (kubeflow/tf-operator#536) and Niklas and I talked about the logging issue earlier today. Let me get more of what he had in mind - it may not involve a specific CLI approach. Does this sound ok to you?

nqn:
+1 to @kkasravi and just to clarify, let's close this PR and take the individual ideas and proposals in their own sharp PRs

jlewi:
+1 to the recent comments by @kkasravi & @nqn

jlewi:
@kkasravi Thoughts on what to do with this PR?

nqn:
Thought we agreed on closing this one? I don't have the privileges to do so. Maybe @jlewi you could click the button?

kkasravi:
i can close it. A separate proposal or issue can be  opened if there is a need.

PR #58, opened 2018-04-06
Start a directory to keep track of contact information for members

### LINKED ISSUE #57: Create CSV/Yaml directory in kubeflow/community
I propose we use a csv file in kubeflow/community to list 
  - First Name
  - Last Name
  - GitHub id
  - Slack Id
  - Email (preferably corporate)

This would be entirely opt-in and users could provide a subset of information.

I think this would serve two benefits
  - Short term help folks map user ids across different systems
  - Eventually be used to manage other permissions (e.g. membership in the GitHub org) e.g #49 and access to our release/test infrastructure (#55).

This would  ultimately allow us to use the normal code review process to manage permissions. I think this will be necessary as we continue to grow as a community.

jlewi:
This is entirely optional.

Immediate use case is to make it easy to map ids between various systems.

Fix #57

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/58)

### DIFF
--- a/members.yaml
+++ b/members.yaml
@@ -1,0 +1,17 @@
+# This file is used to provide a mapping between various ids of members
+# in the community.
+# It is entirely opt in. The only required field is github id.
+#
+# Please keep sorted by github id.
+# For slack please use the id on kubeflow.slack.com
+# For email corporate/professional email addresses are preferred.
+#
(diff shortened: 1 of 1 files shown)


jlewi:
@ewilderj You need to actually do "/lgtm"

PR #62, opened 2018-04-11
[WIP] Kubeflow operator and API

inc0:
For some reason github closed my PR #52 . re-creating.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/62)

### DIFF
--- a/proposals/kubeflow-operator.md
+++ b/proposals/kubeflow-operator.md
@@ -1,79 +1,58 @@
 
-## Motivation
+# Motivation
 As Kubeflow supports more than one framework, there would be use for overarching control plane to help with common machine learning
 steps like uploading code or training model. This control plane then could expose API that can be leveraged by various programming languages, CLI,
 plugins etc.
 
-## Goals
+# Goals
 Create service deployed as part of kubeflow-core that would expose API that then could be consumed by swagger to generate client libs and CLI tool.
 
-## Proposed API
+# Use cases
 
-### Model
+## Easy start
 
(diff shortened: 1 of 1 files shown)


jlewi:
There were unresolved comments in #52 e.g. about the scope of this proposal. 

Consider the following use case

- A user deploys Kubeflow using the POC you demoed last week i.e. by doing kubectl create -f /some/spec.yaml
- User is happy with Kubeflow
- User now wants to customize Kubeflow; in particular they want to make the default Jupyter images custom images that include proprietary packages

How would the user do this customization? Would they continue to deploy using /some/spec.yaml or would they have to switch to creating the ksonnet app directly and customizing deploying that?

inc0:
@jlewi 

well, all configuration will be stored in App resource. You can always kubectl edit it and rerun ks apply (if this is appliable, sometimes we might need to recreate resource. I'm not sure how ksonnet itself would deal with various config upgrades, we won't be better than that:)). To rerun ks apply you can rerun job with different config, which can be wrapped in CLI to do something like:

1. kubectl edit my-app-definition
2. kf env update

jlewi:
If we're proposing  using an App resource then I think the proposal should explicitly talk about the sig-apps App resource and whether we will use that and if not why not.

jlewi:
What's the plan for this PR?

inc0:
I'll get back to it soon, playing with few ideas first

jlewi:
Closing this because its stale.

PR #66, opened 2018-04-14
Update member organizations list

ddutta:
from Cisco

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/66)

### DIFF
--- a/members.yaml
+++ b/members.yaml
@@ -15,3 +15,10 @@
   lastName: Lewi
   affiliation: Google
   email: jlewi@google.com
+
+- github: ddutta
+  slack: debo
+  firstName: Debo
+  lastName: Dutta
+  affiliation: Cisco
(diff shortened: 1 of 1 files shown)


members.yaml
```diff
@@ -15,3 +15,10 @@
   lastName: Lewi
   affiliation: Google
   email: jlewi@google.com
+
+- github: ddutta
+  slack: debo
+  firstName: Debo
+  lastName: Dutta
+  affiliation: Cisco
+  email: dedutta@cisco.com
```
gaocegege: I think it has conflicts with your another PR #67 , maybe we could remove the changes here.

ddutta: thx ..

ddutta: I committed when the branch did have the conflict. Will resolve it ASAP

PR #105, opened 2018-04-20
A simple script to print the names of the companies.

jlewi:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/105)


gaocegege:
Why do we need such a script?

jlewi:
Because we are preparing slide decks for Kubecon and we want to list all the participating organizations.

PR #116, opened 2018-04-24
Add Uber to member organizations

jiezhang:
* Per our legal team, please inform us if/when Kubeflow starts using this list for more than attributing to the contributors.
* Please also note that we are not authorizing Kubeflow any right to use Uber’s name and logo in any of its promotion material.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/116)

### DIFF
--- a/member_organizations.yaml
+++ b/member_organizations.yaml
@@ -147,6 +147,14 @@
   hiringForKubeflow: true
   hiringUrl: https://www.seldon.io/careers/
 
+- name: Uber
+  contacts:
+    - asergeev@uber.com
+    - jiez@uber.com
+  url: https://eng.uber.com/michelangelo/
+  hiringForKubeflow: false
+  hiringUrl: https://www.uber.com/careers/
+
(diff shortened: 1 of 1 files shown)


jiezhang:
@jlewi Can you please also add @alsrgv as kubeflow members?

jlewi:
If you don't want Uber's name to be included in promotional materials e.g. mentions at Kubecon about participating organizations then you shouldn't add Uber to member_organizations.yaml.

The primary purpose of this file is to keep track of companies that would like to be acknowledged as being part of Kubeflow.

jlewi:
Sent @alsrgv an invite.

jlewi:
SGTM.

alsrgv:
@jlewi, as discussed offline, we're fine with being on the list as long as we can remove ourselves if needed.

jlewi:
Yup just submit a PR.

/lgtm
/approve

PR #124, opened 2018-05-02
Kubeflow Bootstrapping Modifications

kkasravi:


proposals/bootstrapper-proposal.md
```diff
@@ -0,0 +1,253 @@
...
+| initialization<br/>(authentication) | cluster-admin | bootstrapper | Configure auth provider under ambassador<br/>(separate proposal) |
+| deployment<br/>(authorization) | data scientist | controller | Process ManagedNamespace CR<br/>Create RoleBindings for users in Organization
+| deployment | data scientist | controller | Continue with normal deployment processing normally done in [Server.Run](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/cmd/bootstrap/app/server.go) |
+
+
+RBAC rules will be created to enable actions on resources at the cluster level and actions on resources scoped by a namespace. The initialization phases will perform actions at the cluster level. The deployment phases will perform actions within a namespace. ServiceAccounts will be modified to use RoleBindings of the active user.
+
+
+## Non-Goals
+Authentication of a data scientist using an auth provider. This will be described in a subsequent **authn** proposal that also adds github as an auth provider.
```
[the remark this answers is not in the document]
jlewi: +1

proposals/bootstrapper-proposal.md
```diff
@@ -0,0 +1,253 @@
...
+- Use a `<provider>.libsonnet` which includes:
+  - ManagedNamespace, Organization, Team, Member CRDs<br/>
+  - CompositeController, DecoratorController from metacontroller
+  - ClusterRoles for kubeflow (admin, write, read)
+
+``` yaml
+apiVersion: apiextensions.k8s.io/v1beta
+kind: CustomResourceDefinition
+metadata:
+  name: member.kubeflow.org
```
jlewi: It seems like you are proposing we create a bunch of CRDs to handle groups and users? Shouldn't this be handled by K8s auth providers.

For example, on GCP my expectation is I can use Google Groups and IAM to manage teams and users. Ideally I should be able to associate a Google Group with an RBAC role so that I can grant a set of users appropriate permissions in the cluster.

proposals/bootstrapper-proposal.md
```diff
@@ -0,0 +1,253 @@
...
+      affiliation: Doe Inc
+      email: msmith@doeinc.com
+```
+
+
+
+## Alternatives Considered
+Using [kubeless](https://kubeless.io/) functions rather than [metacontroller](https://github.com/GoogleCloudPlatform/metacontroller) controllers
+
+Using the [Application CRD](https://github.com/kubernetes-sigs/application#what-is-the-application-crd) instead of the ManagedNamespace CRD.
```
jlewi: Why did you rule out Application CRD? I think the Application CRD solves different problems. Its basically a way of specifying that a group of K8s resources together form an application. I don't think it handles deployment or lifecycle management.

proposals/bootstrapper-proposal.md
```diff
@@ -0,0 +1,253 @@
...
+| :---: | :---: | :---: | :---: |
+| train | tf-job-operator | tf-job-operator | tf-job-operator |
+| job-dashboard | tj-job-dashboard | tf-job-dashboard | tf-job-dashboard |
+| central-ui | centraldashboard | centraldashboard | centraldashboard |
+| nfs | system:persistent-volume-provisioner | nfsnfs-provisioner | kubeflow:nfs |
+| IAP | cloud-endpoints-controller | cloud-endpoints-controller | kubeflow:cloud-endpoints-controller |
+
+
+## Goals
+Divide bootstrapper into 2 phases: an initialization phase and a deployment phase. Each phase will run with different authorizations/users.
```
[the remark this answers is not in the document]
kkasravi: the main idea is to enable data scientists to create kubeflow deployments w/o running with cluster level RBACs. It's similar to the deployment manager / bootstrapper issues opened by @kunmingg [#757](https://github.com/kubeflow/kubeflow/issues/757) [#758](https://github.com/kubeflow/kubeflow/issues/758)

Once certain cluster level resources are created (PVs, namespace) - the deployment of the app should only require rolebindings for that namespace.

proposals/bootstrapper-proposal.md
```diff
@@ -0,0 +1,253 @@
...
+
+Introduce a ManagedNamespace CRD and controllers. Submitting a ManagedNamespace CR will execute the deployment phase without requiring the data scientist to have cluster-admin privileges.
```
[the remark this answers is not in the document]
kkasravi: the ManagedNamespace is watched by a controller that will create cluster level resources including the namespace. The second controller watches the namespace and will deploy into that namespace using the data-scientist's rolebinding.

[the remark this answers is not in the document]
kkasravi: the deployment of the manifests could be done with different tools - this is to separate the initialization of cluster resources including the namespace from the deployment of the manifests into that namespace which could be done with less privileges

proposals/bootstrapper-proposal.md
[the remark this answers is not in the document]
kkasravi: yes auth is a non-goal. As you mention this proposal looks to define the RBAC roles that data-scientists would bind to and kubeflow-admins would bind to.

proposals/bootstrapper-proposal.md
```diff
@@ -0,0 +1,253 @@
...
+
+RBAC rules will be created to enable actions on resources at the cluster level and actions on resources scoped by a namespace. The initialization phases will perform actions at the cluster level. The deployment phases will perform actions within a namespace. ServiceAccounts will be modified to use RoleBindings of the active user.
+
+
+## Non-Goals
+Authentication of a data scientist using an auth provider. This will be described in a subsequent **authn** proposal that also adds github as an auth provider.
+
+## UX
+
+| An admin wants to initialize a cluster for kubeflow using a provider  |
```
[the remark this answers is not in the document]
kkasravi: admin deploys cluster-level resources and creates the namespace. data-scientist deploys into that namespace using rolebindings scoped to the namespace

proposals/bootstrapper-proposal.md
kkasravi: yes different auth providers would create the users and teams. The management of these users to kubeflow deployments still needs to be done so they can deploy into a namespace without cluster rolebindings. Some entity (a controller) needs to check if the user is whitelisted to deploy kubeflow instances and then create the namespace on the user's behalf.

proposals/bootstrapper-proposal.md
kkasravi: I thought it had some aspects that might allow it to be used or extended in a way that would address what this proposal is suggesting and could also be leveraged for generating and deploying different 'bundles' (pytorch, tf, ...)

proposals/bootstrapper-proposal.md
jlewi: I don't see why requiring cluster admin privileges to manage a Kubeflow deployment is problematic.

The workflow I envision is using GitOps to automatically run admin ops with elevated privileges.

* The Kubeflow configs are checked into source control
* If a datascientist wants to change the deployment they submit a PR making the change
* The PR is reviewed/approved by suitable admin
* After PR is merged, GitOps tooling (e.g. Weave Flux) notices the change and applies the changes to the deployment
    * This tooling runs with suitable cluster admin privileges

proposals/bootstrapper-proposal.md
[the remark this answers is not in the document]
kkasravi: > My mental model is that you need to be an admin to deploy Kubeflow.

Internally, we had talked with a number of data scientists that really liked the idea of not having to ask devops to deploy kubeflow. They wanted a self-serve environment but with RBAC protections. The list of users under an organization would be allowed to deploy kubeflow although the initial namespace creation (and other cluster level resources) would be delegated to a separate process that would first validate the user. The user deploying to a namespace would be able to control resources in that namespace. The user could also share the deployment with other team members. The analogy is linux users - where creating the home directory requires special privileges. Then the new user owns everything under their home directory. That said - this does differ from your mental model. 

> Each Kubeflow deployment should be completely isolated. The implication of what you are suggesting is that each Kubeflow deployment created by a datascientist is sharing some set of resources.

I think it makes sense to have things like PV's shared among deployments - especially if they are pointing to common large datasets. It's also possible to have a PV that only one data-scientist is required to use exclusively. 

> A given Kubeflow deployment could support multiple datascientists and we want to isolate datascientists within a Kubeflow deployment. As an example, I could imagine that when we deploy Kubeflow we have a list of users allowed to use that deployment. Each of those users gets their own namespace that isolates the resources that are specific to that datascientist.

I hadn't considered a mixed namespace model. But a list of users allowed to use/share a deployment is something data scientists wanted. If each data-scientist has a user-based  ServiceAccount, then control of the processes running on a pod could be set by Pod.serviceAccountName="data-scientist-service-account".

proposals/bootstrapper-proposal.md
```diff
@@ -0,0 +1,253 @@
+# Kubeflow Bootstrapping Modifications
+
+## Motivation
+
+### Current Kubeflow bootstrapper requires cluster-admin privileges
+The current bootstrapper generates a kubeflow application into a PersistentVolume and will optionally deploy it to the API-server. The bootstrapper uses ksonnet to generate the kubernetes manifests that define the kubeflow runtime. The manifest generation by ksonnet is ordered across phases:
+- optional creation of a Namespace
```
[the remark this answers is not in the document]
kkasravi: i think other than the namespace and PV creation, RBAC rules can be scoped to the namespace and deployment is possible w/ no cluster level bindings.

proposals/bootstrapper-proposal.md
```diff
@@ -0,0 +1,253 @@
...
+- creation of ClusterRoles and ClusterRoleBindings
+- creation of namespace scoped resources (services, serviceaccounts, ...)
+- creation of IAP/GCP single signon
+
+The bootstrapper is run as cluster-admin although not all phases above require cluster-admin privileges.
+
+### RBAC rules for data scientists aren't utilized
```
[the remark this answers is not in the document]
kkasravi: see below - most of the RBAC discussion is around enabling data-scientists to deployment and run processes within their deployment with minimum RBAC privileges at the cluster level.

proposals/bootstrapper-proposal.md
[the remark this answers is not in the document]
kkasravi: yes i think it was a good choice to have bootstrapper optionally deploy so that other deployment tools could possibly be used. The gist of the proposal is to allow data scientists to deploy w/o cluster-admin or cluster wide privileges and to have data-scientists use their own service accounts to allocate pods.

proposals/bootstrapper-proposal.md
jlewi: I don't think self-service and GitOps are mutually exclusive. 

To get self-serve you just need to be able to automatically decide which changes are safe and which are risky and need to be reviewed by an expert.

The advantage of the GitOps pattern is that it allows us to punt on the hard part of deciding which changes are safe and just rely on manual review by knowledgeable expert.

I think that simplifies things for Kubeflow because we can focus on creating RBAC and other restrictions to things where a review process would be cumbersome.

IIUC the core of this proposal is to allow Kubeflow to be deployed without cluster-admin privileges.

Is this really a viable prospect? Deploying Kubeflow requires setting up custom resources which are cluster scoped objects. You also need to create service accounts (could be both GCP and K8s). More importantly you potentially need to set up things like Ingress, firewall rules, etc...

I'm not sure I understand what it would mean to deploy Kubeflow if you can't perform these admin actions. 

I think we agree that we have  least two different roles (in an IAM sense).

kubeflow-admin
datascientist

I think the question is which privileges should the datascientist role have? 

Its not clear to me why splitting the bootstrapper into two phases and using two different role accounts would have any impact on what privileges we assign to the datascientist role.

So why not just assume bootstrapper runs with cluster-admin privileges and then separate from the boot strapper  think about how we can use RBAC, namespaces, and other features to create one or more suitable roles.

kkasravi: Maybe we're saying the same thing. At least I'm in agreement with just about everything you've noted above. 

So if bootstrapper were to run with cluster-admin privileges but setup ServiceAccounts and PodTemplates (ie the PodTemplate.serviceAccountName) so they run with the data-scientist Role (which we need to define appropriately) and bootstrapper first verifies the data-scientist is on a whitelist or otherwise authorized to 'deploy' then all subsequent actions of the data-scientist in that namespace will run with 'least' privileges. The data-scientist is able to ask the bootstrapper to 'deploy' but bootstrapper will make sure the data-scientist doesn't have RBAC rules that would allow damage the cluster or things beyond the namespace the deployment is created under. 

Is this closer to what you have in mind? I should mention things like increasing resource quotas are a good fit for the gitops style approval and hopefully we can define the data-scientist role so it would prevent modification of resource quotas.

Side Note: It looks like @ankushagarwal already has a PR to do pod presets that inject GCE credentials using a kubeflow account so there may be some similar efforts from Ankush along these lines.

jlewi: Yes, I think we agree we should define appropriate RBAC roles and try to run with least privilege.

I'm a little hung up on bootstrapper. All bootstrapper is doing is calling `ks apply` to create a bunch of K8s resources and (possibly some Cloud resources). Why does the bootstrapper need to do any sort of checking of permissions or whitelisting? Shouldn't this be handled by RBAC and IAM respectively?

Can you list the privileges and actions datascientists should have?

> ServiceAccounts and PodTemplates (ie the PodTemplate.serviceAccountName) so they run with the data-scientist Role

The bootstrapper is deploying microservices (e.g. JupyterHub, Argo controller, TFJob controller, Seldon Controller). Each of these services should run with an appropriate service account and role that is scoped to the privileges they need. These privileges will be different from those needed by datascientists. 

The TFJob controller for example uses the [tf-job-operator](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/tf-job.libsonnet#L171) service account which needs to be able to things like list pods etc...

We should have a separate RBAC role and IAM role corresponding to datascientistits. They will need permission to do things like create TFJobs but the set of permissions will be different from those the controller needs so they will use a different service account.

Similarly actual TFJobs, Argo Workflows, Jupyter Pods, need different permissions from the controller itself (e.g. to access data on behalf of the datascientist).

So I think the only connection between the Kubeflow deployment and RBAC is that  when we deploy Kubeflow we should create appropriate datascientist RBAC and IAM roles and add users/groups to them.

Some open questions/ work items are:

- What RBAC permissions should we assign to the datascientist role?
- Should we change our spec for TFJobs, Argo Workflows (the resources themselves not the controllers) to use this new role by default?
- When users setup Kubeflow how should they assign users to these roles?

    - Ideally this would be handled by external auth systems (e.g. on GCP we would just use Google Groups and assign roles to groups and admins could just add/remove users from these groups)
         - Unfortunately you can't assign [GoogleGroups to role bindings](https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#assigning_permissions_to_users_with_rolebinding)

Pod presets are just a different way of specifying credentials to attach credentials to pods. What we currently do if we need to attach GCS/S3 credentials to a pod is modify the manifest for that resource (e.g. the TFJob or Argo Workflow) to set the appropriate environment variable and secret.  This is fine but leads to complex jsonnet that is hard to read. PodPresets are a way of simplifying this because we can just attach a label to the pod and then define an admission controller that will inject the credentials into the pod when the pod resource is created. This will make it easier to support authenticating to different external systems (GCS vs. S3) without creating really complex jsonnet that accounts for all the different systems.

kkasravi: +1 on your comments above. I think this narrows the scope to a few items

> I'm a little hung up on bootstrapper. All bootstrapper is doing is calling ks apply to create a bunch of K8s resources and (possibly some Cloud resources). Why does the bootstrapper need to do any sort of checking of permissions or whitelisting? Shouldn't this be handled by RBAC and IAM respectively?

Currently a user calling `kubectl create -f bootstrapper.yaml` will need something like the following RBAC in order to create bootstrapper and run it (but not have it deploy):
```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: bootstrapper-role
rules:
- apiGroups: [""]
  resources: ["ClusterRoleBinding", "Namespace", "Service", "StatefulSet", "PersistentVolumeClaim", "PersistentVolume"]
  verbs: ["create"]
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: bootstrapper
subjects:
- kind: Group
  name: kubeflow-admins
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole
  name: bootstrapper-role
  apiGroup: rbac.authorization.k8s.io
---
```

Data scientists would need a weaker RBAC, hopefully scoped to the namespace unless we want to allow datascientists to create their own kubeflow instances via the bootstrapper. In this case they would be adding a new directory to the PersistentVolume. 

As you mention there isn't a mapping from GoogleGroups to role bindings so the auth provider will need to take GoogleGroups and create a Group for data scientists. Since we've been talking about doing authz via the ambassador gateway i think group creation is part of the authz provider functionality and different authz providers will need to do different things (eg github auth provider would traverse an organization or team or repos looking for members). 

I can create an issue with your work items noted above. 
I'm not sure if you wanted a declarative way of specifying users, groups that each provider would use (eg CRD) or just rely on implementing this as part of each auth provider. 

Thoughts?

kkasravi: The ClusterRoles defined are listed here:
https://raw.githubusercontent.com/kubernetes/kubernetes/master/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/testdata/cluster-roles.yaml

So a datascientist rolebinding would aggregate a number of these and might be able to borrow from the existing rolebindings: https://raw.githubusercontent.com/kubernetes/kubernetes/master/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/testdata/cluster-role-bindings.yaml

kkasravi: @jlewi i can also update the proposal and provide more details on the RBACs for kubeflow-admin, datascientist. The mapping of different auth providers to the users/datascientists in kubeflow.

proposals/bootstrapper-proposal.md
jlewi: I think the initial version of the Application CRD is to just provide metadata. i.e. we could create an Application CRD and list all the differen K8s resources that are part of Kubeflow. I think currently application CRD is just intended to serve as a container to make it easy for things like UIs to understand which resources should be grouped together.

I think longer term it might be used for lifecycle management but I'm not sure yet.

Based on the other discussion, it seems like lifecycle management might be different from what you are proposing in this proposal.

proposals/bootstrapper-proposal.md
jlewi: If you want to start opening issues and PRs to add/update RBAC roles I think you can. I suspect the roles and specific permissions will evolve over time. So I think we can use code review (as opposed to a proposal) to work out specific details.

I think it would be great to update this doc with more information about how you think we will manage groups and users. I suspect there is work to be done by the underlying systems (e.g. Kubernetes or authors of individual Authenticator Modules like Google Accounts).  So it would be great to have a doc that we could share with other teams to let them know what we need.

According to the [docs](https://kubernetes.io/docs/admin/authorization/rbac/#referring-to-subjects) it looks like Groups are handled by the Authenticator module.  So can you explain what you meant by doing it via the Authz provider?

Are you suggesting we use Ambassador and Authz as the authentication module for the K8s module?

On GKE I was thinking Ambassador would only be used to authenticate/secure access to web apps running in the cluster; not access to the K8s APIs themselves. GKE already takes care of providing auth to the master and its not clear to me that messing with that would be a good idea (I'm not even sure its configurable on GKE).

I'm less familiar with non-GKE platforms so maybe on non-GKE platforms it makes more sense.

For GKE we would most likely file a feature request against GKE to support binding RBAC roles to Google Groups.

An alternative work around that would support more than GKE and Google Groups would be to create a CRD that would sync Groups to RBAC bindings (e.g. member of the group would be added as an individual subject in the binding).

proposals/bootstrapper-proposal.md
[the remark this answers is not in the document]
jlewi: Seems good to me.

proposals/bootstrapper-proposal.md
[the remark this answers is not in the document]
jlewi: Good point. Although I think there's work to be done to support multiple Kubeflow deployments per cluster.

CRD's are cluster scoped so we can only have 1 CRD per cluster. But CRD's are versioned so in theory all Kubeflow deployments using the same version of a CRD should be using the same CRD spec.

The controllers themselves can be scoped to only act on resources for a specific version and namespace so we could potentially support only claiming resources in a particular namespace. (This is an issue for our E2E tests as well).

There's also the question of ingress and Ambassador. Setting up ingress should probably require admin privileges.  

Would each deployment have its own use case?

It might be useful to try to define concrete user scenarios so we can figure out how to support them.

For example, the scenario we currently handle today is one cluster shared by a team and everyone uses a service account owned by the team to do work.

One use case we don't handle well today is when you have a team of datascientists each with access to different datasets. I think that might be the use case you are getting at because we would need a way to isolate each datascientist.

proposals/bootstrapper-proposal.md
kkasravi: sure, i can open an issue on this independent of the bootstrapper modifications and lifecycle

proposals/bootstrapper-proposal.md
kkasravi: when i said authz provider i meant authentication provider (not authorization provider) though your comments reflect what I meant to say :)

> I think it would be great to update this doc with more information about how you think we will manage groups and users. I suspect there is work to be done by the underlying systems (e.g. Kubernetes or authors of individual Authenticator Modules like Google Accounts). So it would be great to have a doc that we could share with other teams to let them know what we need.

In progress. I'll shoot for mid-week.

> On GKE I was thinking Ambassador would only be used to authenticate/secure access to web apps running in the cluster; not access to the K8s APIs themselves. GKE already takes care of providing auth to the master and its not clear to me that messing with that would be a good idea (I'm not even sure its configurable on GKE).

Yes - GKE doesn't seem to allow setting api-server flags as described in k8's authentication. The ambassador authenticator will need to check if the user's identity is listed as a subject in the RoleBinding that includes the ServiceAccount of the service ambassador is intercepting

> An alternative work around that would support more than GKE and Google Groups would be to create a CRD that would sync Groups to RBAC bindings (e.g. member of the group would be added as an individual subject in the binding).

Yes this is along of the lines of existing CRD's in the proposal. Thanks Jeremy. Let me target this week for both the proposal update and one or more issues with PRs not too far behind.

jlewi: Thanks. Really appreciate your diligence.

It might also be good to setup a zoom call to discuss so we can have a higher bandwidth conversation. If you'll be attending next week's community call (5:30 pm PST) and think it can wait until then we could just use that call. Otherwise we could try to setup a call before then.

kkasravi: the guys I've been working with internally suggested the same thing. Would you be free on Monday for 30-45minutes for a zoom call? Pick your time - but probably mid-morning would be good for us.  Agenda could include the use cases you had in mind as well as feedback we've gotten from data scientists internally at intel. I'll put together a set of questions we have and objectives / goals. I don't want to take more time than necessary given how busy you are but i think a short discussion can have a high return value. Thanks for all the clarifications you've provided thus far in how you envision the bootstrapping and deployment.

proposals/bootstrapper-proposal.md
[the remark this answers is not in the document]
kkasravi: @jlewi 
> Would each deployment have its own use case?

yes

> One use case we don't handle well today is when you have a team of datascientists each with access to different datasets. I think that might be the use case you are getting at because we would need a way to isolate each datascientist.

yes this is one of the use cases. We've had feedback from data scientists that there are common use cases where they may need to limit visibility to specific datasets due to [HIPPA](http://www.onlyhealthy.com/understanding-the-hipaa-law/) or other constraints.

proposals/bootstrapper-proposal.md
kkasravi: Note - i'll update the proposal after our meeting.

jlewi:
Should we close this issue for now due to inactivity?

kkasravi:
@jlewi We are actively working on a variation of this proposal which we will open an issue on. It includes Application, KfApp, KfCluster CRDs (the latter 2 being new). It addresses the isolation/self-serve environment that we feel is a requirement for data-scientists. We can close this since most of the proposal no longer applies.

Issue #136, opened 2018-05-22
[Request] MXNet operator support


gaocegege:
Now we support TensorFlow and pytorch well. MXNet is another popular ML framework and I think we should implement the operator for it to attract more DL practitioners.

We need to create a proposal like https://github.com/kubeflow/community/blob/master/proposals/pytorch-operator-proposal.md and create the repository for the operator.

### References

- https://mxnet.incubator.apache.org/faq/distributed_training.html
- https://github.com/apache/incubator-mxnet/blob/master/example/gluon/image_classification.py

/cc @brucechin

suleisl2000:
[TuSimple](http://www.tusimple.com) have just written a draft version of [MXNet operator](https://github.com/TuSimple/mxnet-operator). We'd like to contribute it to Kubeflow community as part of ecosystem and continuously improve it in the future. In addition to better deep learning lifecycle management, we plan to integrate the operator with kube-arbitrator for better job scheduling. Any process I have to follow up to make it happen? @jlewi 

BTW, TuSimple is one of the major contributors in [MXNet community](http://mxnet.incubator.apache.org/community/powered_by.html).

/cc @k82cn @jzp1025 @jjmtraveller

jlewi:
@suleisl2000 that's a fantastic offer.

Would you and TuSimple be willing to continue supporting the operator and making the changes (see below) to better integrate it into Kubeflow? 

Assuming we go forward integrating this into Kubeflow there's a couple things we should do

1. Add a ksonnet package see [here](https://github.com/kubeflow/kubeflow/blob/master/docs_dev/ksonnet_packages.md)
1. Add an mxnet guide to our website [here](https://github.com/kubeflow/website/tree/master/content/docs/guides/components
1. Figure out where the code should live
1. Figure out what changes are needed to provide an API consistent with other operators see kubeflow/pytorch-operator#49

@suleisl2000 @johnugeorge is currently working on refactoring the PyTorchJob and TFJob operators so that they can share implementation; see for example kubeflow/tf-operator#773. The current thinking is that we should have a different CRD for each type of job, but the underlying implementation should be shared.

As part of this there is ongoing discussion about whether we should use separate repos or move all the 
code into a single operator.

@gaocegege @johnugeorge what do you think? Should we put the code in its own repository or should we add it to tf-operator to make it easier to start refactoring to use a shared implementation?

johnugeorge:
[...]

PR #137, opened 2018-05-22
propopsal: Add MXNet

gaocegege:
/assign @jlewi 

Ref #136 

/cc @brucechin

@brucechin is interested in implementing mxnet operator for kubeflow community

Signed-off-by: Ce Gao <gaoce@caicloud.io>

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/137)

### DIFF
--- a/proposals/mxnet-operator-proposal.md
+++ b/proposals/mxnet-operator-proposal.md
@@ -1,0 +1,103 @@
+_Authors:_
+
+- [@gaocegege][] - Ce Gao &lt;gaoce@caicloud.ioo&gt;
+
+## Motivation
+
+Apache MXNet is a popular machine learning framework which currently does not have an operator for Kubernetes. This proposal is aimed at defining what that operator should look like, and adding it to Kubeflow.
+
+## Goals
+
+A Kubeflow user should be able to run training using MXNet as easily as then can using TensorFlow. This proposal is centered around a Kubernetes operator for MXNet. A user should be able to run both single node and distributed training jobs with MXNet.
+
+This proposal defines the following:
+
+- A MXNet operator
+- A way to deploy the operator with ksonnet
+- A single pod MXNet example
+- A distributed MXNet example
+
+## Non-Goals
+
+For the scope of this proposal, we won't be addressing the method for serving the model and the [UI](https://github.com/awslabs/mxboard).
+
+## API (CRD and resulting objects)
+
+The custom resource submitted to the Kubernetes API would look something like this:
+
+```yaml
+apiVersion: "kubeflow.org/v1alpha1"
+kind: "MXNetJob"
+metadata:
+  name: "example"
+spec:
+  replicaSpecs:
+    worker:
+      replicas: 4
+      restartPolicy: Never
+      template:
+        spec:
+          containers:
+            - name: mxnet
+              image: kubeflow/dog-food-for-mxnet:1.0
+    server:
+      replicas: 2
+      restartPolicy: Never
+      template:
+        spec:
+          containers:
+            - name: mxnet
+              image: kubeflow/dog-food-for-mxnet:1.0
+    scheduler:
+      restartPolicy: Never
+      template:
+        spec:
+          containers:
+            - name: mxnet
+              image: kubeflow/dog-food-for-mxnet:1.0
+```
+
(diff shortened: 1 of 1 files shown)


jlewi:
@brucechin are you the only one that's going to be working on the MXNet operator?

gaocegege:
I am not sure if he is the only one, I will help him but have to give more attention to tf-operator.

brucechin:
yeah， I'm the only one for mx operator now in my group. maybe more in the summer holiday.

jlewi:
Is there a reason to begin working on the mxnet operator now? @brucechin do you have a specific need for mxnet or are you just looking to get involved with Kubeflow?

We already have 3 operators in progress

- TfJob
- PyTorch
- Caffe2

From a community perspective I think it would be preferable to invest more resources in getting the existing operators done before taking on more operators.

Building these operators is proving to be fairly expensive. The TFJob operator has been in development for almost a year and had significant contributions from over 6 people.

Ideally, if we do want to support more operators, I think we might want to consider how to build shared libraries and reuse code to reduce the cost of supporting additional frameworks.

brucechin:
Algorithm group in my company(Tusimple) only uses MXNet, so we would like to add support for MXNet in Kubeflow. And it is important to reuse code to reduce the cost of supporting other frameworks.

gaocegege:
Personally, I agree with @jlewi . Maintaining these operators is a heavy work and I prefer to make one of these operators production ready then we could build more based on the experience we got from the previous work.

But TuSimple is using MXNet now and they are investigating using kubernetes to manage ML workloads. Then I think it also works to create a MXNet operator with the help of people in TuSimple.

WDYT?

BTW, @brucechin Do you have a plan when to use it? Maybe we could focus on tf-operator now and after the release of v0.2 I can help you to build the MXNet operator.

jlewi:
I agree. If there is someone with a compelling need for a framework then it makes sense to support them.

Is there a way we can start trying to refactor TFJob operator code into reusable libraries to avoid duplicating the same logic for all the different operators?

gaocegege:
@jlewi 

I think it is hard to do. There are some operator frameworks to help programmers to create a new operator, while the code reuse is still be a problem.

Maybe we could maintain a sample controller similar to https://github.com/kubernetes/sample-controller then every operators could fork it. With such a sample controller we could make all operators in the similar architecture.

jlewi:
I don't think there's any downside to incubating an mxnet operator.

What do others think?

brucechin:
@gaocegege we are not in a hurry to deploy it on kubenetes now. I just need to figure out how to combine MXNet and Kubenetes in advance.

jlewi:
@brucechin What do you think the next step should be? If you're ready to start work on an MXNET operator, I'm happy to LGTM this proposal and give you a repository to begin work.

ankushagarwal:
I have been experimenting around with mxnet-operator here https://github.com/ankushagarwal/mxnet-operator/

Generally speaking, I am finding a lot of overlap b/w the various distributed training operators tensorflow, pytorch, mxnet, caffe2, etc. Maintaining a separate controller for each of the operators is a lot of work - and most of this work is shared b/w various operators. Some examples: PodCleanupPolicy, PodRestartPolicy, CRDValidation, Releasing, E2E Testing etc. Other than that, we are just bringing managing pods with special environment variables set.

I am currently writing a proposal which tries to abstract out the details of a specific framework and provides a single operator for all frameworks' distributed training jobs - tensorflow, pytorch, mxnet, caffe2.

gaocegege:
@ankushagarwal 

Awesome idea. I am looking forward to your proposal. We have lots of operators. If we could maintain one for all, it will be better. I had the same idea before, while I find that caffe2 requires backend and does not has the similar pattern compared with tf.

jlewi:
@brucechin @gaocegege What's the status of this? Is anyone working on mxnet?

wsxiaozhang:
A MXNet operator would be good before we having some generic pattern operators like PSOperator, MPIOperator, etc.  I know someone is planning to contribute a MXNet operator. I'll introduce him here.

gaocegege:
@jlewi 
Could we merge the PR or close it? MXNet operator is existed now and upgrading to v1alpha2.

jlewi:
I think we should close it rather than merge it because I don't know how relevant the technical details are e.g. I think @suleisl2000 already started changing the APi to be compatible with our v1alpha2 APIs. 

My suggestion would be to close it and then at some point we can create a more up to date doc describing the actual implementation.

@gaocegege  @suleisl2000 WDYT?

gaocegege:
SGTM

PR #139, opened 2018-05-23
Add MPI Operator proposal.

rongou:
This is something we've been working on for a while at NVIDIA. Beyond simple benchmarking, we've used it to train real-world models for the self-driving car project, both on kops clusters on aws and on-prem bare metal clusters. So far seems to work pretty well, making it pretty easy to scale up and down the number of GPUs for training. Hope the kubeflow community also finds this useful.

@jlewi @alsrgv @jiezhang

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/139)


jlewi:
Looks great to me.

/lgtm
/hold so others working on mpi have a chance to comment.

aronchick:
+1 this looks fantastic!

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+1.  Create the launcher Job. It’s run under the `ServiceAccount` created in
+    step 2. Environment variables are set to tell `mpirun` to use our helper
+    script for remote execution, and the `hostfile` to use. The `kubectl` binary
+    is delivered to an `emptyDir` volume through an init container.
+1.  After the launcher job finishes, set the `replicas` to 0 in the worker
+    `StatefulSet`.
+
+![MPI Operator](diagrams/mpi-operator.png)
+
+It may be desirable to schedule all the GPUs in a single Kubernetes resource
```
alsrgv: That sounds like a good default option.

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+apiVersion: kubeflow.org/v1alpha1
+kind: MPIJob
+metadata:
+  name: tensorflow-benchmarks
+spec:
+  gpus: 128
+  template:
+    spec:
+      containers:
+      - name: tensorflow-benchmarks
```
alsrgv: Are users able to define sidecar containers that should run on workers?  Sidecars may need to share filesystem (upload snapshots) and/or run in privileged mode (set up `tc qdisc` rules).

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+        image: rongou/tensorflow_benchmarks:latest
+```
```
alsrgv: Can you add example `mpirun` command here?  Can it be templated to automatically substitute # of allocated GPUs?

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+  template:
+    metadata:
+      labels:
+        app: ${job-id}-worker
+    spec:
+      containers:
+      - name: tensorflow-benchmarks
+        image: rongou/tensorflow_benchmarks:latest
+        command: ["sleep"]
+        args: ["365d"]
```
alsrgv: You can do "infinity" :-)

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+Compared to parameter servers, this approach arguably is more
+[bandwidth optimal](http://www.cs.fsu.edu/~xyuan/paper/09jpdc.pdf) and
+[scales better](https://eng.uber.com/horovod/).
+
+However, right now there is no easy way to launch this type of training jobs on
+Kubernetes. By providing a CRD and a custom controller, we can make
+allreduce-style distributed training as simple as training on a single node.
+
+## Goals
+* Provide a common Custom Resource Definition (CRD) for defining a single-gpu,
```
galv: I may be missing something, but what is the value of a single-gpu training job using all-reduce? I'm not even sure nccl would run in that case because forming a singleton ring is rather odd.

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+multi-gpu, or multi-node training job.
```
galv: I'm guessing "multi-gpu" here means multiple gpus on on node. I don't really know best practices in kubernetes (I'm new here!), but why not use a single container for this with all the GPU devices you want to use whitelisted?

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+* Implement a custom controller to manage the CRD, create dependent resources,
+and reconcile the desired states.
+* The cross-pod communication should be secure, without granting unnecessary
+permissions to any pod.
+* Though the initial version focuses on TensorFlow/Horovod, the approach can be
+applied to other frameworks with MPI support, such as CNTK.
```
galv: There is something that I am missing here.

It sounds like you could be doing multiple things here:

1) Allow using horovod in kubeflow (the simplest)
2) Allow running MPI jobs within kubeflow. In this case, how do we handle possible race conditions between both NCCL and MPI creating a ring for the computations?

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+In theory, this operator can be used to run arbitrary MPI jobs (e.g.
+computational fluid dynamics), but it's not our focus.
+
+## API
+
+### Container Image
+
+A requirement on user's container image is that it includes an MPI
+implementation. The initial version is tested on Open MPI 3.0.0, but it can be
+extended to other MPI implementations if there is a need.
```
galv: Could someone from the community comment on why Open MPI < 3.x would be valuable? I am not familiar.

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+We create a new custom controller that listens for `MPIJob` resources. When a
+new `MPIJob` is created, the controller goes through the following *logical*
+steps:
+1.  Create a `ConfigMap` that contains:
+    *   A helper shell script that can be used by `mpirun` in place of ssh. It
+        invokes `kubectl exec` for remote execution.
+    *   A `hostfile` that lists the pods in the worker `StatefulSet` (in the
+        form of `${job-id}-worker-0`, `${job-id}-worker-1`, ...), and the
+        available slots (GPUs) in each pod.
+1.  Create the RBAC resources (`Role`, `ServiceAccount`, `RoleBinding`) to allow
```
galv: typo: you are doing "1." "1." "1." "1." here.

proposals/mpi-operator-proposal.md
alsrgv: It's important to support a single-GPU use case as well to minimize friction between doing a single-GPU and multi-GPU training.  In case of a single-GPU Horovod optimizes away allreduce and so there's no performance penalty.

proposals/mpi-operator-proposal.md
alsrgv: Horovod uses MPI and handles NCCL ring creation - so there will be no race condition.

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+          - key: kubexec.sh
+            path: kubexec.sh
+            mode: 365
+          - key: hostfile
+            path: hostfile
+            mode: 292
+```
+
+The launcher pod invokes `mpirun` and communicates with worker pods through MPI.
+The initial handshake is done through `kubectl exec` instead of SSH. Logs can
```
everpeace: SSH-less(`kubectl exec`-based) mpi is fantastic 👍 !! I really love it!!

- It perhaps similar to @alsrgv pointed out, but, it would have to detect proper kubernetes version from the conatiner(side-car container?) because kubernetes cluster woud be upgraded in time. 
- Would kubernetes api server be overloaded if large openmpi jobs are spawned and communicated heavily each other??

proposals/mpi-operator-proposal.md
everpeace: @galv I think yes.  OpenMPI 3.0.x has a bug on GPUDirect  (I actually use this feature).

ref: https://github.com/open-mpi/ompi/issues/3972

NOTE:  infiniband cannot be scheduled by default, but it can be used via very thin device plugin which just exposes device files in a host to the containers.

everpeace: oh, I noticed OpenMPI v3.0.2(not released yet) will fix the issue !! 👍

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+new `MPIJob` is created, the controller goes through the following *logical*
+steps:
+1.  Create a `ConfigMap` that contains:
+    *   A helper shell script that can be used by `mpirun` in place of ssh. It
+        invokes `kubectl exec` for remote execution.
+    *   A `hostfile` that lists the pods in the worker `StatefulSet` (in the
+        form of `${job-id}-worker-0`, `${job-id}-worker-1`, ...), and the
+        available slots (GPUs) in each pod.
+1.  Create the RBAC resources (`Role`, `ServiceAccount`, `RoleBinding`) to allow
+    remote execution (`pods/exec`).
```
everpeace: So all the pod names(`-0`, ... , `-(N-1)`) will be whitelisted in `resourceNames` attribute in the role binding??  Could we utilize [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) too?

proposals/mpi-operator-proposal.md
galv: @alsrgv First of all, it is my understanding that you are the Horovod maintainer. Great to have you.

Maybe race condition is the wrong word. What I mean is that I am not sure what will happen if you use both CUDA-aware MPI and NCCL (i.e., you make a communication clique in both nccl and MPI) at the same time. (I was never really sure why you depend on both MPI and NCCL in Horovod, actually, but never took a deep enough dive into the code to understand why. Sumimasen). I mention this because this documentation sprinkles the word nccl around in a few places despite being primarily about MPI, so I wanted to point it out.

gaocegege:
/lgtm

Thanks for your contribution!

proposals/mpi-operator-proposal.md
gaocegege: I think it will be supported since this is a pod template spec.

kinglion811:
I  am very interested in it，recently，I run lightlda distribution，It‘s manage by mpi，I think it will very usefull

proposals/mpi-operator-proposal.md
kinglion811: Yes，I think the pain points is ssh,if can SSH-less it will very fantastic

proposals/mpi-operator-proposal.md
alsrgv: Yes, I am - thanks :-)

Horovod uses NCCL because it's much faster than CUDA-aware MPI.  However, it still uses MPI for control communications between chief worker and everyone else - workers report which tensors they're ready to reduce and chief commands the order that is suitable for all.  This is necessary for TensorFlow since it has non-deterministic operation execution order in presence of forks in tue graph.

NCCL and MPI on CPU buffers will not deadlock.

proposals/mpi-operator-proposal.md
```diff
@@ -0,0 +1,224 @@
...
+multi-node training using TensorFlow; the TensorFlow team is also working on an
+`allreduce`
+[implementation](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/distribute).
+Compared to parameter servers, this approach arguably is more
+[bandwidth optimal](http://www.cs.fsu.edu/~xyuan/paper/09jpdc.pdf) and
+[scales better](https://eng.uber.com/horovod/).
+
+However, right now there is no easy way to launch this type of training jobs on
+Kubernetes. By providing a CRD and a custom controller, we can make
+allreduce-style distributed training as simple as training on a single node.
```
ankushagarwal: Can you link to some docs explaining what `allreduce` is?

proposals/mpi-operator-proposal.md
rongou: Do you all think this should be the only option? If no one ever wants to allocate GPUs to the launcher, then we can just get rid of that mode, to keep things simple.

proposals/mpi-operator-proposal.md
rongou: Yes as @gaocegege said.

proposals/mpi-operator-proposal.md
rongou: OK, I added the `mpirun` command override in the "flexible" version down below. In the hostfile we create the number of slots are set to the number of GPUs requested, so you don't need to worry about # GPUs if you just do `-map-by slot`.

proposals/mpi-operator-proposal.md
rongou: Yes exactly. You can do your debugging and development on a single-gpu job. When you're ready to train the model for real, just change a single number to scale it up.

proposals/mpi-operator-proposal.md
rongou: That's exactly what we are doing. When you have 8-GPU nodes and run an 8-GPU job, we launch a single pod and rely on mpi/horovod/nccl to parallelize the training across the 8 GPUs (internally `mpirun` forks a new process for each GPU). The basic idea is a single interface can be used to train on as many GPUs as you have available, regardless of node boundaries.

proposals/mpi-operator-proposal.md
rongou: @everpeace added link to ChainerMN.

proposals/mpi-operator-proposal.md
rongou: This is just markdown "lazy" numbered list. Github shows it as the default syntax: https://guides.github.com/features/mastering-markdown/#syntax

jlewi:
There seems to be overwhelming consensus that this is a great contribution.

@rongou any changes you want to make before committing the initial version? If not just cancel the hold.
/approve

proposals/mpi-operator-proposal.md
rongou: The `kubectl` binary is really only used for the `pods/exec` api call, so version skew might not be too big of an issue. We could write a standalone go binary, but it'd be replicating the `kubectl` functionality, so I'm not sure if it's worth it. If someone can figure out how to call `pods/exec` from a shell script, maybe we can switch to that. :)

proposals/mpi-operator-proposal.md
rongou: Yes the `Role` is restricted to the specific `resourceNames`. Haven't looked at network policies yet, but it should be pretty easy to add.

proposals/mpi-operator-proposal.md
rongou: Linked to a survey paper that goes into more detail than you want to know about distributed deep learning: https://arxiv.org/abs/1802.09941.

rongou:
@jlewi you are too fast. :) I have some minor edits to address review comments. Will send you another PR.

proposals/mpi-operator-proposal.md
alsrgv: I think it could be an only option.

proposals/mpi-operator-proposal.md
galv: This is reasonable, it's a good design choice, makes things simpler.

PR #145, opened 2018-05-30
Add chainer-operator proposal.

everpeace:
I've been working on k8s operator for [Chainer](https://chainer.org/)/[ChainerMN](https://github.com/chainer/chainermn) to make spawning jobs with chainer/chainermn easier in k8s cluster. (I'm now prototyping the operator.)

Thus, I would like to propose chainer-operator in this community.   And, I want to say thanks to @aronchick and @jlewi because they encouraged me to post this proposal to the community via private email communication.  I'm very happy to accelerate my contribution to kubeflow community via this proposal and introducing chainer-operator.   

I would be glad if I could have feedbacks.

disclosure:  I'm working for [Preferred Networks, Inc.](https://www.preferred-networks.jp/en/) which is the company leading development of [Chainer](https://chainer.org/)/[ChainerMN](https://github.com/chainer/chainermn).

/assign @jlewi @aronchick

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/145)

### DIFF
--- a/proposals/chainer-operator-proposal.md
+++ b/proposals/chainer-operator-proposal.md
@@ -1,0 +1,222 @@
+## Motivation
+
+[Chainer][Chainer] is a Python-based, standalone open source framework for deep learning models. Chainer provides a flexible, intuitive, and high-performance means of implementing a full range of deep learning models, including state-of-the-art models such as recurrent neural networks and variational autoencoders.  
+
+[ChainerMN][ChainerMN] is an additional package for [Chainer][Chainer], which enables multi-node distributed deep learning in a scalable, flexible and easy way.  [ChainerMN][ChainerMN] currently supports MPI to initialize process groups or do collective communications(e.g. broadcast, all-reduce, etc.) among processes attending the distributed learning.  They are now planning to extend the support to other communication backends (e.g. [gloo][gloo] or other custom ones).
+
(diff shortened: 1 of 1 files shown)


jlewi:
This looks great to me 

@everpeace are you and Preferred Networks planning on taking the lead on the operator and responsibility for driving it to beta and then v1. 

My main concern is that the community is already spread pretty thin across the operators/components under development. If you are willing to take the lead on this then that addresses my concern.

Does someone else from the community want to LGTM this as well?

/cc @gaocegege @rongou @johnugeorge @brucechin

proposals/chainer-operator-proposal.md
```diff
@@ -0,0 +1,222 @@
...
+apiVersion: kubeflow.org/v1alpha1
+kind: ChainerJob
+metadata:
+  name: example-chainermn-job
+spec:
+  # only "mpi" is supported in the first scope
+  # "gloo", or custom backend will be supported in the future.
+  backend: mpi
+  replicaSpecs:
+  - replicas: 1
```
gaocegege: Could we use map instead of slice as tf-operator does? Ref https://github.com/kubeflow/community/blob/master/proposals/tf-operator-design-v1alpha2.md

It will ease the implementation of the operator.

everpeace:
@jlewi 

> @everpeace are you and Preferred Networks planning on taking the lead on the operator and responsibility for driving it to beta and then v1.

I'm very honored that I could lead the development.  Yes, I'm very happy to take the lead on the development of chainer-operator.  Preferred Networks and Chainer team understands the value of this development and supports my activity.  Some colleague will join the development of chainer-operator.

>  If you are willing to take the lead on this then that addresses my concern.

I understood your concern.  I hope my above statement addresses your concern.

proposals/chainer-operator-proposal.md
everpeace: done in https://github.com/kubeflow/community/pull/145/commits/9dd0bb3b2f39c02d550e3ec013b8a32b52212e84

jlewi:
@everpeace Yes thank you very much.

/lgtm
/approve

If you need a repository just file an issue on github/community and I will create one.

rongou:
I think we are still in the early stage of figuring out how to support all the frameworks effectively, so it's probably ok to experiment with different operators. Longer term it might make more sense to converge to fewer if not a single operator(s).

gaocegege:
I think custom apiserver may be an awesome solution from the longer term.

PR #175, opened 2018-09-03
add jzp1025 into members.yaml

jzp1025:
work in mxnet-operator now

### DIFF
--- a/members.yaml
+++ b/members.yaml
@@ -226,3 +226,10 @@
   LastName: Su
   affiliation: Tusimple
   email: lei.su@tusimple.com
+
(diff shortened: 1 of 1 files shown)


members.yaml
```diff
@@ -226,3 +226,10 @@
   LastName: Su
   affiliation: Tusimple
   email: lei.su@tusimple.com
+
+- github: jzp1025
```
jlewi: Can you sort this file by github id please? it looks like jinxingwang and suleisl2000 are also out of place.

jzp1025:
OK , I have sorted the file by github id

PR #193, opened 2018-10-02
Add yangpa as Kubeflow member

IronPan:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/193)

### DIFF
--- a/members.yaml
+++ b/members.yaml
@@ -136,13 +136,6 @@
   affiliation: Cisco
   email: anubhgar@cisco.com
 
-- github: ironpan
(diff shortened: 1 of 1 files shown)


members.yaml
```diff
@@ -135,6 +135,13 @@
   lastName: Garg
   affiliation: Cisco
   email: anubhgar@cisco.com
+
+- github: ironpan
```
gaocegege: Please put your name after inc0 since the file is in alphabetical order

IronPan: done

Issue #188, opened 2018-09-19
[Proposal] Make external SQL-compatible database requirement for Kubeflow


inc0:
# SQL database as requirement to Kubeflow

Currently only one component of Kubeflow requires SQL, and that's Katib. In various discussion we determined that another planned component, experiment tracking, will also require some sort of proper database (due to scale and requirement for advanced querying).

Running database in Kubernetes is very hard to maintain and setup properly, we shouldn't be in business of writing operation code for databases. Currently in Katib we deploy simplistic mysql instance (single node, lack of persistence), which isn't viable for production use cases. Building code that will manage fully clustered, persistent and stateful relational database is very high and outside of scope of Kubeflow.

During discussions we determined that external database is reasonable requirement:
* Every major public cloud already have managed database offering
* On-prem can either deploy SQL compatible database on their own (inside or outside Kubernetes, for example with Vitess)

I'd like to propose that we will make SQL database hard requirement for those components of Kubeflow which require it, and make sure that configuration of it is uniform across project.

## Additional cost to deployment

Adding big requirement like that shouldn't be taken lightly. Kubeflow already struggles with deployment experience, this will further require deployers (especially first time deployers) to make few additional steps to it. This problem can be partially solved by adding good documentation for all public clouds + bare metal. Besides docs we could also add small yaml manifest to deploy PoC database, that's not production ready but is enough to try out Kubeflow.

## Uniform configuration

I think it should be requirement across project that all database configs should be handled in same way. One example approach would be to ask users to manually create Kubernetes secret and set it's name as parameter (with default being {kubeflow component name}-database). For example:

```
apiVersion: v1
kind: Secret
metadata:
  name: katib-database
type: Opaque
data:
  database_username: YWRtaW4=
  database_password: MWYyZDFlMmU2N2Rm
  database_name: ZGF0YWJhc2VfbmFtZQo=
  database_host: MTkyLjE2OC4xLjE6MzMwNgo=
```

After that we can mount secret values as envs to pods and use them in code.

## Note about schema upgrades

Since we'll assume existing relational database, we will need to manage schema migrations. This can be handled by projects like alembic in python. Seamless upgrades will be hard requirement for production projects. We will also need to take account that certain types of schema upgrades are very expensive computationally and can lock database for hours at the time, these should be avoid. Each schema migration should be detailed in release notes including warning if it's one of problematic ones.

We also should build CI infra that will test:
* Deploy version X of Kubeflow and fill it with fake data
[...]

PR #195, opened 2018-10-03
[WIP] Experiment tracking proposal

inc0:
This PR is supposed to start proper design discussion regarding
experiment tracking. Please, feel free to review, comment or commit new
changes.

Relevant conversations:
https://github.com/kubeflow/kubeflow/issues/264
https://github.com/kubeflow/kubeflow/issues/136

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/195)

### DIFF
--- a/proposals/experiment-tracking-proposal.md
+++ b/proposals/experiment-tracking-proposal.md
@@ -68,6 +68,7 @@ It's very important to provide good UI that allows easy model navigation and mos
 * Tensorboard - Wasn't meant for large number of models. It's better for very detailed examination of smaller number of models. Uses tf.Event files
 * MLFlow - One of big cons for this is using files as storage for models. That would require something like dask or spark to query them efficiently. Can store files in multiple backends (S3 and GCS among other things)
 * ModelDB - Requires mongodb which is problematic
+* StudioML - Also uses FS/object store as backend, which have same querying considerations as MLFlow
 
 All 3 have some subset of features, but none of them seems to be designed for scale we're aiming for. They don't have integration with Kubeflow as well.
 
--- a/proposals/experiment-tracking-proposal.md
+++ b/proposals/experiment-tracking-proposal.md
@@ -65,9 +65,9 @@ It's very important to provide good UI that allows easy model navigation and mos
 
 ## Alternatives
 
-* Tensorboard
-* MLFlow
-* ModelDB
+* Tensorboard - Wasn't meant for large number of models. It's better for very detailed examination of smaller number of models. Uses tf.Event files
+* MLFlow - One of big cons for this is using files as storage for models. That would require something like dask or spark to query them efficiently. Can store files in multiple backends (S3 and GCS among other things)
+* ModelDB - Requires mongodb which is problematic
 
 All 3 have some subset of features, but none of them seems to be designed for scale we're aiming for. They don't have integration with Kubeflow as well.
 
--- a/proposals/experiment-tracking-proposal.md
+++ b/proposals/experiment-tracking-proposal.md
@@ -10,6 +10,13 @@ If training jobs are generated automatically (for example via hyperparameter tun
 thousands of models. It's important to be able to navigate this, select those with best performance, examine them in detail and setup inference cluster
 out of them. We need to track things like model location (on S3, GS or disk), model metrics (end accuracy, P1 score, whatever experiment requires) or logs location.
 
+## Example user stories
+
+* I'm a data scientist working on a problem. I'm looking for easy way to compare multiple training jobs with multiple sets of hyperparameters. I would like to be able to select top 5 jobs measured with P1 score and examine which model architecture, hyperparameters, dataset and initial state contributed to this score. I would want to compare these 5 together in highly detailed way (for example via tensorboard). I would like rich UI to navigate models without need to interact with infrastructure.
+* I'm part of big ML team in company. Our whole team works on single problem (for example search) and every person builds their models. I'd like to be able to compare my models with others. I want to be safe that nobody will accidentally delete model I'm working on.
+* I'm cloud operator in ML team. I would like to take current production model (architecture+hyperparams+training state) and retrain it with new data as it becomes available. I would want to run suite of tests and determine if new model performs better. If it does, I'd like to spawn tf-serving (or seldon) cluster and perform rolling upgrade to new model.
+* I'm part of highly sophisticated ML team. I'd like to automate retraining->testing->rollout for models so they can be upgraded nightly without supervision. 
+
 ## Scale considerations
 
 We need to support scale in tens of thousands of models. Potentially adding garbage collection above this. Within this scale we need to be able to quickly select best models
--- a/proposals/experiment-tracking-proposal.md
+++ b/proposals/experiment-tracking-proposal.md
@@ -1,0 +1,71 @@
+# Experiment tracking
+
+This document is design proposal for new service within Kubeflow - experiment tracking. Need for tool like this was
+expressed in multiple issues and discussions.
+
+## What is experiment tracking
+
+Production machine learning systems can generate huge amount of models. Every training pass can produce at least one, potentially multiple models.
+If training jobs are generated automatically (for example via hyperparameter tuning or retraining as new data becomes available) this can quickly become
+thousands of models. It's important to be able to navigate this, select those with best performance, examine them in detail and setup inference cluster
+out of them. We need to track things like model location (on S3, GS or disk), model metrics (end accuracy, P1 score, whatever experiment requires) or logs location.
+
+## Scale considerations
+
+We need to support scale in tens of thousands of models. Potentially adding garbage collection above this. Within this scale we need to be able to quickly select best models
+for particular problem.
+
+## Model provenance
+
+Another feature commonly asked for is model provenance. It's crucial to be able to reproduce results. For every model we need to record:
+
+* Inital state, whether it's random weights or based on preexisting models
+* Dataset used for training
+* Dataset used for testing
+* Feature engineering pipeline used
+* Model architecture (code used)
+* Hyperparameters
+
+Part of it can be solved by integration with Pachyderm.
+
+## Model performance
+
+To be able to pick best model for problem, we need to record metrics. Metrics can differ problem to problem, but we can support single number as quality weight
+(user can define this number per experiment, whether it's accuracy, p1 score etc). We need to support very efficient queries using this metric.
+
+## Model introspection
+
+For selected models we should be able to setup model introspection tools, like Tensorboard.
+Tensorboard provides good utility, allows comparison of few models and recently it was announced that it will integrate with pytorch. I think it's reasonable to use Tensorboard
+for this problem and allow easy spawn of tensorboard instance for selected models. We might need to find alternative for scikit-learn.
+
+## Inference cluster setup
+
+For best model, we should easily spawn inference cluster. We should support tf-serving and Seldon.
+
+## Integration with TFJob and Katib
+
+It should be very easy, even automatic, to make entry to experiment tracking from TFJob. TF operator should be tightly integrated with it
+and Katib should be able to both read models and write new ones.
+
+## UI
+
+It's very important to provide good UI that allows easy model navigation and most of features listed via button click:
+
+* Examine in tensorboard
+* Train with new data
+* Spawn inference cluster
+* Run tests
+* Show model provenance pipeline
+
+## Alternatives
+
+* Tensorboard
+* MLFlow
+* ModelDB
+
+All 3 have some subset of features, but none of them seems to be designed for scale we're aiming for. They don't have integration with Kubeflow as well.
+
+## Authors
+
+* inc0
--- a/proposals/experiment-tracking-proposal.md
+++ b/proposals/experiment-tracking-proposal.md
@@ -23,6 +23,7 @@ Another feature commonly asked for is model provenance. It's crucial to be able 
 * Dataset used for training
 * Dataset used for testing
 * Feature engineering pipeline used
+* Katib study id
 * Model architecture (code used)
 * Hyperparameters
 
@@ -47,6 +48,10 @@ For best model, we should easily spawn inference cluster. We should support tf-s
 
 It should be very easy, even automatic, to make entry to experiment tracking from TFJob. TF operator should be tightly integrated with it
 and Katib should be able to both read models and write new ones.
+
+Katib workflow could look like this:
+
+Get study hyperparameter space -> select all existing model for study_id -> find out which hyperparameter combination is missing -> create relevant training jobs and add records to experiment tracking.
 
 ## UI
 


jlewi:
Thanks for doing this.

Can your provide more information about the alternative solutions?
Do they provide APIs?
What do they use for DBs?
What are the DB schemas?

What does Katib need? 
Some background on how we are using ModelDB and what problems we are looking to address?

inc0:
I'll slowly do research, but could use help if anyone have experience with any of alternatives. Also, let me know if you know any alternatives that are missing. I'll try to dig more into each architecture, but at the end of the day experience is priceless.

/cc @holdenk - Holden, maybe you could help us with MLFlow? Or point us to someone who knows more about this project?

As for Katib, I'm not sure what Katib really needs aside from experiment tracking. @YujiOshima could you help us figure out what requirements Katib has from it and if ModelDB is enough or we need something more?

YujiOshima:
@inc0 @jlewi In Katib, the long term DB for experimental tracking is completely separated from katib DB.
But experimental tracking is extremely needed since users want to check and evaluate the hyper-parameters Katib generated by their eyes.
So requests from Katib is not so unique.
* Storing Metrics, Hyper-Parameters, DataSet-path, Model-path etc.
* Sort, filter models by Metrics or Hyper-Parameters.

Above requests are needed to do by both of GUI and API.
Katib uses ModelDB and it is OK now. But since it is not active, I want to look for alternatives.

There is many choices(MLFlow, StudioML..) and the best choice is depending on the user.
I try to do in Katib, abstracting the API and make pluggable the backend.
Users or other projects only need to know one API(grpc). They can use ModelManagement tool they want and switch it easily.

YujiOshima:
@inc0 In this proposal, you plan to make a new tool for model management, experimental tracking?
If so, I'm so interested and happy to contribute. 
But the GUI is included? GUI is so important.

johnugeorge:
@YujiOshima I agree. I feel that we have to first list down missing features/requirements in the current tools and then take a call on whether to support the existing ones or implement a new tool.

proposals/experiment-tracking-proposal.md
```diff
@@ -0,0 +1,77 @@
...
+* Train with new data
+* Spawn inference cluster
+* Run tests
+* Show model provenance pipeline
+
+## Alternatives
+
+* Tensorboard - Wasn't meant for large number of models. It's better for very detailed examination of smaller number of models. Uses tf.Event files
+* MLFlow - One of big cons for this is using files as storage for models. That would require something like dask or spark to query them efficiently. Can store files in multiple backends (S3 and GCS among other things)
+* ModelDB - Requires mongodb which is problematic
```
YujiOshima: Not only mongo but sqllite. And it reset the sqlite at the beginning of a process.
We can't persistent data without modification.

proposals/experiment-tracking-proposal.md
```diff
@@ -0,0 +1,77 @@
...
+* Examine in tensorboard
+* Train with new data
+* Spawn inference cluster
+* Run tests
+* Show model provenance pipeline
+
+## Alternatives
+
+* Tensorboard - Wasn't meant for large number of models. It's better for very detailed examination of smaller number of models. Uses tf.Event files
+* MLFlow - One of big cons for this is using files as storage for models. That would require something like dask or spark to query them efficiently. Can store files in multiple backends (S3 and GCS among other things)
```
YujiOshima: Does it really need dask or spark? It has [REST API](https://mlflow.org/docs/latest/rest-api.html).

inc0: Well if you'll try to query 50000 records from one file (and by query I mean "highest value of X") it's going to require something more...

YujiOshima:
TensorBoard is so useful but it is not suitable for general Experimental Tracking.
It is difficult to manage a huge number of models and very specialized to TF ( or ONNX ). The Experimental Tracking is not only for DL models but all ML experiments.
Ideally, TensorBoard is linked from Experimental Tracking UI (MLFlow or StudioML).

proposals/experiment-tracking-proposal.md
inc0: Right, so it's no good for persistent experiment tracking, which we're after

inc0:
That's the idea @YujiOshima :) I was thinking of something like button "spawn Tensorboard from these 3 models"

inc0:
Also Tensorboard will have support for PyTorch, which is super cool. We would still need something for scikit-learn but it's getting better!

proposals/experiment-tracking-proposal.md
ddutta: How about first design a better modelDB equivalent and then use that for tracking experiments? I would recommend we keep each of these very independent for now. So that Kubeflow components/apps can integrate with a wide variety of tools e.g. TFX, katib, autoML.

inc0: Reason I think our current model is flawed is because we have 2 sources of truth. Katib uses sql, modeldb uses mongodb or sqlite. Every time you want modeldb will sync stuff from katibs db. That means if you do sync with tens of thousands of models, it's going to lock whole system. I think we should build single source of truth of where models are and how they performed and Katib should use it. This would negate need for Katibs database alltogether and, therefore, made it much easier to handle. In another issue we've discussed Katib as model management tool, but we've decided that Katibs scope is hyperparameter tuning, and model management is something different (however required).

proposals/experiment-tracking-proposal.md
zmhassan: Although MLFlow does store files on disk. It would save sometime if folks looked at forking it and then integrating a database to store the tracking information.

inc0:
@zmhassan with MLFlow, let's hypothetically assume we have 50 000 of models for detecting cats. Is there easy way to select model with highest accuracy and spawn seldon (or tf-serving) out of them? Quick look at API doesn't look like MLFlow has any form of querying. I also don't see whole lot of model provenance out there, but that probably could be implemented. Also, how easy would it be to integrate it to TFJob? As in start tfjob from this run, retrain model X etc. Another thing is integration with tensorboard. MLFlow seems to be alternative to tensorboard, and tensorboard, for what it is (UI for examining models performance) is excellent (imho). Any chances we could keep using it?

zmhassan:
@inc0 Without going into implementation details. MLFlow has 3 sub components. 1 for tracking, other for model serving, and lastly project dependency management. I was proposing that we take the tracking server and then just point a db to track the metadata for the experiment. The goal was to have a solution that works across different ML frameworks tensorflow, scikitlearn, mxnet, pytorch, or whatever new ML framework comes out next. imho, once we have the model training tracking setup proper then i think it would be good to have that dialog on model serving. I think the specification on experiment tracking can be expanded on to include more details. I'll send a pr to @inc0 branch with more details.

AdrianLsk:
How about to use [dvc](https://dvc.org/features), a git-like framework with cloud storage buckets for managing the versioning and storing of models? They provide a metric tracking as well.

jlewi:
The Katib page has a screencast illustrating ModelDB.
https://user-images.githubusercontent.com/10014831/38241910-64fb0646-376e-11e8-8b98-c26e577f3935.gif

As a strawman this looks like it has most of the features we want

- API to report metrics
- A UI to browse models
- Some built in visualizations
- Ability to launch TensorBoard

I realize there are concerns about MongoDB but for a strawman running it in a container with a PVC seems fine.

Have folks checked out the demo:
http://modeldb.csail.mit.edu:3000/projects/6/models

Its pretty slick and it looks like it provides most of what we want. 

I'd be thrilled if we managed to get that working and part of the 0.4 release. 

My conjecture is that if we get some more first hand experience with ModelDB we'll be in a better position to figure out where to go from here.

ddysher:
Are we going to track all experiments, last time I talked to data scientist, it's not very useful for them to track all experiments, much like software debugging, where we change code and experiment, without using git commit; we checkin code only when we feel comfortable about our change.

modeldb wraps existing libraries (sklearn, sparkml) to sync model data, e.g. users are required to use sync version instead of stock methods, which I think can be fragile to library change and require extra work to support more frameworks.  Also, IIRC, syncing model data takes considerable time.

proposals/experiment-tracking-proposal.md
```diff
@@ -0,0 +1,84 @@
...
+
+Another feature commonly asked for is model provenance. It's crucial to be able to reproduce results. For every model we need to record:
+
+* Inital state, whether it's random weights or based on preexisting models
+* Dataset used for training
+* Dataset used for testing
+* Feature engineering pipeline used
+* Katib study id
+* Model architecture (code used)
+* Hyperparameters
```
durandom: If the code to create the model is in a VCS, e.g. git, it should also track the version of the code used to create the model

proposals/experiment-tracking-proposal.md
```diff
@@ -0,0 +1,84 @@
...
+
+To be able to pick best model for problem, we need to record metrics. Metrics can differ problem to problem, but we can support single number as quality weight
+(user can define this number per experiment, whether it's accuracy, p1 score etc). We need to support very efficient queries using this metric.
+
+## Model introspection
+
+For selected models we should be able to setup model introspection tools, like Tensorboard.
+Tensorboard provides good utility, allows comparison of few models and recently it was announced that it will integrate with pytorch. I think it's reasonable to use Tensorboard
+for this problem and allow easy spawn of tensorboard instance for selected models. We might need to find alternative for scikit-learn. Perhaps we can try mlflow for scikit-learn.
+
```
durandom: There is also http://tensorboardx.readthedocs.io which can create a tensorboard from any python code.
We've started working with mlflow because it has a nice web ui and is easy to use with it's python framework.
I dont know if tensorboard with tensorboardx has some benefits though

proposals/experiment-tracking-proposal.md
```diff
@@ -0,0 +1,84 @@
...
+* Examine in tensorboard
+* Train with new data
+* Spawn inference cluster
+* Run tests
+* Show model provenance pipeline
+
+## Alternatives
+
+* Tensorboard - Wasn't meant for large number of models. It's better for very detailed examination of smaller number of models. Uses tf.Event files
+* MLFlow - One of the cons for this is that experiment metadata is stored on disk. In kubernetes may require persistent volumes. A better approach would be to store metadata in a database. Pros, can store models in multiple backends (S3, Azure Cloud Storage and GCS among other things)
```
durandom: I would not care if mlflow stores it in a db or in a file. If you use mlflow we should use their REST api as the interface and let them handle persistence. And for a DB you'd also need a PV, so 🤷‍♂️ 

We have started a repo to make kubeflow run on openshift: https://github.com/AICoE/experiment-tracking

proposals/experiment-tracking-proposal.md
inc0: agree, that's what I meant by "model architecture". But good idea would be to make it point to
* code (including commit id)
* docker image

proposals/experiment-tracking-proposal.md
```diff
@@ -0,0 +1,84 @@
...
+* I'm cloud operator in ML team. I would like to take current production model (architecture+hyperparams+training state) and retrain it with new data as it becomes available. I would want to run suite of tests and determine if new model performs better. If it does, I'd like to spawn tf-serving (or seldon) cluster and perform rolling upgrade to new model.
+* I'm part of highly sophisticated ML team. I'd like to automate retraining->testing->rollout for models so they can be upgraded nightly without supervision.
+* I’m part of the ML team in company. I would like to be able to track my parameters used to train an ML job and track the metrics produced. I would like to have isolation so I can find the experiments I worked on. I would like the ability to compare multiple training rules side by side and pick the best model. Once I select the best model I would like to deploy that model to production. I may also want to test my model with historical data to see how well it performs and maybe roll out my experiment to a subset of users before fully rolling out to all users.
+
+
+## Scale considerations
```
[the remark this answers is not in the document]
inc0: We need more than just number of replicas. It's important thing to consider when selecting underlying database

proposals/experiment-tracking-proposal.md
inc0: Managed database (like big table) doesn't (or at least you don't care about it). Biggest issue for MLFlow API (which is directly tied to file storage) is lack of querying. Currently (unless I'm mistaken) there is no good way in MLFlow to ask for model with highest accuracy. It could be implemented, but then comparisons would be done in python, so not super scalable.

proposals/experiment-tracking-proposal.md
inc0: I think it does! You could use it to add Tensorboard to scikit-learn (just log accuracy every batch of training)

proposals/experiment-tracking-proposal.md
```diff
@@ -0,0 +1,84 @@
...
+* I'm part of big ML team in company. Our whole team works on single problem (for example search) and every person builds their models. I'd like to be able to compare my models with others. I want to be safe that nobody will accidentally delete model I'm working on.
+* I'm cloud operator in ML team. I would like to take current production model (architecture+hyperparams+training state) and retrain it with new data as it becomes available. I would want to run suite of tests and determine if new model performs better. If it does, I'd like to spawn tf-serving (or seldon) cluster and perform rolling upgrade to new model.
+* I'm part of highly sophisticated ML team. I'd like to automate retraining->testing->rollout for models so they can be upgraded nightly without supervision.
```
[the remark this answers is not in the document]
inc0: Model management is alternative term to experiment tracking I think. At least I've understood it as such. As for functionality, because we'll make it k8s-native, cost of adding this feature will be so low that I think we should do it just for users benefit. Ongoing monitoring of models isn't something in scope, but as long as this monitoring agent saves observed metrics (say avg accuracy over last X days) back to this service, you still can benefit from this.

inc0:
```
Are we going to track all experiments, last time I talked to data scientist, it's not very useful for them to track all experiments, much like software debugging, where we change code and experiment, without using git commit; we checkin code only when we feel comfortable about our change.
```
I think we can track all experiments, that's why we've been putting emphasis on scale. I expect great majority of models to never see production. I'd say commit to branch->train->save to exp_track->look up tensorboard, compare with current best model etc->decide what to do, move to prod or ignore forever

proposals/experiment-tracking-proposal.md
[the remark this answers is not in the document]
durandom: I think the notion of "model management" and "experiment tracking" is slightly different. "management" has a production connotation and "experiment" has a devel connotation. Did @jlewi in [this](comment thread) get to a common definition? [This](mlflow issue) has also a discussion around this. And a google search for ["experiment tracking" ai ml](https://www.google.de/search?q=%22model+management%22+ai+ml&oq=%22model+management%22+ai+ml) vs ["model management" ai ml](https://www.google.de/search?q=%22model+management%22+ai+ml&oq=%22model+management%22+ai+ml) gives 500 vs 75k results. 
Please dont get me wrong. I'm all for having a solution for this, because I think too this is a missing component of kubeflow. 
I'd just limit the scope to the devel side of the house and let pachyderm and seldon focus on the production side.

proposals/experiment-tracking-proposal.md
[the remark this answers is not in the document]
durandom: Oh, I'm thinking an `experiment` would be a kubernetes primitive, like a `job` - no replicas involved. The `job` will be scheduled by kubernetes. So if you run 100 or 1000 experiments, you just create them and let kubernetes handle the scaling, i.e. the scheduling

proposals/experiment-tracking-proposal.md
durandom: > Managed database (like big table) 

wouldnt this introduce a dependency that kubeflow wants to avoid?

> Biggest issue for MLFlow API (which is directly tied to file storage) is lack of querying

Actually [there is a REST API for that](https://www.mlflow.org/docs/latest/rest-api.html#search-runs). But I havent used it and I'm not sure how good it scales

proposals/experiment-tracking-proposal.md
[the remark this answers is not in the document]
inc0: I'm not sure if that's what you mean, but we've discussed using CRDs as experiments but decided against it. Sheer number of experiments involved and lack of querying is a problem. We still need database somewhere. As for running actual experiments, then yeah, they will be tfjobs so regular pods.

proposals/experiment-tracking-proposal.md
inc0: https://github.com/kubeflow/community/issues/188 I've noted it here briefly with some consequences and how to make it manageable for operators (imho).

As for search string it's really not much. I still can't see option of "get me best model" without using spark/dask

proposals/experiment-tracking-proposal.md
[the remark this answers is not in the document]
inc0: So one clarification - when I'm saying, for example, model rollout, what I mean is single call to k8s to spawn seldon cluster. Actual serving, monitoring etc is beyond scope, I agree, but I think it'd be a nice touch to allow one-click mechanism. For Pachyderm integration look lower, I actually wanted to keep pipeline uuid in database. If someone will use pachyderm, we'll integrate with it and allow quick navigation. For example one-click link to relevant pachyderm ui

zmhassan:
### ML Hyperparameter Tuning:
I think if we make the proposal more focused on the use case here. The purpose of tracking experiments is that we want to perform hyperparameter tuning.  

### Unrelated items in proposal:
* imho, we don't need to include model provenance into this proposal as it is already being done by pachyderm. Correct me if I'm wrong??
* We should have a conversation about rolling this out but I don't think this proposal is the right place. Anything like ML DevOps qa-> staging -> prod roll out stuff is another conversation too.

### Conclusion
I think if we keep it simple we can get something working (mvp). The main motivation why I suggested MLFlow was because it supports multiple ML frameworks, not just tensorflow. To have wider adoption I suggest we cast our net wider and allow for more ML frameworks to be utilized.

karlmutch:
Some thoughts,

At Sentient, studioml, our focus has been on enabling our evolutionary NN experiments.  As a previous contributor has mentioned, workflows concentrating on single experiments (or individuals) have scale friction.  That however should not pose an issue if all meta data placed by experiments on traditional storage is machine readable.  In some instances we are specifically interested in looking at individual experiments, for example during development or when beginning work in a specific domain.  In other situations we can ingest experiment data into our own tooling and investigate at much higher levels of granularity.

Couple this with having idempotent data and a small set of rules about incremental results enables the storage to become the system of record, rather than a DB.  it also helps with the who owns what issue deferring that to the storage platform.

Defining an entity model, relationships and forcing an architecture might reduce deployment freedom.  One alternative, for example, we have chosen is to concentrate on a portable data format as the primary means of defining our interfaces.  Each technology we then use for UI, reporting, project authoring etc aligns around the formats.  For our own purposes we are using JSON and S3 artifacts to act as our system of record.  Queries on the S3 platform, for example, can be done using DSLs from cloud providers or tooling can be used to migrate meta-data to other DB's etc.  This defers technology compromises to the consumer.  For us that means we can avoid the mongo loose schema, and lost data problems, or going the spark big-data route.  This also avoids the issue of who is the responsible party for the cost of non-artifact storage and queries, the user pays rather than the experimenter.

Anyway thats my 2 cents.

jlewi:
Thanks @karlmutch 

@zmhassan 

> I think if we make the proposal more focused on the use case here. The purpose of tracking experiments is that we want to perform hyperparameter tuning.

I don't think we want to limit our selves to hyperparameter tuning. I think a very common use case is just training a model in notebook; saving that model somewhere and wanting a simple way to track/browse all such models.

zmhassan:
@jlewi Good point. Definitely good to capture all use cases. We experimented with mlflow you can import the python library into Jupyter notebook and be able to track parameters/metrics. 

@karlmutch Definitely interested in reading up on studioml.

inc0:
> For our own purposes we are using JSON and S3 artifacts to act as our system of record.

One issue with this approach (unless coupled with database, which I propose) is that it would be extremely hard to find model based on performance. I'm talking about query like "show me top 5 models with highest p1 score for experiment cat or dog". Can you deal with this kind of queries too?

karlmutch:
> > For our own purposes we are using JSON and S3 artifacts to act as our system of record.
> 
> One issue with this approach (unless coupled with database, which I propose) is that it would be extremely hard to find model based on performance. I'm talking about query like "show me top 5 models with highest p1 score for experiment cat or dog". Can you deal with this kind of queries too?

In the cases where experimenters are using cloud based storage the cloud vendors typically offer query engines that support this use case.  AWS Athena and from memory Google Cloud Datastore will both do this, not sure from memory about Azure as we use our own tech on their stack.

In the case where a tool like Minio is being used, or our customers/experimenters wish to make use of their own query engine then the json data is ingested into a DB.   Things remain coherent between the store and the DB because we follow idempotency and simple rules around experiments that are in progress when the ingest occurs.  From the Studio perspective however the choices in this area are not mandated by the experimentation framework.

proposals/experiment-tracking-proposal.md
andyk: Hi team, I'm the PM on the MLflow team at Databricks. Some of the engineers will chime in here too. Adding a database-backed tracking store to the tracking server is on our roadmap, and there is already a pluggable API!

andyk:
This is an awesome discussion. I'm a PM at Databricks, I work full time on MLflow, and I'm a huge fan of KubeFlow; it's been on our roadmap to engage about exactly this topic. In addition to what @zmhassan has been sharing, if there is any way we can be helpful, I know @mateiz @mparkhe @pogil @aarondav and others in the MLflow community would be excited to help answer any questions, discuss architecture, etc. In particular, I think @mparkhe -- an MLflow core dev -- has been thinking about this and has some more details.

mparkhe:
As @andyk  mentioned above, the MLflow team has been thinking about various engagement efforts and how we can support those within MLflow architecture. We'd be excited to have MLflow be a component of the Kubeflow ecosystem. To that effect, we would love to understand what would be required to make this happen.

Here are some details about specific projects on our roadmap, that would help answer questions about MLflow.

### Scalable backend store for Tracking Server data
The current storage layer for experiments, runs and other data is files. However, we are planning to add support for other scalable storage. Query layer will be built using a generic query layer like SQLAlchemy that would support most relational databases. There have been some queries for support of KV stores, however search query pattern supported by MLflow APIs, is most suited for relational databases out of the box, without the need for additional modeling of tables for each different types of key value stores.

As a side note, MLflow supports ability to store large file artifacts like models, training and test data, images, plots, … etc in several implementations of artifact stores: AWS S3, Azure Blob Store, GCP, … etc. The query pattern for these APIs is to access the actual artifact and purely dependent on these object store implementations.


### APIs and Interfaces
One of the design principles for all 3 MLflow components was API first. We designed these interfaces, first and then various different implementations to be able to plug into them.

Current open source has implementations for FileStore and RestStore, which implement these interfaces over storage layer. Even with FileStore implementations, most of the APIs are efficient, since they index into a specific experiment/run folder. With many experiments and runs, search API can get slow since FileStore will access underlying data and then the search functionality is implemented in python.

With the above mentioned change to have a SQLAlchemy layer, these queries would be re-written to be pushed down the the appropriate backend query engine. This would enable to plug in any database backend store that can be queried through SQLAlchemy and we expect this solution to scale. For production use cases, using relational DB like MySQL would work. We have stress tested with realistic production workloads (1000s of experiments, ~ 100K runs, millions of metrics, … etc) for MLflow API pointed to MySQL backend (local machine and RDS) and found it to be performant. For instance, an indexed MySQL table for metrics returned desired results in single digit milliseconds. We believe that such an implementation would be suitable for most production workloads.


### MLflow UI
The current UI implementation supports almost all APIs, including Search and viewing artifacts. We are working on releasing the next version to have more feature coverage like CRUD operations on MLflow entities (eg: deleting and renaming experiments), support easy visibility of multi-step workflows. Many UI components like graphical view for run metrics has been contributed by non-Databricks community members. The MLflow team at Databricks will continue to support and add more functionality here.


####   

We look forward to contributions to MLflow and also collaborating to Kubeflow ecosystem towards this common goal.

zmhassan:
+1 

I've been experimenting with porting over mlflow to run on kubernetes with crd's and operators. Once that is done and setup to work with ksonnet. It will be a simple ksonnet installation. 

imho in this space are a lot of technical choices, imho I think in this case we have two communities that could combine forces to create something great.

jlewi:
Thanks @andyk @mparkhe that sounds great would be great to explore possible collaborations.

Perhaps someone could give an MLFlow demo at one of our community meetings.

mparkhe:
> Thanks @andyk @mparkhe that sounds great would be great to explore possible collaborations.
> 
> Perhaps someone could give an MLFlow demo at one of our community meetings.

Hi @jlewi:  That sounds fantastic. We would love to get on the schedule.

cc: @mateiz, @pogil

mpvartak:
Hey all, this is Manasi from ModelDB. It's pretty clear that multiple groups are working towards the same goal of a general-purpose model management system. Can the KubeFlow community think about defining a generic interface for model management? (or requirements for a model management system to be compatible with KubeFlow?)

That way there can be multiple implementations of KubeFlow-compatible model management systems and users can pick the one that works best for them. I imagine this would be similar to having multiple model serving implementations for KubeFlow.

jlewi:
Close this because its stale.

BioGeek:
What is the current status of this proposal/of  experiment tracking in Kubeflow? Is there another proposal which supersedes this one or is progress tracked somewhere else?

PR #201, opened 2018-10-27
Adding apache beam operator proposal for streaming data pipelines

zak-hassan-hs:
# Introduction

Kubeflow currently doesn't have support for running streaming/batch workloads. There are efforts of bringing spark operator but I think it would be good to be able to experiment with other streaming backends like flink and others. In AI/ML systems there may be some data cleaning or data preparation involved before running through deep learning frameworks like tensorflow.

# What is Apache Beam?
Apache Beam is an opensource, multiple backend system that allows for batch and streaming workloads. It is parallel processing and portable so developers have the choice of using runners. The backends/runners supported are apache spark, apache flink, apache gearpump, apache apex, google cloud dataflow and apache samza. The beam api also allows you to write your pipelines in multiple programming languages such as java, golang, python, etc.

One of the missing features of kubeflow is handling streaming pipeline. I think apache beam might help.

# Related Issues
 Judging from these issues these issues this may be something the community wants. 

- [Use Flink to provide an alternative runner for TFX](https://github.com/kubeflow/kubeflow/issues/1583)
- [Explore usage of the spark-operator within kubeflow](https://github.com/kubeflow/kubeflow/issues/155)
- [Pachyderm and Kubeflow integration](https://github.com/kubeflow/kubeflow/issues/151)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/201)


zmhassan:
@texasmichelle  @gaocegege What do you think about this proposal?

Jeffwan:
This would be a generic way to support BEAM jobs? In my case, I want to start from supporting more TFX backed runners. For cloud without cloud runners like DataFlow or on premise users, we need to leverage Flink or Spark to achieve distributed computing other than local runner. This is very important for enterprise users to adopt Kubeflow Pipeline in production environment. 

The next step might be beam operator. In my case, we can start from KF pipeline use case.

zmhassan:
All good points we can work together on building the first prototype for
the beam operator and then present it at one of their community meetings. A
lot of people start discussions but nothing is more real then a live demo.
On to implementation. What language do you want to write the operator ?
Golang? By the way I’ve written operators before in golang. I can help
quickly build up our first prototype.

proposals/apache-beam-proposal.md
```diff
@@ -0,0 +1,55 @@
...
+# UI or API
+
+WIP
+
+
+# Design
+
+THe beam operator will run the pipeline for preprocessing and will connect to spark  cluster to run on the backend. 
+
+![MPI Operator](diagrams/beam-operator.png)
```
gaocegege:
```suggestion
![Beam Operator](diagrams/beam-operator.png)
```

gaocegege:
Personally, I have no objection about it. I am not working on pipelines, thus cc @jlewi @IronPan 

/assign @IronPan

proposals/apache-beam-proposal.md
zmhassan: Fixed

jlewi:
There is now a flink operator
https://github.com/GoogleCloudPlatform/flink-on-k8s-operator

jlewi:
Looks like this issue got lost.

@zak-hassan-hs apologies it has taken so long to resolve this.

At this point I think we no longer need to build our own operators as Flink and Beam have built out appropriate K8s operators.

We have open issues like kubeflow/kubeflow#1583 to incorporate these operators where it makes sense.

PR #202, opened 2018-10-30
Add jwwandy to member.yaml

jwwandy:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/202)

### DIFF
--- a/members.yaml
+++ b/members.yaml
@@ -171,6 +171,13 @@
   affiliation: Google
   email: jlewi@google.com
 
+- github: jwwandy
+  slack: ajwwei
+  firstName: Andy
+  lastName: Wei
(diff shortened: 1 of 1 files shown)


members.yaml
```diff
@@ -296,3 +296,10 @@
   lastName: Yan
   affiliation: Caicloud
   email: yanpenghui@caicloud.io
+
+- github: jwwandy
```
abhi-g: Hi Andy, We organize this file in alphabetical order of your github id. Can you please move this section up accordingly. thanks

jwwandy: Thanks for reminding me. Didn't notice that. Just move it to the correct position

Issue #125, opened 2018-05-03
[devstats] Set companies


gaocegege:
http://devstats.kubeflow.dev/d/9/companies-table?orgId=1

Now we do not support the company related statistics. Maybe it requires some additional configuration.

gaocegege:
- [import_affs](https://github.com/cncf/devstats/blob/master/cmd/import_affs/import_affs.go)
- `import_affs` takes one parameter - JSON file name (this is a file from [cncf/gitdm](https://github.com/cncf/gitdm): [github_users.json](https://raw.githubusercontent.com/cncf/gitdm/master/github_users.json)
- This tools imports GitHub usernames (in addition to logins from GHA) and creates developers - companies affiliations (that can be used by [Companies stats](https://k8s.devstats.cncf.io/dashboard/db/companies-stats?orgId=1) metric)

gaocegege:
I followed the instructions to generate a github_users.json: https://raw.githubusercontent.com/gaocegege/gitdm/master/github_users.json

jlewi:
Couldn't we use members.yaml to determine which company each user belongs to?

gaocegege:
We can but devstats already has a mature tool to do it. I think we could reuse it.

jlewi:
#208 added a simple script that can use the github_users.json file and the data exported from devstats to compute # of PRs from different companies.

Here's my fork of https://github.com/jlewi/gitdm
I made some changes to compute github_users.json for Kubeflow.

amsaha:
Following up on https://kubeflow.slack.com/archives/C8Q0QJYNB/p1542385363024300, what is pending? It seems that #208 is already merged.

jlewi:
@amsaha I think we would like to get devstats updated so that it automatically and continuously computes PRs by companies.

I suspect there's two things we need to do

1. Figure out how to make devstats use github_users.json to compute PRs per company
1. Figure out how to regularly update github_users.json

amsaha:
Thanks @jlewi. Will take a look soon.

gaocegege:
@gyliu513 Awesome work!

BTW, I am not sure how to change the affiliation, in the dashboard, I am independent while I am not.

gyliu513:
@gaocegege just create a review request to openstack like this https://review.openstack.org/#/c/625774/ , some detailed steps for how to create review request in openstack https://docs.openstack.org/infra/manual/developers.html 

[...]

PR #208, opened 2018-11-13
A script to compute PRs by company.

jlewi:
* Script is based on gitdm (https://github.com/jlewi/gitdm) and exporting
 number  of PRs from devstats.kubeflow.org

* Script falls back to trying to use the user's email if no company is set.

Related to #125

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/208)

### DIFF
--- a/scripts/prs_by_company.py
+++ b/scripts/prs_by_company.py
@@ -1,0 +1,84 @@
+#!/usr/bin/python
+#
+# Product a CSV file containing number of PRs produced by each company.
+# The input to the script is:
+# 1) csv file containing number of PRs for different GitHub users. this_file
+#    file is produced by exporting the data from devstats table
+#    https://devstats.kubeflow.org/d/13/developers-table?orgId=1&var-period_name=v0.2.0%20-%20now&var-metric=prs&var-period=anno_1_now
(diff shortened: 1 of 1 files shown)


ashahba:
Hi @jlewi would you mind providing the link for the current spreadsheet here?

abhi-g:
/lgtm
Thanks @jlewi

abhi-g:
I noticed that in community/members.yaml we have a better representation of the "affiliation". that might be one of the fields we could use to improve stats further? Probably as the next iteration..

johnugeorge:
@jlewi  How often is the spreadsheet updated?

jlewi:
Here's the current spreadsheet.
https://docs.google.com/spreadsheets/d/1OjGt3Cjmbbmk4fRFzCPaojRO3d64S06TqIoTwBXOffA/edit#gid=726306870

The spreadsheet is updated manually; so infrequently. Realistically it will only be recomputed when we want to present stats about the 

The goal of #125 is to get these stats to be computed by devstats.kubeflow.org so they are updated regularly.

This script was a hack just to get stats for @aronchick for his presentation at Kubecon yesterday.

PR #224, opened 2019-01-29
Create a script to plot total and open issues for current release.

jlewi:
* The new script replaces the existing get_issue_counts.py.
* The corresponding notebooks uses plotly to create an interactive plot.
  The interactive plot is published on plotly and is publicly available.

* Use GitHub's GraphQL API rather than the older REST API

Create a script to identify issues that need attention.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/224)


scripts/project_stats.ipynb
```diff
@@ -0,0 +1,174 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 194,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "<module 'project_stats' from '/home/jlewi/git_kubeflow-community/scripts/project_stats.py'>"
```
abhi-g: is this intended to run for everyone? If so, might need to replace /home/jlewi/

jlewi: This is just the output value. So when run on a different machine it would output a different value.

PR #232, opened 2019-02-17
Upgrade devstats and move it to its own project.

### LINKED ISSUE #229: [devstats] exclude homebrew repos
We forked homebrew repos to Kubeflow.

We should exclude those rep from stats.

jlewi:
* Create a new instance of devstats running on its own devstats cluster
  as opposed to the ci cluster.

* Upgrade devstats to the latest version
  influxdb is no longer used.

* Update the docker image.

Fix #229  - developers table not showing any data

* Use Cloud NFS to store all the data and mount it on all volumes using PVC.
* Update the instructions
* Update the Grafana dashboard configs.

* Configure grafana datasource using a YAML file

* Add a NFS fileshare to use for grafana and postgres rather than PD.

* Remove dashboards configmap because we will use NFS
* Remove projects configmap because we will use NFS.
* Create PV/PVC for NFS.

* Dockerfile should create symbolic links to allow loading the projects.YAML

Fix #229 - Exclude homebrew repos

* Enable the ingress
* Add back the certificate.

Add comment about tcountries.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/232)


abhi-g:
Thanks Jeremy.

/lgtm
/approve

PR #243, opened 2019-03-29
Add Clive from Seldon to members.yaml

cliveseldon:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/243)


jlewi:
members are now maintained in github.org/kubeflow/internal-acls

PR #245, opened 2019-04-01
Add bloomberg to member org and dsun to members

yuzisun:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/245)

### DIFF
--- a/member_organizations.yaml
+++ b/member_organizations.yaml
@@ -46,6 +46,13 @@
     - kimwnasptd@arrikto.com
   url: https://www.arrikto.com
 
+- name: Bloomberg
+  contacts:
+    - dsun20@bloomberg.net
+    - klaban1@bloomeberg.net
+    - ifilonenko@bloomberg.net
+  url: http://www.bloomberg.com
+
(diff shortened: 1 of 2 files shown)


member_organizations.yaml
```diff
@@ -46,6 +46,14 @@
     - kimwnasptd@arrikto.com
   url: https://www.arrikto.com
 
+- name: Bloomberg
+  contacts:
+    - dsun20@bloomberg.net
+    - klaban1@bloomeberg.net
+    - ifilonenko@bloomberg.net
+  url: http://www.bloomberg.com
+  hiringUrl: https://www.bloomberg.com/careers
```
abhi-g: Do we really need the hiringUrl in here?

abhi-g:
@yuzisun I'd prefer if we not pollute that file with hiring solicitation :)

member_organizations.yaml
yuzisun: no, let me take it off.

abhi-g: thanks

PR #249, opened 2019-04-08
Fix project_stats; properly compute closed time.

### LINKED ISSUE #244: Script to plot number of open issues has a bug; number of open issues shouldn't be monotonic
https://github.com/kubeflow/community/blob/master/scripts/project_stats.ipynb

See plot below the number of open issues shouldn't be increasing monotonically.
It should go up and down. 
at time t the number of open issues should be the number of open issues at that time. However, I suspect right now we are only counting issues if they are still open and were opened prior to time t.

newplot (2)

jlewi:
* When an issue is closed we need to set the time of the event to the closed
  time for the issue not the created time.

* This explains why the number of open issues was always monotonically
  decreasing.

Fix #244

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/249)


jlewi:
@zabbasi ping

Issue #218, opened 2018-12-04
Can we delete https://github.com/kubeflow/community/blob/master/members.yaml


jlewi:
Can we delete 
 https://github.com/kubeflow/community/blob/master/members.yaml

Is all the information stored in other places?

We now use
https://github.com/kubeflow/internal-acls/blob/master/github-orgs/kubeflow/org.yaml

To maintain a list of members in the GitHub org.

* Company info is obtained from the users GitHub profile
  * We have dashboards and stats to get that information.
* What about slack id and email?

It doesn't look like GitHub profiles have a field for slack.
Slack doesn't have a profile field for GitHub.

In slack you can customize the profile but that requires a paid plan
https://get.slack.help/hc/en-us/articles/212281478-Customize-member-profiles

In slack its possible to create a link to a particular user
https://get.slack.help/hc/en-us/articles/360003827751-Create-a-link-to-a-member-s-profile-#free,-standard,-and-plus-plans-1

But its not a pretty link. 

So maybe the best option is to keep members.yaml but move to internal-acls so that at least people can submit a single PR modifying both.

We could also make it optional.

jlewi:
I'm leaning towards delete. I don't think members.yaml will be used very often if its just about looking up slack ids.

texasmichelle:
ah, I was wondering how we maintained org membership. 👍 to deleting `members.yaml`.

PR #254, opened 2019-04-12
added self to members.yaml

kevinbache:


jlewi:
This file isn't used for anything and we'd like to delete it: #218

If you want to join the github org join this file.
https://github.com/kubeflow/internal-acls/blob/master/github-orgs/kubeflow/org.yaml

Would you mind updating this PR to instead delete this file?

jlewi:
This file is being deleted.see #258

PR #268, opened 2019-05-15
Adding instructions for reinit script

johnugeorge:
`/devel/reinit.sh` needs to be run to regenerate all TS data once affiliations are changed

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/268)

@assign @jlewi


devstats/config/devel/restore_db.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
jlewi: What's this for?

johnugeorge: This is part of reinit script which is not in our fork

Issue #274, opened 2019-06-29
Automate Updating Kubeflow Calendar


scottilee:
Please add more weekly meetings to the Kubeflow calendar (https://calendar.google.com/calendar/embed?src=kubeflow.org_7l5vnbn8suj2se10sen81d9428@group.calendar.google.com) such as "KF Serving Working Group Sync @ Weekly from 9am to 10am on Wednesday (PST)".

Also, are the associated Hangouts for each meeting recorded? If so, are recordings posted to a central place or retroactively added as an attachment to the meeting invite?

scottilee:
@theadactyl I'd appreciate your help on this! The tasks would be:

1) Add all weekly meetings to the Kubeflow calendar such as "KF Serving Working Group Sync"
2) I've never seen meeting recording links anywhere. I think two good options would be to start adding them on the meeting notes doc or add them to the meeting invite afterwards.

By the way, is there a bit.ly or other url shorterned link to the Kubeflow calendar?

jlewi:
@scottilee and others a great way to help make this happen would be to automate it with the source of truth being GitHub.

Here's how this might work

1. Define a config file (e.g. YAML) that we can check into Git
1. Create a script to populate Google calendar based on the YAML config 

This way we can use GitHub to source community contributions for keeping the calendar up to date.

scottilee:
Good idea, I'll see if I can come up with something using https://developers.google.com/calendar/v3/reference/events/insert.

jlewi:
@scottilee as discussed in the bug; I think we will want to use a service account to run this program.

Do you know how we authorize a service account to have access to a Google calendar? Do we just make the service account an admin of the Google calendar?

theadactyl:
A bit unrelated, but possibly? is the fact that even trying to manually add
things to the calendar has been not working consistently. I get events
rejected after being accepted every so often.

scottilee:
@jlewi I updated the script to use a service account. If the Google account hosting the Kubeflow calendar has the Google Calendar API enabled, create the service account and share the Kubeflow calendar with the service account email and the script should be able to modify and create events. Let me know if I can clarify anything.

https://github.com/kubeflow/community/pull/288

jlewi:
@scottilee reopening this to continue to track automating sync'ing of the KF calendar.

scottilee:
[...]

PR #288, opened 2019-09-01
Cal Update Service Account

scottilee:
Updates the code to use a service account.

@jlewi This script would need to be passed a service account file created on the Google account that hosts the Kubeflow calendar and the calendar would need to be shared with the service account email address. Is this something you could help me with?

/assign @jlewi

Related to #274.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/288)


jlewi:
@scottilee @theadactyl owns the calendar. Can we use  a GitHub issue to track what actions are needed? 

It looks like we need to figure out. We might want to setup a new Kubeflow community project to host this infra. Lets use the issue to track that and figure it out.

scripts/calendar_import.py
```diff
@@ -4,20 +4,26 @@
...
 from datetime import datetime
 import logging
 import os.path
 import yaml
 import googleapiclient.errors
 from googleapiclient.discovery import build
-from google_auth_oauthlib.flow import InstalledAppFlow
+from oauth2client.service_account import ServiceAccountCredentials
 
+SERVICE_ACCOUNT_FILE = 'kubeflow_calendar_sa.json'
```
jlewi: Why are you hard coding the service_account file.

The convention is to look for a GOOGLE_APPLICATION_CREDENTIALS environment variable which is pointing at the key.

Moving forward though we'd actually like to use GKE Workload identity;
https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

Do you know if this code will work with Workload Identity?

scottilee: I read over that page and how I understand it is that a service account is available to a pod... so I'm assuming the service account could still be referred to as `GOOGLE_APPLICATION_CREDENTIALS`? If so, that should work.

jlewi: When workload identity is used the identity will not be determined by GOOGLE_APPLICATION_CREDENTIALS

Instead the pod should have a KSA (Kubernetes Service Account) that is mapped to a GSA(Google Service Account).  Google client libraries should be initiated to use the application default credentials in which case the credential should map to whatever GSA is mapped to the KSA.

At any rate I think we can figure it out when we get there.

jlewi:
@scottilee What is the next step with this PR?

jlewi:
I pulled these changes into kubeflow/community#298 which includes additional changes so I think we can go ahead and close this PR.

PR #292, opened 2019-09-24
Kf1 stats

abhi-g:
Compute effort for KF 1.0 and progress tracking stats for various components.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/292)


jlewi:
@abhi-g Should we count how many issues lack effort labels? This would give us some measure of confidence in our effort calculation.

We could also potentially add some default effort for those labels.

abhi-g:
<p>Added a count for issues missing labels. I haven't yet added a default effort label on those because of two reasons:</p><ul><li>It will probably be non-trivial to come up with a default effort estimate based on the variety of issues on the project.</li><li>Not all the issues will have effort labels, for eg. some issues such as tracking issues should probably not be labeled with effort otherwise we'll end up double counting.</li></ul>

PR #333, opened 2020-05-11
Fix YAML syntax areas in labels-owners file.

jlewi:
* The list of owners should be in field OWNERs in the area label.


yanniszark:
Thanks @jlewi!
/lgtm

Issue #327, opened 2020-04-15
[Guidelines] Add guideline on Authentication and Authorization


yanniszark:
Add a Guideline on how to perform Authorization for Kubeflow's Web Applications.
Authorization should rely on K8s RBAC as the source of truth.
For that reason, focus on SubjectAccessReview.

jlewi:
@yanniszark Are you planning on writing up a doc covering the points discussed in the thread?
https://groups.google.com/forum/#!searchin/kubeflow-discuss/subject$20access$20review%7Csort:date/kubeflow-discuss/whS_j4RlEJo/98xnBHH4BgAJ

In particular

* Getting rid of asymmetric permission model
* Making profiles a namespace scoped resource
* Possibly in scope - how we make namespace setup more extensible and not make profile controller this monolithic thing

yanniszark:
@jlewi yes, this is a doc I want to write, but it's so general I find it a bit hard to think of a proper issue title.
I think this issue should remain seperate and we should open a new one for the doc you mention.
Tbh, I haven't thought about making Profiles namespaced (was this mentioned somewhere?), but the discussion in the aforementioned thread is what I want to capture.
Something like `Evaluate and Improve our current Multi-User story`.
I will open a new issue for that, let's leave this one for the guideline on authorization (use SubjectAccessReview in order to depend on K8s RBAC).

jlewi:
@yanniszark We can use this issue to track the doc. I just wanted to clarify what the deliverable is.

PR #334, opened 2020-05-14
Add guideline on authentication and authorization

yanniszark:
Closes: https://github.com/kubeflow/community/issues/327

Adapted from: https://docs.google.com/document/d/11Xi-I2OqJvUuy_Zg0NskMF9GmworRDJhlu68poDgK5c

Add a guideline on authentication and authorization for Kubeflow
Applications. The guideline documents current best practices for
applications to follow in order to effectively integrate with Kubeflow.

/cc @jlewi @krishnadurai @vpavlin @pdmack @kunmingg


guidelines/auth.md
```diff
@@ -0,0 +1,126 @@
...
+
+However, we still want to make sure that the user has the same identity and the
+same RBAC permissions are consulted.
+
+![kubeflow request flow](images/kubeflow_request_flow.svg)
+
+
+### Authentication
+
+Authentication is done at the Istio Gateway and the result is propagated to web
```
jlewi, line 62: Are we actually using the ISTIO gateway for authentication? At least in some cases I think Authentication (e.g. login) happens external to the gateway (e.g. IAP). In this case the gateway is just doing authorization; verifying user/system is authorized to access the service mesh.

guidelines/auth.md
```diff
@@ -0,0 +1,126 @@
...
+      escalation checks, PSPs, Admission Controllers that rely on the user’s
+      identity).
+    * The issue with impersonation is that it is too powerful and there is no
+      way to scope it down. For example, if the Jupyter Notebooks Web App used
+      impersonation, it would have the full permissions of every user it could
+      impersonate. This presents too much of a surface for Web Apps that would
+      normally need little permissions, like the Jupyter Notebook Web App. For
+      Web Apps that would require elevated permissions anyway (e.g. an admin
+      panel web app), impersonation could be a viable option.
+
```
jlewi, line 118: How about adding a reference section with reference to past docs, discussions, etc...?

guidelines/auth.md
```diff
@@ -0,0 +1,126 @@
+# Authentication and Authorization Guideline
+
+This guideline documents the current best practices for authentication and
+authorization in Kubeflow.
+* Kubeflow Application Developers should use this guide to effectively integrate
+  their applications with Kubeflow.
```
jlewi, line 6: Should we split this into two docs? These seem like very different audiences and topics. My suggestion would be to focus first on application developers as I believe one of the motivations for this doc was to ensure future app developers to comply.

It looks like there are two core questions this doc is trying to answer for developers.

1. How do application know the identity of the user making a request?
1. How do applications act on behalf of the user?

I would suggest focusing on those questions and trimming out some of the other stuff.

For example, as an application developer is how authentication performed really matter? Really all application developers need to know is that there will be a JWT and what fields in the JWT to look for.

jlewi:
Thanks for writing this up!

guidelines/auth.md
yanniszark: I see what you mean.
In both cases, you have a proxy of some kind that authenticates the user.
IAP is a Google IAM proxy that authenticates all requests with Google.
The AuthService turns the Gateway into an OIDC proxy/client, authenticating every request with an OIDC Provider.
The result is a JWT that the Istio Authentication Policy checks + some headers with certain values of the JWT.

Since, as you pointed out, this document is for application developers, I will trim the details.

```
How about, authentication (JWT vertification) is performed by Istio. The result is propagated to  web apps...
```

> In this case the gateway is just doing authorization

A small correction here, as authorization is always done server-side (at the workload, e.g., the Istio Sidecar of a Jupyter Notebook) and not client-side (at the gateway).

guidelines/auth.md
yanniszark: Sounds good, I'll work on such a section.

guidelines/auth.md
yanniszark: This doc is for app developers first and foremost, agreed.
I mentioned cluster admins as they may be able to get some useful info out of it.
The questions you enumerate are very accurate and I will include them explicitly in the doc.
I will try to trim based on that, however I think it's good to see a bit of the bigger picture before diving into the details.

> Really all application developers need to know is that there will be a JWT and what fields in the JWT to look for.

I'm a bit confused here. Currently all of our apps are looking for certain headers.
Is that what you mean?

guidelines/auth.md
```diff
@@ -0,0 +1,126 @@
...
+* A username header, configurable via the `KUBEFLOW_USERID_HEADER` setting 
+    * Some IdPs (like Google) add a prefix to this header (e.g.,
+      `accounts.google.com:`). If the Web App wants to support stripping that
+      prefix, it should expose a `KUBEFLOW_USERID_PREFIX` setting).
+* A group header, configurable via the `KUBEFLOW_GROUPS_HEADER` setting.
+* A prefix for headers containing extra identity information (e.g., scopes),
+  configurable via the `KUBEFLOW_EXTRAS_HEADER_PREFIX` setting. The prefix is
+  removed, the remainder header name is percent decoded and becomes the key, the
+  header value becomes the value. ([see also](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy)).
+
```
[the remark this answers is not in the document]
yanniszark: Hmm, I'm not sure what the situation on no-auth setups is.
Do application developers need to do something extra to what is described in this doc in order to work?

yanniszark:
@krishnadurai @jlewi thanks a lot for taking the time to review this guideline! :smile: 
I will go over your points and provide a second iteration.
My feeling is that there is general agreement on the content and we are now polishing the guideline to make it extra helpful for the target audience.

guidelines/auth.md
jlewi: I think we are doing Authorization in both places. At the gateway we provide a coarse grained policy (at least we do with GCP). So that policy basically says if the user isn't allowed to access any KF service then reject the request.  We then apply a finer grained policy in the istio side car.

guidelines/auth.md
jlewi: So if I recall correctly there are two bits of information that the app needs to know in order to be able to identify the user.

1. Which header field stores the JWT
1. How to parse the actual email from the email field
     * Because at least with IAP I think email ends up including a prefix like accounts.google.com

ref: https://cloud.google.com/iap/docs/signed-headers-howto

I may have gotten the details wrong. Goal is just to tell app developers how to parse the JWT

guidelines/auth.md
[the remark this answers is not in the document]
jlewi: Should that go into the doc for cluster admins? 
For app developers we should aim to make anonymous identical to non-anonymous use case. i.e. whether its jeremy@lewi.us or anonymous@nodomain.org shouldn't really matter to app developers.

guidelines/auth.md
yanniszark: Got it! Addressing the original comment, I will amend the doc to say authentication is done at a proxy in front of the web-app. This is factually correct and generic enough.

guidelines/auth.md
yanniszark: Currently, our web applications don't parse any JWT.
They expect to find the user's identity in http headers.
This is what I describe in the doc.

guidelines/auth.md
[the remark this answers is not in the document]
yanniszark: > For app developers we should aim to make anonymous identical to non-anonymous use case.
Agreed!

jlewi:
@yanniszark SGTM; let me know when this is ready for another look.

jlewi:
@yanniszark ping?

yanniszark:
@jlewi I was preparing some bigger additions but we can merge the doc in a good shape and iterate later. I have added changes in a fixup commit to make it easier to review. Please take a look at your convenience!

guidelines/auth.md
```diff
@@ -0,0 +1,142 @@
...
+### Authentication
+
+Authentication is done at a proxy (e.g., the Istio Gateway, Google IAP) in front
+of the web app and the result is propagated to the web app in a header, in a way
+similar to the [authenticating proxy](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy)
+method. The authenticated entity can be a user (OIDC, Google IAP) or a
+ServiceAccount (TokenReview).
+
+At the moment only userid is added as a header. However, there are other
```
jlewi, line 69: Should we mention JWT's and cite https://jwt.io/ here? There's a couple points that might be worth calling out.
JWT's are bearer tokens so they need to be protected and they should be signed. In practice, this means we should be using ISTIO side cars to verify signatures and creating a zero trust network (using mTLS and encrypting traffic between all proxies). We don't do that everywhere.

jlewi:
Thanks @yanniszark 
/lgtm

Any idea on what we need to do to achieve SSO?

guidelines/auth.md
yanniszark: I would go with no mention of JWTs here, because I view it as an implementation detail of how authentication is done. Since this doc is about defining the contract to the application (find the user's identity in headers), how authentication is implemented (JWT, TLS certificate, opaque access token) is out-of-scope for this doc, because no matter what the implementation is, the application will only see the result in headers.
This would be an excellent discussion for a doc reviewing the infrastructure configuration (Istio, Gateways, External Proxies) of Kubeflow (and we should definitely start one of these).

yanniszark:
@jlewi thanks for your comments! If you are ok with the content of the current guideline, could you approve?

> Any idea on what we need to do to achieve SSO?

The SSO implementations I have seen need support from the identity provider.
Usually the provider sets an SSO cookie on their page, so when users are redirected to the Provider for login, the Providers authenticates them with that cookie and immediately redirects them back, giving the SSO experience.
Dex doesn't currently support this, I am pretty sure Google supports it as well as various enterprise offerings.

PR #339, opened 2020-05-27
Document the different assets that power kubeflow.org and their owners.

jlewi:


jlewi:
/assign @animeshsingh 

@animeshsingh Could you take a look at this?

jlewi:
ping @animeshsingh ?

how-to/kubeflow_assets.md
```diff
@@ -0,0 +1,164 @@
...
+
+* Password is stored in GCP Secret Manager project `kubeflow-admin`
+* You will also need a recovery token which are also stored in secret manager
+
+  * These are onetime use so remove it when you use it
+  * When you login you need to select use recovery token.
+
+## kubeflow.org GSuite Domain
+
+Admins
```
animeshsingh, line 29: Believe this is immutable? Is anyone collecting overall github statistics for the org?

how-to/kubeflow_assets.md
```diff
@@ -0,0 +1,164 @@
...
+
+https://medium.com/kubeflow
+
+Editors
+
+* @chrisheecho_78982
+* @ankitbahuguna
+* @jlewi
+* @edd
+* @thealamkin
```
animeshsingh, line 75:
```suggestion
* @thealamkin
* @animation2007
```

how-to/kubeflow_assets.md
```diff
@@ -0,0 +1,164 @@
...
+Root account:
+
+username: kubeflow
+password: 
+
+ * This is stored in Google Secret Manager inside project kubeflow-admins
+
+The following twitter accounts are linked as teammebers
+
+* @aronchick
```
animeshsingh, line 111:
```suggestion
* @animeshsingh 
* @aronchick
```

how-to/kubeflow_assets.md
jlewi: What do you mean by GitHub statistics? Policy doesn't allow us for granting non-googlers admin privileges. Also following the principle of least privilege we don't want to grow the number of root users/admins.
In accordance with that, we are trying to great policy groups (https://github.com/kubeflow/internal-acls) that have privileges needed to fulfill different roles (e.g. test/release admin, drive manager etc...) and then add folks as necessary to those groups.

Ideally everything is automated through GitOps so there is very little need to have root privileges.

how-to/kubeflow_assets.md
jlewi: I assume animation2007 is you? I don't have permission to add you as an editor; I think maybe Thea does. This doc is only intended to capture the current set of permissions. So I suggest we leave it as is and then you can follow up with Thea if you want to be added as an editor.

jlewi:
@animeshsingh I added you to the twitter team. For the others lets follow up. This PR was mainly intended to capture the state of the world and let people know who to contact about various resources.

how-to/kubeflow_assets.md
animeshsingh: sure

Issue #342, opened 2020-06-11
Additional calendar admins needed


jlewi:
We need additional admins for the [Kubeflow Calendar](https://calendar.google.com/calendar/embed?src=kubeflow.org_7l5vnbn8suj2se10sen81d9428%40group.calendar.google.com&ctz=America%2FLos_Angeles)

We currently have some partial tooling for managing the calendar. The [script](https://github.com/kubeflow/community/blob/master/scripts/calendar_import.py) configures the calendar based on [calendar.yaml](https://github.com/kubeflow/community/blob/master/calendar.yaml)

The script currently needs to be run manually by someone with admin privileges to the calendar.

Expectations for would be volunteers

* Figure out what needs to happen to gain access to the calendar
* Commit to helping SIGs/WGs/ etc... get meetings added to the calendar
* Document the process
* Work on automating the process to make it more scalable
* Move calendar.yaml into its OWN folder and create an OWNERs file
* Review and approve updates to the calendar

8bitmp3:
@jlewi Maybe we should also update the format/expand the community page https://www.kubeflow.org/docs/about/community/ and include a list of all SIGs in addition to the existing Slack channels with all the details about how people can participate. 

For instance, we could use something similar to https://www.tensorflow.org/community/ that leads to > https://www.tensorflow.org/community/forums > https://www.tensorflow.org/community/forums#special_interest_groups (there's a sense of flow).

Things are little fragmented right now but the users are growing, so it's all good, just an admin thing. Just got an email from Model Management SIG from @karlschriek - there's an issue with accessing the calendar. (I can help with all this later - right now there're things on https://github.com/kubeflow/website that require attention that I signed up for)

jlewi:
@8bitmp3 maybe open a separate issue about that?

@karlschriek Are you willing to be a calendar admin?

jlewi:
@karlschriek regarding your question in https://github.com/kubeflow/community/issues/342 

I believe the only ones with calendar access are myself and @theadactyl; @theadactyl and I can't afford to keep being the only ones administering the calendar. So before anything else gets added to the calendar we need to find 1-2 people willing to take on being a calendar admin.

jtfogarty:
I'm happy to help out and be a calendar admin.

jlewi:
Thanks @jtfogarty What's your Google account?

@scottilee also mentioned helping out with automating the calendar sync.

jlewi:
[...]

PR #343, opened 2020-06-11
Added recurring Model Management SIG Call

karlschriek:


calendar.yaml
```diff
@@ -259,3 +259,18 @@
...
+  name: Model Management SIG Call
+  date: 26/5/2020
+  time: 8:00am-9:00am
+  frequency: bi-weekly
+  video: https://meet.google.com/dxz-zthy-xjn
+  description:
+    - |
+      Notes & agenda: https://docs.google.com/document/d/1R2r1z5O4USpn3BW29-4blCoPueB3p2LAjvdKa97fT7U/edit?usp=sharing
+
+      Join at: https://zoom.us/j/799749911
```
jlewi, line 273: I don't think you will have access to this zoom call. Do you have a different way to host the meeting e.g. Google meets where you can host the meeting?

jlewi:
@karlschriek Is there an issue tracking setting up the call; in addition to figuring out how you will host the call you will need to get

kubeflow/community#342 resolved in order to update the calendar

calendar.yaml
karlschriek: Sorry that was a copy paste error. The correct link is under "video" above

karlschriek:
> ... you will need to get #342 resolved in order to update the calendar

I understood #342 as a request to remove the bottleneck of just a few people currently being able to do this. Or is it impossible to update the calendar at the moment?

PR #346, opened 2020-06-21
Add D2iQ to member orgs

Gilbert88:
Add D2iQ to member orgs


Gilbert88:
@googlebot I signed it!

PR #347, opened 2020-06-22
fix(fate-operator): Fix the egg replicas

gaocegege:
I think the replicas should be the sub-field of eggroll if I understand FATE design right.

/assign @LaynePeng  WDYT


LaynePeng:
> I think the replicas should be the sub-field of eggroll if I understand FATE design right.
> 
> /assign @LaynePeng WDYT

Yes, it is a typo in the proposal. Thanks for correcting.
/lgtm

PR #353, opened 2020-07-03
Added Slack Guidelines

mameshini:
Added Slack Guidelines
Created OWNERS file

### DIFF
--- a/slack/OWNERS
+++ b/slack/OWNERS
@@ -1,4 +1,2 @@
 approvers:
-  - abhi-g
-  - jlewi
   - mameshini
--- a/slack/OWNERS
+++ b/slack/OWNERS
@@ -1,0 +1,4 @@
+approvers:
+  - abhi-g
+  - jlewi
+  - mameshini
--- a/slack/README.md
+++ b/slack/README.md
@@ -1,0 +1,125 @@
+# Slack Guidelines
+
+Slack is the main communication platform for Kubeflow outside of our mailing
+lists. It’s important that conversation stays on topic in each channel, and that
+everyone abides by the [Code of Conduct](https://github.com/kubeflow/community/blob/master/CODE_OF_CONDUCT.md). Community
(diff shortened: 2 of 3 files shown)


slack/OWNERS
```diff
@@ -0,0 +1,4 @@
+approvers:
+  - abhi-g
```
jlewi, line 2: remove @abhi-g ; he's working on other things.
No need to list me as an OWNER since I'm in the root and don't have additional context here.

slack/README.md
```diff
@@ -0,0 +1,125 @@
...
+## Code of Conduct
+
+The Kubeflow [Code of Conduct](https://github.com/kubeflow/community/blob/master/CODE_OF_CONDUCT.md) applies throughout the
+project, and includes all communication mediums.
+
+## Admins
+
+(by Slack ID)\
+@Igor Mameshin\
+@Jeremy Lewi
```
jlewi, line 26: Don't list me; I'm trying to offload as much responsibility as I can to others.

slack/README.md
```diff
@@ -0,0 +1,125 @@
...
+community: promote your meetings, post agendas, etc.
+
+We may make special accommodations where necessary.
+
+## Escalating and/or reporting a problem
+
+Join the #admins channel or contact one of the admins via DM directly and describe the situation. If the issue can be
+documented, please take a screenshot to include in your message.
+
+## Bots, tokens, and webhooks
```
jlewi, line 69: I think we've reached the quota limits w.r.t bots

jlewi:
Thanks so much for driving this; just a couple nitpicky comments.

slack/README.md
mameshini: Sounds good, we will have to deny most requests for Slack bots.

mameshini:
I have updated the list of admins as we discussed.  There is also #admin group where all current admins can be reached on Slack.

Issue #364, opened 2020-07-14
Cancel outreach community meeting


jlewi:
Attendance for the outreach meeting is pretty low < 10 people.
https://github.com/kubeflow/community/blob/49e849fa935558583a227ccad9eecf788715843d/calendar.yaml#L81

Should we cancel this meeting?

Especially now that we are forming WGs; maybe we should chann

/cc @elviraux @jbottum @jtfogarty @animeshsingh @theadactyl

animeshsingh:
yes - can be discussed on main community calls if needed

jbottum:
@aronchick @elviraux @pdmack @kkasravi Participation in the Kubeflow PM / Outreach calls is very low.   Outbound marketing of KF by the Community is low.   It appears that we may not have a PM Working Group.    I think this structure will impact our ability to grow the Community.   I am cautious to eliminate these calls without a replacement.

jlewi:
@jbottum I have removed the call from the calendar for now. If someone would like to resume them they will needed to be added back to the calendar with a suitable host.

Issue #341, opened 2020-06-11
Additional community call (zoom) hosts needed


jlewi:
We need additional people to host community meetings. We can only give 1-2 additional people permission to use the kubeflow zoom account and access the recordings.

Current admins are listed here
https://github.com/kubeflow/community/blob/master/how-to/kubeflow_assets.md#zoom

* What is the minimum permission needed to host zoom calls and access the recordings?
* Existing [process](https://github.com/kubeflow/community/blob/master/how-to/community_meetings.md)
   *  Doc is outdated

Expectations for anyone who volunteers

* You are willing to host KF meetings 
* You are willing to sort out and document the process around doing this

/cc @animeshsingh

lalithvaka:
@jlewi, I can definitely help out as needed. Please grant me access. I will start with all the community zoom meeting recordings posted to the drive as well as the notes document. 
/cc @animeshsingh

jlewi:
@lalithvaka Could you do me a favor and look up the zoom permission model and tell me what level of access you think you need? Fewer priveleges is better.

lalithvaka:
@jlewi Per what I can see, you just need to invite me as a member with my email ID (lalith.vaka@gmail.com) and I should be able to access the cloud recordings if available. Also, each meeting thats scheduled, you can enable settings to add one or more members as host(s) - below screenshot.
![image](https://user-images.githubusercontent.com/8292387/84460379-a0809880-ac1e-11ea-8748-5ffb532b606b.png)

You can also enable auto record all meetings if they are not enabled
![image](https://user-images.githubusercontent.com/8292387/84460495-eccbd880-ac1e-11ea-882a-ac521d53149f.png)

Let me know if you need any additional info. There is a custom role(s) you can create, but not much of a use in this context. I do not have a paid zoom account to see if there are any settings / limits we can do beyond this.

jlewi:
@lalithvaka Thanks! I've sent you an invite can you please accept.

I also added @rmgogogo 

@lalithvaka  could you 

1. Send me a PR updating the assets page
1. Update the page describing how we manage zoom

Any thoughts on how we should rotate and manage who hosts a particular meeting?

lalithvaka:
[...]

PR #365, opened 2020-07-16
Calendar cleanup

### LINKED ISSUE #362: Kubeflow calendar should include kubeflow-discuss in the invite list
Right now it looks like the events on the kubeflow community calendar don't include kubeflow-discuss@googlegroups.com on the invite list.

As a result the invites aren't automatically added to the calendars of people who join the group.

People can still add the kubeflow calendar using the URl but the events won't necessarily show up on their events.

@jtfogarty @yuzisun @karlschriek Since you volunteered to be calendar admins could you please looking into fixing this?

jlewi:
* Support adding ending dates for recurring meetings.

* Per #341 set an end date for the community meetings until we
  find new hosts

* Per #364 cancel outreach and product meetings

Fix #362 - Add support for listing attendees; include kubeflow-discuss
  so that the invite is added to people's calendars.

### DIFF
--- a/calendar.yaml
+++ b/calendar.yaml
@@ -8,17 +8,24 @@
 #   date - The date of the first meeting (month/day/year).
 #   time - The start and end of the meeting in PST time zone.
 #   frequency - How often the meeting takes place.
+#   until - Optional the last day to repeat until
+#   attendees: See: https://developers.google.com/calendar/v3/reference/events/insert#python
 #   video - Zoom or Hangouts meeting link.
 #   description - Any other information about the meeting such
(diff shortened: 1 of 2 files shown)


Swikar:
#341 should I delete meeting from yaml file or should we add enddate in yaml file. This will require modifying calendar_import.py script to handle enddate. Do you have specific end date in mind?

Swikar:
@jlewi Should we add API call delete() or update() in order to modify/end event? https://developers.google.com/resources/api-libraries/documentation/calendar/v3/python/latest/calendar_v3.events.html

PR #384, opened 2020-08-04
Remove KFP community meeting from calendar. 

james-jwu:
There was a mistake in time (should be 5:30pm but entered 5:30am), but also it was scheduled for every week instead of monthly. Removing until the problem can be figured out.


Bobgy:
/assign @jlewi 
/lgtm

I also created a PR to add us as OWNERS: https://github.com/kubeflow/community/pull/385

PR #391, opened 2020-08-11
Update community calendar invite

thesuperzapper:
This PR updates the invite with the new Zoom links.


thesuperzapper:
@jlewi will this automatically fire off the new invite? Or do you need to trigger something?

PR #392, opened 2020-08-11
feat: Update caicloud team

gaocegege:
Caicloud is acquired by Bytedance, thus we are going to update the team to Bytedance.

/assign @jlewi 

Is there anything we need to do? I am not sure if we need to rename the team manually.

Signed-off-by: Ce Gao <gaoce.gaoc@bytedance.com>

### DIFF
--- a/member_organizations.yaml
+++ b/member_organizations.yaml
@@ -62,13 +62,13 @@
     - ifilonenko@bloomberg.net
   url: http://www.bloomberg.com
 
-- name: Caicloud
+- name: Bytedance
   contacts:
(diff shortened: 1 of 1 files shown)


jlewi:
@james-jwu and @Bobgy Will be the primary maintainers so one of them should review.

/assign @james-jwu

gaocegege:
Thanks

/assign @james-jwu @Bobgy

Bobgy:
/lgtm
/approve

I don't know if there's anything else.
/cc @jlewi

PR #408, opened 2020-09-01
Add a Feature Store SIG

woop:
Adding a feature store special interest group to drive feature store development in Kubeflow, and to take ownership of Feast as a component after its introduction in [Kubeflow 1.1](https://www.kubeflow.org/docs/components/feature-store/).

Person | WG | Response
-- | -- | --
@animeshsingh | KFServing, Pipelines |  
@cliveseldon | KFServing |  
@yuzisun | KFServing |  
@ellistarn | KFServing |  
@neuromage | Pipelines/Metadata |  
@paveldournov | Pipelines/Metadata |  
@elikatsis | Pipelines |  
@vpavlin | Control plane |  
@yanniszark | Control Plane/common services |  
@Jeffwan | Control Plane |  
@krishnadurai | Common Services |  
@terrytangyuan | Training |  
@gaocegege | Training |  
@jlewi | Training |  
@johnugeorge | Training |  
@aronchick | Metadata |  
@StefanoFioravanzo | Metadata |  
@jbottum | PM |  
@elviraux | PM |  
@kimwnasptd | Notebooks |  
@krazyhaas |


jlewi:
Thanks @woop do you think a SIG (#409) might be better than a WG? We can always start with a SIG and change to a WG later on. 

Primary difference is that WGs own code and repositories. Since Feast is currently developed outside of Kubeflow it wouldn't be owned by a KF WG.

I think a SIG would be the right forum to coordinate cross cutting issues with respect to feature management, data, and Kubeflow.

woop:
> Thanks @woop do you think a SIG (#409) might be better than a WG? We can always start with a SIG and change to a WG later on.
> 
> Primary difference is that WGs own code and repositories. Since Feast is currently developed outside of Kubeflow it wouldn't be owned by a KF WG.
> 
> I think a SIG would be the right forum to coordinate cross cutting issues with respect to feature management, data, and Kubeflow.
> 

Sure, makes sense.

> I think to propose a SIG rather than WG just grep/replace WG->SIG

Do you mean that literally for this PR as well? I mean the complete structure is identical?

jlewi:
> Do you mean that literally for this PR as well? I mean the complete structure is identical?

Yes in this PR references to "wg" e.g. `wg-feature-store-leads` should be changed to `sig-feature-store-leads`.


@kubeflow/wg-automl-leads @kubeflow/wg-training-leads @kubeflow/wg-serving-leads  WDYT about a SIG related to feature store/data?

wg-feature-store/charter.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Scope
+
+WG Feature Store covers the definition, management, storage, discovery, and serving of features to models.
+
+### In scope
+
+#### Code, Binaries and Services
+
+- Provide a unified feature serving layer for both model training and online serving.
```
vpavlin, line 13: Is the word serving correct? (Since the WG/SIG is called Feature **Store**)

woop: Yes, it is correct. Providing a consistent interface to accessing data in both training and online serving is one of the essential parts of a feature store.

woop: Happy to use a different term like "access" or "retrieval" if we want to disambiguate it.

woop:
@jlewi I've updated the PR, although some of the reference documentation doesn't exist for SIGs. Should that be defined anew, or should I just reference the `wg-governance` for example?

wgs.yaml
```diff
@@ -1,4 +1,44 @@
...
+    tech_leads:
+    - github: pyalex
+      name: Oleksii Moskalenko
+      company: Gojek
+    - github: woop
+      name: Willem Pienaar
+      company: Gojek
+  meetings:
+  - description: Monthly SIG Meeting
+    day: Wednesday
```
animeshsingh, line 27: so the meeting cadence is once a month?

animeshsingh (commented):
/lgtm

wgs.yaml
woop: Yes, if that is acceptable. I would rather start slower and have more fruitful meetings.

sig-feature-store/charter.md
```diff
@@ -0,0 +1,42 @@
...
+
+This charter adheres to the conventions, roles and organization management outlined in [wg-governance].
+
+## Scope
+
+SIG Feature Store covers the definition, management, storage, discovery, and serving of features to models.
+
+### In scope
+
+#### Code, Binaries and Services
```
jlewi, line 11: @woop I would suggest rewording this section a bit. The SIG doesn't technically own these artifacts since they are being developed outside of Kubeflow; eg. in Feast.

I would define the scope as more along the lines of coordinating between the various projects using or depending on feature stores.

woop: I've taken a stab at rewording it. Mind taking another look?

woop: https://github.com/kubeflow/community/pull/408/commits/fc86d5c151eb28908255020b8dbef94100abf6cf

sig-feature-store/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Technical Leads
+The Technical Leads of the SIG establish new subprojects, decommission existing
+subprojects, and resolve cross-subproject technical issues and decisions.
+
+* Oleksii Moskalenko (**[@pyalex](https://github.com/pyalex)**), Gojek
+* Willem Pienaar (**[@woop](https://github.com/woop)**), Gojek
+
+## Contact
+- Slack: [#feast](https://kubernetes.slack.com/messages/feast)
```
jlewi, line 35: @woop Slack channel doesn't look right; Do you mean kubeflow?

jlewi:
@woop One thing I just want to make sure you are aware of. The KF community is currently having a lot of growing pains in terms of maintaining community infrastructure. For example

  * Sync'ing of community calendar kubeflow/community#274
  * We have no scalable process for creating mailing lists for sigs/wgs

@paveldournov @theadactyl WDYT about this proposal?

sig-feature-store/README.md
woop: Good catch. Fixed.

woop:
> @woop One thing I just want to make sure you are aware of. The KF community is currently having a lot of growing pains in terms of maintaining community infrastructure. For example
> 
> * Sync'ing of community calendar #274
> * We have no scalable process for creating mailing lists for sigs/wgs
> 
> @paveldournov @theadactyl WDYT about this proposal?

Thanks for taking a look @jlewi 

Do you mean the tasks above require human action? Are there any particular challenges?

Also, should `feast` be removed as a sub-project for this SIG, or should it be kept in?

kevinstumpf:
I'd love to contribute to this SIG

jlewi:
> Also, should feast be removed as a sub-project for this SIG, or should it be kept in?

Yes; SIGs don't own projects so they shouldn't have sub projects.

> Do you mean the tasks above require human action? Are there any particular challenges?

It means there is no SLA and you need to plan accordingly. For a SIG, the risk is pretty minimal because creating your own calendar or mailing list isn't too difficult.

For WGs its more of an issue because there is a lot of dependency on shared test/release infra (see kubeflow/testing#737).

jlewi:
/lgtm
/approve

/hold 
Lets give @paveldournov and @theadactyl another day to chime in. If they don't chime in by then we can approve via lazy consensus.

woop:
> I'd love to contribute to this SIG

Added to SIG. Just need to have your ACL merged I think (https://github.com/kubeflow/internal-acls/pull/342)

> /lgtm
> /approve
> 
> /hold
> Lets give @paveldournov and @theadactyl another day to chime in. If they don't chime in by then we can approve via lazy consensus.

Sgtm

btw, lgtm and approval are removed now because I added @kevinstumpf

sig-feature-store/README.md
```diff
@@ -0,0 +1,52 @@
...
+- [Open Community Issues/PRs](https://github.com/kubeflow/community/labels/sig%2Farea/sig-feature-store)
+- GitHub Teams:
+    - [@kubeflow/sig-feature-store](https://github.com/orgs/kubeflow/teams/sig-feature-store) - Feature store development and roadmap
+
+## Subprojects
+
+The following subproject(s) are owned by sig-area/sig-feature-store:
+### feast
+- **Owners:**
+  - https://github.com/feast-dev/feast/blob/master/OWNERS
```
gaocegege, line 48: Will feast donate to Kubeflow?

woop: It's a bit overloaded as is because we are hosting our #feast channel in Kubeflow.org Slack. I wanted to make sure that there is a group taking responsibility of the project from the Kubeflow perspective, but mostly in terms of integration.

There wouldn't be a contribution/donation into the Kubeflow org.

@jlewi @gaocegege should we remove that line?

thesuperzapper: I am pretty sure @jlewi said here: https://github.com/kubeflow/community/pull/408#issuecomment-692396749 that SIG cant have assets they own. (And that makes sense)

gaocegege: I think it may be confusing for readers since the project is not owned by Kubeflow. But I have no objection to it.

thesuperzapper: I am saying to remove it, because it a SIG cant have a sub-project?

gaocegege: Oh SGTM.

thesuperzapper: @woop ?

woop: @thesuperzapper @gaocegege sgtm, removed!

woop:
I think we need another /lgtm

btw, can somebody please confirm that @kevinstumpf has received a membership invitation after having his ACL added?

PR #412, opened 2020-09-15
fix(calendar): fixed Kubeflow community call (US East/EMEA)

Bobgy:
/assign @james-jwu @jlewi 

I tried to find the existing Kubeflow community call (US East/EMEA), but the calendar event says "canceled or deleted" when I open the web url.

The easiest fix I figured out is to use a new event id so that a new event is created.
I've tried multiple approaches to restore the old event, but it doesn't exist in event trash. I guess it's been deleted for too long.

I've already run the scripts, so the new event is already showing up in calendars.

### DIFF
--- a/scripts/calendar_import.py
+++ b/scripts/calendar_import.py
@@ -99,7 +99,8 @@ def update_meeting(service, meeting):
 
   try:
     event = service.events().insert(calendarId=CALENDAR_ID, body=event).execute()
-    logging.info("Event created: {}".format(event.get('htmlLink')))
+    logging.info("Craeted Event: {}".format(meeting['name'][:100]))
+    logging.info(event.get('htmlLink'))
   except googleapiclient.errors.HttpError as e:
     content = json.loads(e.content)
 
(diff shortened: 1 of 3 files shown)


scripts/calendar_import.py
```diff
@@ -99,7 +99,8 @@ def update_meeting(service, meeting):
 
   try:
     event = service.events().insert(calendarId=CALENDAR_ID, body=event).execute()
-    logging.info("Event created: {}".format(event.get('htmlLink')))
+    logging.info("Craeted Event: {}".format(meeting['name'][:100]))
```
Bobgy, line 102: improve logging to make finding events easier

Bobgy:
Strange, although the calendar shows in kubeflow calendar and invites kubeflow-discuss@googlegroups.com, I'm not getting the calendar invite on my own calendar. I'm suspecting I don't have enough permissions: https://support.google.com/calendar/answer/172013?hl=en

Bobgy:
OK, everything working correctly now.
/approve

jlewi:
Thanks!
/lgtm
/approve

PR #419, opened 2020-10-04
Add Feature Store SIG meetup to calendar

woop:


jlewi:
@nicholas-abad and @yuzliu PTAL

PR #427, opened 2020-10-21
[WIP] Add WG Testing

### LINKED ISSUE #426: Create WG-testing
We have a few discussion in past two month on the approach owning and maintaining test/release infrastructure. 
@PatrickXYS and I will like to proposal a WG to handle common test/release infra. 

The WG-testing proposal will come later. 

/cc @PatrickXYS

PatrickXYS:
Added a WG Testing

resolves #426


PatrickXYS:
/cc @Jeffwan @andreyvelich 

This is still in draft, feel free to leave comments and I can improve the doc.

jlewi:
I've been thinking about this; does Kubeflow really need a testing WG?

Are we creating an unnecessary (and unsustainable) alternative to GitHub?

An alternative would be
  * Each WG is responsible for its own test/release infra
  * Each WG can pick whatever solution it is willing and able to ;
       * GitHub Actions, AWS Prow, CNCF Prow, Google Prow, etc..

If AWS wants to create a "community" prow to be used by OSS projects and offer it to Kubeflow and other projects; I think that's great. 

But should this be done outside of Kubeflow? 

/cc @kubeflow/wg-automl-leads 
/cc @kubeflow/wg-pipeline-leads 
/cc @kubeflow/wg-serving-leads 
/cc @kubeflow/wg-notebook-leads

johnugeorge:
@PatrickXYS For clarity, what is the target/roadmap for this WG? Are all the WGs in the org supported in this effort? If yes, What is the expectation from each WG to be part of it?  And Is it optional if a WG wants to handle testing effort independently?

PatrickXYS:
> If AWS wants to create a "community" prow to be used by OSS projects and offer it to Kubeflow and other projects, I think that's great.

We want to start from a smaller scope, which is we want to provide a "kubeflow-org" prow to be used by kubeflow WGs. For larger scope, we can talk about it later.

/cc @jlewi 

>what is the target/roadmap for this WG? 

The goal of this WG is to provide and manage a shared test-infra for kubeflow community. Any WG wants to migrate to the shared test-infra; they can follow our instructions to do so.

> Are all the WGs in the org supported in this effort? If yes, What is the expectation of each WG to be part of it?

This is two different levels of effort; test-infra is divided into 2 parts, the common infra composed of Prow + Argo, and test cases written to be running on the infra. The WG-testing will be focusing on the common infra part; other WG will be focusing on the test cases.

>And Is it optional if a WG wants to handle testing effort independently?

Yes, this is not mandatory for all the WG; they can choose to handle testing effort by themselves.

/cc @johnugeorge

jlewi:
I have reservations about a WG owning centralized/shared infra. This creates a single point of failure. If at some point in the future maintaining the WG becomes unfeasible then multiple WGs would be impacted and it becomes a huge problem to fix. 

In some sense this is the meta issue we are seeing with #737. Google was maintaining shared infra and this worked really well until KF outgrew what Google was willing to support. If all/multiple WGs end up depending on the shared AWS infra then we could end up being in the same boat.

I think a better approach would be to make each WG responsible for its own test/release infra.

I think this makes KF more resilient and ensures that the support/ops costs get evenly distributed among the WGs.

So it makes sense that for WGs where AWS has a leadership role (e.g.training) AWS provides the infra and for WGs where GCP  has a leadership role (e.g. pipelines) GCP provides the infra.

This also ensures that the leads have skin in the game. If the leads of a WG are unable to come up with a solution for test/release infra that seems like a sign that the leads may not have the resources and willingness to sustain their projects long term.

Jeffwan:
@jlewi I agree. From past experiences, centralized solution is not a scalable way if community lack of resources to maintain it. I think from AWS perspective, at the beginning, it wants to test each Kubeflow distribution and some critical components. AWS likes to move one further step to provide it as an alternative testing solution that each WG can leverage.
It could be done outside Kubeflow.  major reason to propose this inside Kubeflow is

1. AWS can not use internal accounts for OSS projects. It needs an entity to host an account and AWS can give credits to it. In this case, it would be Kubeflow testing WG. This is done similarly for CNCF projects like Kubernetes.
2. Leverage community knowledges and don't reinvent the wheel. Folks who work on `Kubeflow/testing` did awesome jobs and community already have some assets. It would be great to keep working on this. If people still like to do these work for their Kubeflow project, they still have cloud resources to do that.

I think AWS is pretty open to do it whether do it inside/outside Kubeflow.

jlewi:
> It needs an entity to host an account and AWS can give credits to it.

Will Kubeflow work for this purpose? Kubeflow WGs aren't actually setup to accept monetary donations and this would likely create all sorts of problems.  Who would you actually give the credits to? Kubeflow (e.g. kubeflow.org) is owned by Google.

If you already have a relationship with CNCF would that be a better option? Should the leads that are interested in relying on this shared infra think about forking and donating themselves to the CNCF? 

> Leverage community knowledges and don't reinvent the wheel. Folks who work on Kubeflow/testing 

I agree we should continue to collaborate. I've been thinking of how to do this effectively.

Right now kubeflow/testing is a loose collection of adhoc scripts that encode a particular way of doing things (e.g. using ksonnet, then python, then argo, then tekton). This is tightly coupled to the preferences of whatever project is using those scripts for testing.

I think this is going to be a barrier for collaboration because we don't have well established boundaries (e.g. applications, libraries, modules).

This will probably only get worse as the WGs take on more responsibility and diverge. 

I think to collaborate effectively, we will need to start defining sub-projects that can be developed relatively independent of the projects that are relying on them. As an example

* GitOps tooling for setting webhooks
* Resuable tooling (e.g. custom resources) for CD pipelines

What if we asked existing WGs to incubate/sponsor any such projects? We can then consider spinning up a WG when we have a critical mass of sufficiently mature applications.

Projects can continue to live in kubeflow/testing and we can use OWNERs file to indicate ownership.

My personal hope is that  forming a WG will never be necessary because any infra we build is just a stop-gap until the functionality is superceded by upstream functionality.

* e.g. Prow is adding Tekton integration so I'm hoping we can eventually eliminate run_workflow_e2e.py
* Hopefully, prow (or GitHub) will provide more GItOps functionality for GitHub so we won't have to.

Jeffwan:
@jlewi 

> If you already have a relationship with CNCF would that be a better option? Should the leads that are interested in relying on this shared infra think about forking and donating themselves to the CNCF?

make sense. Let us handle this problem internally. 


> What if we asked existing WGs to incubate/sponsor any such projects? We can then consider spinning up a WG when we have a critical mass of sufficiently mature applications.
> Projects can continue to live in kubeflow/testing and we can use OWNERs file to indicate ownership.

Sounds good. Let's continue develop kubeflow/testing. If any existing WGs would like to incubate or sponsor those projects, then we can come back to revisit if we need to spin up a WG. 


> My personal hope is that forming a WG will never be necessary because any infra we build is just a stop-gap until the functionality is superceded by upstream functionality.
> e.g. Prow is adding Tekton integration so I'm hoping we can eventually eliminate run_workflow_e2e.py
> Hopefully, prow (or GitHub) will provide more GItOps functionality for GitHub so we won't have to.

Makes sense. I had some offline discussion with @PatrickXYS. I think these are very reasonable real world feature requests in Kubernetes community and most of our team member are making contribution in different sigs. For long term, we will try to resolve these issues in Kubernetes community and then bring back to Kubeflow. That will make bigger impact and save efforts here to develop one-off solutions. 

/cc @PatrickXYS

jlewi:
@PatrickXYS and @Jeffwan are we in agreement then that we should put the WG testing on hold for now and close this PR?

Jeffwan:
@jlewi 

Yes.  @PatrickXYS, let's close the PR for now and revisit it when necessary

PR #428, opened 2020-10-22
Automatically sync the calendar

jlewi:
* Update the code to work with service accounts

  * We need to use domain delegation when using a service account otherwise
    we get an error when trying to add attendees

  * With a service account there's no need to pickle credentials and there's
    no need to specify an oauth client as each GSA has an implicit oauth client.

* Add an option to run continuously

  * Continuously check if there are changes to the file and if there are
    run the calendar sync.

* Create a Makefile to build the image
* Create a Kubernetes deployment which runs git-sync in a sidecar.
Related to #274

### DIFF
--- a/calendar/Dockerfile
+++ b/calendar/Dockerfile
@@ -1,0 +1,16 @@
+FROM google/cloud-sdk
+
+RUN apt-get update -y && \
+    apt-get install -y python3-distutils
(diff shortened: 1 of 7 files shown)


Bobgy:
Thank you so much for building this!

PR #431, opened 2020-10-22
calendar_import.py should use OAuth credentials for autobot@kubeflow.org

jlewi:
* We don't want to use a google service account because that requires
  domain wide delegation.

* Instead we change the script to impersonate autobot@kubeflow.org

  * We do this by generating a refresh credential for autobot@kubeflow.org
    and saving it as a GCP secret. A human needs to do this as the human
    needs to go through the OAuth Web Flow

* When running the sync we read the OAuth credential from the GCP secret
  and impersonate autobot@kubeflow.org

* We use a GSA and workload identity to access secret manager.

* Move to the kf-autobot namespace.

Fix: #274

### DIFF
--- a/calendar/Dockerfile
+++ b/calendar/Dockerfile
@@ -7,6 +7,7 @@ RUN python3.7 -m pip install \
 	fire==0.3.1 \
 	google-api-python-client==1.12.4 \
 	google-auth-oauthlib==0.4.1 \
+	google-cloud-secret-manager==2.0.0 \
 	oauth2client==4.1.3 \	
 	python-dateutil==2.8.1 \
 	pyyaml==5.3.1 \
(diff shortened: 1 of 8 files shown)


Bobgy (commented):
This is awesome! 👍
A readme for how we set up the credentials will help us maintain the infra.

As I recall, refresh tokens expire every 6 months, is that right?
We'll need to redo the set up after it expires.

jlewi:
Looks like my README.md changes weren't saved; that's frustrating.

jlewi:
README updated. PTAL

Bobgy:
Thank you so much!
This is absolutely a huge relief for me having calendar sync running automatically and the instructions are clear and easy to follow.

/lgtm
/approve

PR #402, opened 2020-08-27
create wg-deployment

swiftdiaries:
Represent OWNERs from manifests, kfctl as chairs, tech leads.
If you feel you need to be present in the WG, please comment on the PR.

The `wg-control-plane` can be superset of `wg-deployment`. I wanted to keep things very narrow for scope of this working group. And will be subsequently formed.

Related to: #400

PR #434, opened 2020-10-27
Proposal future ownership/development of Kubeflow distributions should be outside Kubeflow

jlewi:
There has been significant debate about what to do about Kubeflow distributions

See: #402 
[Thread](https://groups.google.com/g/kubeflow-discuss/c/d6whgEgror8)
[Arrikto Proposal](https://docs.google.com/document/d/1BpR4RTp4m2GO_r-c3_xZ-qGnCfvQdqiRh4V0cFUSBqA/edit?ts=5f96e397)

I believe this proposal captures the conensus among the leads.

/cc @kubeflow/project-steering-group 
/cc @kubeflow/wg-notebook-leads 
/cc @kubeflow/wg-pipeline-leads 
/cc @kubeflow/wg-automl-leads 
/cc @kubeflow/wg-training-leads 

/cc @cvenets @yanniszark @RFMVasconcelos @animeshsingh @paveldournov @johnugeorge @krishnadurai @Jeffwan

PR #435, opened 2020-10-27
wg-manifests: Introduce WG Manifests

yanniszark:
As discussed in today's community meeting and after lengthy discussions in:
- https://groups.google.com/g/kubeflow-discuss/c/d6whgEgror8
- https://github.com/kubeflow/community/pull/402
- https://docs.google.com/document/d/1BpR4RTp4m2GO_r-c3_xZ-qGnCfvQdqiRh4V0cFUSBqA

please find here the PR for the Manifests WG (previously called Control Plane).
Please take a look at the PR files for a complete explanation of the scope and responsibilities.

/cc @jlewi @PatrickXYS @Jeffwan @StefanoFioravanzo @elikatsis @vkoukis @cvenets


wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+        - PodDefaults Controller
+- Maintain documentation and instructions for installing a set or all of the
+  apps. Documentation will not depend on any opinionated deployment tool. It
+  will only depend on `kustomize` and `kubectl`. This way, users and platform
+  owners can easily build on top and use their favorite opinionated deployment
+  tool (e.g., Argo, kapp, kpt, kfctl, FluxCD, etc.) with Kubeflow.
+- The maintained manifests will follow certain principles:
+    - The flow of work / separation of responsibilities will be the following:
+        1. Application owners publish manifests in their repos.
+        1. WG copies and tracks upstream manifests in the manifests repo. They
```
PatrickXYS, line 46: I would assume an automated way to do this, at least in master branch?

yanniszark: @PatrickXYS you have selected 3 lines, so I'm not sure to what you're referring 😅 
Can you please clarify, perhaps by quoting the part the question is referring to?

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+          procedure where upstream changes can easily find their way in
+          `kubeflow/manifests`. Thus, it suffered from poor maintainability.
+    - Users (ML Engineers) and platform owners can extend and re-use
+      kustomizations to integrate better with their platforms.
+
+#### Cross-cutting and Externally Facing Processes
+
+##### With Application Owners
+
+- Communicate with application owners to agree upon the version they want to be
```
PatrickXYS, line 70: I think this part, should we include PM team as well, since we rely on them about cross-project conversations.

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+    1. Manifests for a set of specific common services:
+       - Istio
+       - KNative
+       - Dex
+       - Cert-Manager
+    1. Manifests for common, currently unmaintained, apps:
+        - Central Dashboard
+        - Profile Controller
+        - PodDefaults Controller
+- Maintain documentation and instructions for installing a set or all of the
```
PatrickXYS, line 38: Though we should not rely on vendor distribution, I think we still need to deploy applications on a "generic" kubernetes system, what would this be?

wg-manifests/charter.md
PatrickXYS: Oh sorry, should be this line

```
1. WG copies and tracks upstream manifests in the manifests repo. They
form a base `kustomization`.
```

thesuperzapper:
@yanniszark Just to be explicit, assuming the discussion in https://github.com/kubeflow/community/pull/434 decides that we WILL have a "generic" distribution, will the Manifests WG be responsible for it?

wg-automl/charter.md
```diff
@@ -19,7 +19,7 @@ The WG covers researching, developing and operating various targets of ML automa
 #### Cross-cutting and Externally Facing Processes
 
 - Coordinating with Training WG to make sure that all distributed training jobs can be used in AutoML experiments.
```
jlewi, line 21: > Coordinating with Training WG to make sure that all distributed training jobs can be used in AutoML experiments.

Why is this in scope for this WG?

wg-automl/charter.md
```diff
@@ -19,7 +19,7 @@ The WG covers researching, developing and operating various targets of ML automa
...
-- Coordinating with Control Plane WG to ensure that AutoML manifests are properly deployed with Kubeflow.
+- Coordinating with Manifests WG to ensure that AutoML manifests are properly deployed with Kubeflow.
```
jlewi, line 22: > Coordinating with Manifests WG to ensure that AutoML manifests are properly deployed with Kubeflow.

Why would this be necessary?

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
+# WG Manifests Charter
+
+This charter adheres to the conventions, roles and organization management
+outlined in [wg-governance].
+
+## Scope
+
+- Provide a catalog (centralized repository) of Kubeflow application manifests.
+- Provide a catalog of third-party apps for common services.
+- Provide documentation to help users install a set, or all of the included
```
jlewi, line 10: > Provide documentation to help users install a set, or all of the included
  apps.

Please remove this from the scope of the charter.

The primary goal with forming WGs right now is to reduce ambiguity by clarifying ownership of existing assets. 

This  is not an existing asset.

Furthermore, my expectation is that application owners will be responsible for instructions to install their application standalone.

Distribution owners will be responsible for providing instructions to install their distributions.

Can we see how that goes before expanding the scope of this WG? If we find that leaves a gap and this WG ends up addressing the gap then we can expand the scope later on.

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+- The maintained manifests will follow certain principles:
+    - The flow of work / separation of responsibilities will be the following:
+        1. Application owners publish manifests in their repos.
+        1. WG copies and tracks upstream manifests in the manifests repo. They
+           form a base `kustomization`.
+        1. All kubeflow-specific changes (e.g., change the namespace) are done
+           in `kustomize` overlays and they don't touch the upstream files.
+        1. Periodically, the upstream manifests are updated by copying from a
+           later commit.
+    - Contrasted with the current state, in the aforementioned workflow:
```
jlewi, line 52: Please move this out of the charter and into the PR description.

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+        - App owners are not required to do anything other than maintain working
+          installation manifests in their repo. This is something all app owners
+          already do, as it's needed for testing and developing the app.
+        - Manifests depend on the frequently updated manifests of the app repo,
+          instead of the out-of-date, kfctl-specific manifests in the current
+          `kubeflow/manifests` repo. [A recent proposal](https://www.google.com/url?q=http://bit.ly/kf_kustomize_v3&sa=D&ust=1603724692328000&usg=AOvVaw2qgtPzKUz5zqIjpn3Yoas7)
```
jlewi, line 58: Please move this into the PR description it doesn't belong in the charter.

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+      kustomizations to integrate better with their platforms.
+
+#### Cross-cutting and Externally Facing Processes
+
+##### With Application Owners
+
+- Communicate with application owners to agree upon the version they want to be
+  included in the next Kubeflow release.
+
+##### With Platform Owners
```
jlewi, line 73: Platform -> Distribution

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+        - Kubeflow Pipelines (KFP)
+        - Notebooks
+        - KFServing
+        - Katib
+    1. Manifests for a set of specific common services:
+       - Istio
+       - KNative
+       - Dex
+       - Cert-Manager
+    1. Manifests for common, currently unmaintained, apps:
```
jlewi, line 34: Why are you distinguishing between "unmaintained" and "maintained" apps? If your just copying the manifests from the upstream repo to this repo then isn't it the same?

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+- Provide a catalog (centralized repository) of Kubeflow application manifests.
+- Provide a catalog of third-party apps for common services.
+- Provide documentation to help users install a set, or all of the included
+  apps.
+
+### In scope
+
+#### Code, Binaries and Services
+
+- Maintain a set of manifests that will allow users to install Kubeflow apps and
```
jlewi, line 17: Are you actually maintaining the manifests? I thought it would be the application owners that would be maintaining the manifests?

My expectation is that you would be maintaining tooling to automate the copying of the manifests.

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+  apps. Documentation will not depend on any opinionated deployment tool. It
+  will only depend on `kustomize` and `kubectl`. This way, users and platform
+  owners can easily build on top and use their favorite opinionated deployment
+  tool (e.g., Argo, kapp, kpt, kfctl, FluxCD, etc.) with Kubeflow.
+- The maintained manifests will follow certain principles:
+    - The flow of work / separation of responsibilities will be the following:
+        1. Application owners publish manifests in their repos.
+        1. WG copies and tracks upstream manifests in the manifests repo. They
+           form a base `kustomization`.
+        1. All kubeflow-specific changes (e.g., change the namespace) are done
```
jlewi, line 48: I don't understand this. What are "kubeflow-specific" changes? If an application should be installed in the "kubeflow" namespace shouldn't that be part of the application manifest and maintained in the upstream repo.

jlewi:
> @yanniszark Just to be explicit, assuming the discussion in #434 decides that we WILL have a "generic" distribution, will the Manifests WG be responsible for it?

No. This is out of scope. We have been debating WG ownership for over two months in #402. Distributions will not be in scope for WGs

jlewi:
@yanniszark Thanks for writing this up.

High level comment: I would like to err on the side of defining work groups as narrowly as possible. We can expand scope later .on. 

I'd like the scope of this WG to be narrowed to in line with what we had in #402. 

Please remove the bit about installation instructions; see my more detailed comment. Nothing precludes you from contributing to documentation. We can always expand the scope of the WG later on to include it when the following conditions are met
 
   1.  We identify an unmet need
   1.  We identify a solution that requires ownership

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+- Communicate with application owners to agree upon the version they want to be
+  included in the next Kubeflow release.
+
+##### With Platform Owners
+
+- Coordinate with platform owners, to make sure they are in-sync about the
+  release schedule and have time to test and bring their platforms up-to-date.
+
+### Out of scope
+
```
jlewi, line 79: Replace "platform" with "distribution"

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+This WG is NOT going to:
+- Maintain deployment-specific tools like `kfctl`.
+- Maintain platform-specific manifests.
```
jlewi, line 82: Where do distribution overlays get defined? See 
https://github.com/kubeflow/community/pull/402#issuecomment-682704600

I think what we suggested that distribution overlays should be defined upstream in the apps repository or downstream in a distribution repo.

If its defined in the app repo would it be preserved when it gets copied over to the manifests repo?

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+- Support configurations with environment-specific requirements, like special
```
jlewi, line 83: Please add 

> Create and maintain one or more Kubeflow distributions
as out of scope

jlewi:
To provide a bit more context for my perspective. The critical problem that I want to address is ensuring that app owners take responsibility for their manifests. My expectation from day 0 was that application owners were taking responsibility and curating/updating their manifests as needed.

There are various discussions that indicate to me that this is not widely understood yet. For example,

* kubeflow/manifests#1553 indicates to me that not all app owners have fully taken ownership yet of their manifess
* Likewise all this discussion about "base" vs. "base_v3" vs. "kfctl dependencies" indicates to me a confusion about responsibility
   * If app owners don't want their manifests to be structured that way then they should change them.

This is why I was pushing for the scope to be so narrowly defined in #402. Language like the following

> Maintain a set of manifests that will allow users to install Kubeflow apps and

Creates ambiguity about division of responsibilities. Is it the app owners or the manifests WG responsible for maintaining the manifest? 

My hope was to address this by having very narrow and well defined scope and responsibilities

* App owners maintain manifests in their repos
* manifests-wg maintains automation to copy them to kubeflow/manifests

The division of responsibilities is then very clear. In particular, going back to kubeflow/manifests#1553  this would be handled as follows

* Look in the upstream repo; if the image is outdated in the upstream repo the app-owner is responsible
* If the manifest in kubeflow/manifests is not in sync with the upstream then its a bug in the wg's tooling

I would like to see this functioning as intended before expanding scope beyond this very narrow definition. In particular, I'd like to see

1. Every WG maintains its manifests in its upstream repo
1. This WG produces automation to keep the manifests in kubeflow/manifests in sync with the upstream repos

cvenets:
@jlewi I agree with almost all of your comments except for the documentation part.

If we are to automate the whole process of syncing manifests to this repo and then testing them with the help of AWS on the shared infra, shouldn't we have a single page on the documentation to help users on how to install them? Even if they want to install just a set of the apps, and not everything?

In my opinion, currently the users are really struggling, even if they want to install a single app. I think a single page of documentation doesn't really expand the scope of the WG at all. This page can actually provide an intro with a bullet list of the actual catalog as well, so that people at least understand what they can find in the manifests.

We can update the PR to explicitly say that this is going to be a single page on the documentation if that's helpful, but I don't really see the need.

jlewi:
> If we are to automate the whole process of syncing manifests to this repo and then testing them with the help of AWS on the shared infra, shouldn't we have a single page on the documentation to help users on how to install them? Even if they want to install just a set of the apps, and not everything?

Nothing precludes you from contributing to the documentation. There are a variety of discussions about how to better structure the docs: see kubeflow/website#2293. As part of that discussion if we decide that we need a page of instructions for X then we can create such a page and use the OWNERs file to establish who will be responsible for owning and maintaining it.

My concern with including documentation in the scope of the PR; e.g.

> We can update the PR to explicitly say that this is going to be a single page on the documentation if that's helpful, but I don't really see the need.

is that its premature. We haven't figured out how to restructure the docs. I think we need to do that wholistically and ideally not be bound by existing commitments.

I think kubeflow/website#2293 as opposed to the WG charter is a much better place to have discussion about our docs. I think figuring out how to properly restructure our docs and how to properly maintain them is a much bigger issue.

wg-automl/charter.md
Jeffwan: I think autoML covers this part.  https://github.com/kubeflow/community/blame/master/wg-automl/charter.md#L21
It can be removed here.

wg-automl/charter.md
yanniszark: > Why would this be necessary?

@jlewi those are not new scopes, but pre-existing ones.
I simply amended the name since WG Control Plane doesn't exist anymore.

wg-manifests/charter.md
yanniszark: @jlewi what I mean is that some applications are installed in a slightly different ways sometimes.
For example, KFServing in manifests used the kubeflow namespace, while the standalone in the upstream repo used the kfserving-system namespace.

Now, this particular example no longer applies, as KFServing now also maintains a kubeflow overlay which deploys in the `kubeflow` namespace, but the fact remains.

So I will use another example. The user identity header expected by an application is not the same across the board.
- Jupyter Web App expects `kubeflow-userid` (https://github.com/kubeflow/kubeflow/blob/6eb007d0b7b37ece0eb9a072f12274aab0f8fe09/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/authn.py#L12-L13)
- Pipelines expects `x-goog-authenticated-user-email` (https://github.com/kubeflow/pipelines/blob/5435e8724fe1b2c6175f219217c82dee33227b4b/frontend/server/configs.ts#L99)

So for Pipelines and JWA to work together, we have to tweak one of the two slightly.
This change can happen in an overlay, maintained by wg-manifests, so that the user can install pipelines, jwa, or the two of them together.

Ideally, everything will be kind of uniform on the app side, so we won't have to maintain these, but unfortunately this is not always the case. We want to be able to do these minor fixes if needed.

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
...
+
+- Communicate with application owners to agree upon the version they want to be
+  included in the next Kubeflow release.
+
+##### With Platform Owners
+
+- Coordinate with platform owners, to make sure they are in-sync about the
+  release schedule and have time to test and bring their platforms up-to-date.
+
+### Out of scope
```
jlewi, line 78: Please include the following in out of scope

* Deciding which applications  to include in KF
* Deciding which variant of an application to include (e.g. KFP standalone vs. KFP with ISTIO)

wg-manifests/charter.md
jlewi: Thanks @yanniszark Per [comment](https://github.com/kubeflow/community/pull/402#issuecomment-682704600) I would expect this to work as follows.

IIUC the header will be different depending on the distribution. With Google IAP the header gets set to "x-goog-authenticated-user-email" with Dex or Cognito I presume its a different a header.

Its up to each application to make the header configurable; e.g. provide an environment variable that allows the header to be set.

It would then be up to the distribution to set the environment variable correctly for all applications.

So lets suppose you wanted to define an overlay to make the environment variable configurable; Per [comment](https://github.com/kubeflow/community/pull/402#issuecomment-682704600) I  think we should pursue one of the following

1. We push that overlay upstream into the app's repo upstream of manifests
1. We push the overlay downstream into the distribution's repo

How does that compare to what you are thinking?

cvenets:
@jlewi I understand that restructuring of the website is a bigger discussion, and we should definitely have it separately. It is currently unclear to me who really owns the website at this point, and who has commit rights. Is the website going to be maintained by a WG or not?

I think the context for this WG is to establish that we all agree we need that single page of instructions, and that the maintainers of the manifests are accountable of keeping this page up-to-date after they test that everything works, so that users can actually use the page to install their preferred set of KF apps. This is a significant problem for users right now trying to install.

Where this page should fall under in the structure of the website/docs is indeed a separate discussion.

But I think this needs to stay and we should all agree that a page like this is definitely needed. Else we will keep having manifests that don't work with each other and users complaining they can't find proper instructions to install.

jlewi:
> @jlewi I understand that restructuring of the website is a bigger discussion, and we should definitely have it separately. It is currently unclear to me who really owns the website at this point, and who has commit rights. Is the website going to be maintained by a WG or not?

Commit rights are specified by OWNER files. There is significant effort underway to ensure all pages have identified OWNERs.  @RFMVasconcelos could provide links to relevant PRs.

> I think the context for this WG is to establish that we all agree we need that single page of instructions, and that the maintainers of the manifests are accountable of keeping this page up-to-date after they test that everything works, so that users can actually use the page to install their preferred set of KF apps. This is a significant problem for users right now trying to install.

I'm not on the same page here. Having spent two months already debating scope in #402; I'm very reluctant to debate docs and  expand beyond what was in #402. If we are still far apart on scope, then I'd probably recommend we push pause and revisit in 3-6 months. 

What's the downside to starting with reduced scope now and expanding later on? I would like to see the WG deliver on the scope as defined in #402 (i.e. a catalog in sync with upstream manifests) before considering further expansion.

cvenets:
> I'm not on the same page here. Having spent two months already debating scope in #402; I'm very reluctant to debate docs and expand beyond what was in #402. If we are still far apart on scope, then I'd probably recommend we push pause and revisit in 3-6 months.

> What's the downside to starting with reduced scope now and expanding later on? I would like to see the WG deliver on the scope as defined in #402 (i.e. a catalog in sync with upstream manifests) before considering further expansion.

I have to say that I don't really get why a single documentation page becomes such a big blocker and you are reluctant about it. Or what is the downside for having it included. I'm OK with expanding the scope later on as well.

I don't see a reason why anyone in the community would want to stall for 3-6 months. Thus, I'm fine going forward and removing this for now for the WG to be finalized, and re-discussing it on the website side as well.

wg-manifests/charter.md
cvenets: @PatrickXYS yes, this should be automated.

wg-manifests/charter.md
cvenets: I think we will all have to decide this together as a WG, when we pin down how to test.

wg-manifests/charter.md
yanniszark: I agree that ideally we would want all these overlays to be patched on the upstream apps, but we need to have a way to be able to push them to the manifests repo, even in another directory, so we can actually test. And I agree that we should then try to then push upstream or downstream to have the ideal scenario, where this directory is empty.

wg-automl/charter.md
yanniszark: > Coordinating with Training WG to make sure that all distributed training jobs can be used in AutoML experiments.

This line is not related to this PR, unless I am missing something.

wg-manifests/charter.md
```diff
@@ -0,0 +1,100 @@
+# WG Manifests Charter
+
+This charter adheres to the conventions, roles and organization management
+outlined in [wg-governance].
+
+## Scope
+
+- Provide a catalog (centralized repository) of Kubeflow application manifests.
```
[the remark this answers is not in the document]
yanniszark: We won't maintain the current manifests. We will structure new manifests based on the principles laid in this PR. The same manifests repo will be used.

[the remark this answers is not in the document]
cvenets: @Jeffwan if you are asking this due to kfctl, I think there are two options we can explore together:
1. You either keep/copy whatever current manifest you need that is kfctl-specific in the kfctl repo
2. You create a new overlay in the kfctl repo, which is kfctl-specific on top of the new manifests of this repo. This will make things much cleaner and maintainable even for kfctl.

wg-manifests/charter.md
yanniszark: You mean, for example, if KFP had a Google specific overlay in their manifests, if it would be present in the manifests repo because we copy their manifests? I think the answer to that is yes, since we'll most likely copy the whole manifests subtree. Does that make sense?

wg-manifests/charter.md
yanniszark: Will reword

wg-manifests/charter.md
yanniszark: Will reword

jlewi:
| I have to say that I don't really get why a single documentation page becomes such a big blocker and you are reluctant about it.

Lets agree to disagree.

LIke I said not having it in the charter doesn't prevent you from contributing to docs.

wg-manifests/charter.md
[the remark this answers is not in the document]
jlewi: > We won't maintain the current manifests. We will structure new manifests based on the principles laid in this PR. The same manifests repo will be used.

@yanniszark who is maintaining the manifests you or the application owners? My expectation is that this repo is just a mirror of whatever is upstream e.g.

/pipelines = https://github.com/kubeflow/pipelines/tree/master/manifests
/kfserving = https://github.com/kubeflow/kfserving/tree/master/config
/katib = https://github.com/kubeflow/katib

It looks like for the other applications they may not have upstream manifests yet.

So my expectation is that this WG is

1. Building automation to copy over the manifests
1. Coordinating with app WGs to create their upstream manifests

Does that match your understanding?

wg-manifests/charter.md
jlewi: Yes. More generally the question is if someone wants to create a customization to support X; e.g. (external SQL DB, Selecting KFP engine etc...) where should this go. Per [#402](https://github.com/kubeflow/community/pull/402#issuecomment-682704600) I think the options are

1. Push them upstream - they should coordinate with app owners to include it in their manifests
1. Push them downstream - They should work with distribution owners to figure out template free ways of post-processing the manifests

wg-manifests/charter.md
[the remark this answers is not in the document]
yanniszark: Yes it does. I will soon update the charter to reflect that.

wg-manifests/charter.md
yanniszark: @PatrickXYS I agree, but there is no PM WG right now and I don't see something similar in other WGs.
We can always add this in the near future!

yanniszark:
@jlewi I updated the PR according to your suggestions. Please take another look whenever possible.

thesuperzapper:
I made a comment https://github.com/kubeflow/community/pull/434#issuecomment-719106893, which I think is important to discuss. 

Under the definition I propose for "reference" distribution, I think this Working Group is the clear owner.

I define "reference" as the set of YAML which comprises Kubeflow itself (not how you deploy that YAML, or what Applications of Kubeflow you pick).

__This only leaves a few things:__
1. This WG should have authority to __prescribe what format__ Applications use for their YAML
1. This WG should have authority to __prescribe limitations on dependancies__ for Applications [For example, not depending on `Istio 1.5` features]
1. This WG should be tasked with __providing a single "reference" of cluster-level resources__ [For example, `istio`, `knative`, `cert-manager`, `dex`]
    - NOTE_1: this means that Kubeflow Applications will NOT include these cluster-level resources in thier YAML
    - NOTE_2: vendors can still swap in their own cluster-level resources (e.g. use a proprietary auth system)

jlewi:
@thesuperzapper 

> Under the definition I propose for "reference" distribution, I think this Working Group is the clear owner.

This is explicitly not in scope for this WG. And will be explicitly out of scope.

thesuperzapper:
@jlewi can you be precise about why those 3 points should not be responsibilities of this WG.

Or do you believe no one should be responsible for those things?

PatrickXYS:
I think 2 months of discussion about WG responsibilities has been a burden for all the community folks.

Before we come up with a **PERFECT** plan on how do we organize WG deployment and manifest, whether WG responsibilities overlap with another WG or not, 

Let's start from the smaller scope and start development work just now; the responsibilities can be expanded if WG has proved they have enough capacity and resources in making WG success.

jlewi:
@yanniszark and @cvenets thanks for agreeing to the changes. 

LGTM.  @paveldournov @theadactyl @Bobgy PTAL.

@yanniszark and @cvenets are you ok letting some of the existing directories in kubeflow/manifests remain that fall outside the scope? e.g. in particular
https://github.com/kubeflow/manifests/tree/master/aws
https://github.com/kubeflow/manifests/tree/master/gcp

I wouldn't expect the WG to maintain them (they all have OWNERs) files; But I don't want to unceremoniously kick them out. We can figure out an appropriate transition plan over time.

jlewi:
> Before we come up with a PERFECT plan on how do we organize WG deployment and manifest, whether WG responsibilities overlap with another WG or not,

+1 

Scope can always be expanded later on.

I'm intentionally pushing WGs to have very narrow and well defined scope because I want the WGs to be successful. I want to set very clear expectations and give the WGs easy wins. 

[Per my comment with @yanniszark](https://github.com/kubeflow/community/pull/435/files#r514632495)

> My expectation is that this repo is just a mirror of whatever is upstream e.g.

>/pipelines = https://github.com/kubeflow/pipelines/tree/master/manifests
>/kfserving = https://github.com/kubeflow/kfserving/tree/master/config
>/katib = https://github.com/kubeflow/katib

> It looks like for the other applications they may not have upstream manifests yet.

> So my expectation is that this WG is

>Building automation to copy over the manifests
>Coordinating with app WGs to create their upstream manifests

I hope this is unambiguous, uncontroversial and an easy win. In 6-12 months when we want to evaluate the WG and see if its meeting expectations it should be very easy. We just open two browsers and compare the upstream and manifests repo. 

This will demonstrate that the application WGs are up and running and coordinating effectively with the manifests WG. Once that's the case we will be in a much better position to tackle harder, more ambiguous problems.

See my [earlier comment](https://github.com/kubeflow/community/pull/435#issuecomment-717976058) , right now it looks to me like there is still confusion about the division of responsibilities. We need to fix this before tackling more complicated problems.

yanniszark:
> @yanniszark and @cvenets are you ok letting some of the existing directories in kubeflow/manifests remain that fall outside the scope? e.g. in particular
> https://github.com/kubeflow/manifests/tree/master/aws
> https://github.com/kubeflow/manifests/tree/master/gcp
> 
> I wouldn't expect the WG to maintain them (they all have OWNERs) files; But I don't want to unceremoniously kick them out. We can figure out an appropriate transition plan over time.


@jlewi absolutely, supporting and transitioning current users is important for the project.
I was thinking that we should perhaps place everything that should be phased out of `kubeflow/manifests` under a certain folder.
This structure will ensure the repo remains organized and clean, while allowing distribution owners to transition their users to their new distribution repos. The eventual goal is for that folder to become empty and be deleted.

What do you think?

jlewi:
Thanks

> I was thinking that we should perhaps place everything that should be phased out of kubeflow/manifests

Moving folders will likely break references. Lets discuss on a separate issue any migration/refactoring plans.

yanniszark:
@jlewi sounds good, let's discuss after merging this.
I think we have reached agreement on all points and the PR is reflecting that.
Is there anything missing to merge this PR?

theadactyl:
Thanks for ensuring that this effort has WG leadership and for taking the time to discuss scope.

/lgtm 
/approve

PatrickXYS:
/cc @yanniszark 

Can you rebase to pass the conflict issue?

yanniszark:
@theadactyl @jlewi @PatrickXYS rebased ontop of latest changes!

PatrickXYS:
Hi @yanniszark , do we have wg-manifests slack channel established?

PR #436, opened 2020-10-27
wg-notebooks: Update scope

yanniszark:
As discussed in today's community meeting and after lengthy discussions in:
- https://groups.google.com/g/kubeflow-discuss/c/d6whgEgror8
- https://github.com/kubeflow/community/pull/402
- https://docs.google.com/document/d/1BpR4RTp4m2GO_r-c3_xZ-qGnCfvQdqiRh4V0cFUSBqA

Update the scope of wg-notebooks to also include:
- Central Dashboard
- PodDefaults Controller

In addition, include KFAM and Profile Controller in maintenance mode.
Finally, regenerate with `make WHAT=wg-notebooks`.

/cc @jlewi @kimwnasptd @StefanoFioravanzo @cvenets


jlewi:
Thanks!

Really appreciate Arrikto stepping up here and taking on these responsibilities!

/lgtm
/hold
To give the other PWG team members a chance to review
/cc @kubeflow/project-steering-group 

Lets give the PSG and other folks until the end of the week. If there is no response by then we will consider this approved by lazy consensus.

yanniszark:
@jlewi seems like lazy consensus has been reached.
I believe we should go forward and finalize this, what do you think?

PR #453, opened 2020-11-12
Fix calendar for Feature Store SIG

woop:
Ensures that we use every 4th week instead of every 5th.


woop:
@jtfogarty 
@karlschriek 
@nicholas-abad 
@yuzliu 

Anyone can apply this change please?

PR #458, opened 2020-11-24
Update OWNERS, remove karlschriek

karlschriek:
Removing me from the OWNERS file. It has become patently clear that I will be a bottleneck here. At the moment I very rarely manage to react on time, if at all. To avoid frustration from the rest of the community it is best if I step out again.


Bobgy:
No worries! Everyone can have their own priorities. Thank you for making your position clear!

/lgtm
/approve

PR #470, opened 2021-01-21
Add jinchihe to wg_notebooks and fairing to subprojects of  wg_notebooks

jinchihe:
Add jinchihe to wg_notebooks to maintian Kubeflow Fairing. Thanks.


aronchick:
+1

StefanoFioravanzo:
/hold
Holding until at least a majority of the WG Notebooks chairs ACK this

Bobgy:
I think it's good to start the discussion now, but as previously discussed, we'd prefer WGs to start owning their own infra and release process. We should only consider expansion of WG responsibility after it has proven its ability to run independently on its own and successfully move towards the goal in its charter.

So we won't accept WG responsibility expansion right now.

Therefore, we can allow some time (TBD) to let Fairing find WG sponsorship before any other WG can take it.

cvenets:
@jinchihe, just to add to @Bobgy's comment, when we originally created the charter for this WG, we had asked about the inclusion of Fairing and Kale into it.

At that point, Google pointed out clearly that both should be out of scope of the Notebooks WG. Kale should continue to live outside of the `kubeflow` repo and Fairing should either become obsolete if unmaintained, or moved out of the repo/WG, if someone really takes responsibility of it.

jinchihe:
I think it's good choice to maintian in Notebooks WG since Fairing is great Python SDK for ML on k8s cluster. That's very useful for ML user in Notebooks. And I also can take some time to maintian that. Thanks @cvenets @Bobgy

PatrickXYS:
In addition, I think @jinchihe made good contributions to Fairing Development and Release.

But the only concern is we're not sure about what role does Fairing plays in notebook-wg.

thesuperzapper:
Currently the @kubeflow/wg-notebooks-leads agree that including faring is __not allowed__ under the [charter of wg-notebooks](https://github.com/kubeflow/community/blob/master/wg-notebooks/charter.md).

So I will close this PR, however I highly encourage @jinchihe to join the Kubeflow Community meetings or raise a proposal (an issue on this repo) for a new WG to own faring.

/close

PR #483, opened 2021-02-18
Adding a Kubeflow presentation template in pptx and odp

castrojo:
Konstantinos Palaiologos was kind enough to generate a nice presentation template for Kubeflow for people who are doing Kubeflow presentations. I thought it'd be cool to host these in a place where they can be shared/improved on by everyone. 

Reviewers, let me know if you think they should live someplace else if you'd prefer to not have a new directory in the root. Thanks!


theadactyl:
How can I best view the template deck? I believe we have an existing template in the community drive, and am wary of producing an "official" Kubeflow template.

thesuperzapper:
No comment on the template itself, but if you make a new folder, please create an OWNERS file

castrojo:
@theadactyl I wasn't aware there was a google drive, any way I can check it out? The pptx and odp can open in any presentation software, and can also be imported into google docs.

theadactyl:
All kubeflow-discuss members have access to the shared drive.

theadactyl:
View access that is.

Issue #487, opened 2021-02-24
Improve Twitter contribution workflow


theadactyl:
*Tracking ideas & feedback here in preparation for a formal proposal*

I've been mulling the best way to unblock contributions to the Twitter channel while following current Google guidelines around social channels. I think it would be best to require at least 2 LGTMs from a representative set of community reviewers and an approval from at least 1 kubeflow/community OWNER for publishing to @kubeflow. 

A past pain point for managing our blog & social workflows has been working with social management apps (Medium/Hootsuite) that are hard to keep up to date with current project roles and can be out of sight, out of mind. 

Given that GH is very much our source of truth right now, I'd love to implement a GitHub-based workflow for publishing to Twitter like what the Kubernetes community does [here](https://github.com/kubernetes-sigs/contributor-tweets/tree/master/.github/workflows). We could have a lighter-weight approval process for a dedicated contributor twitter account and/or one as I described above for the main kubeflow account. 

A bonus would be creating Twitter guidelines (what do we tweet about directly, what do we retweet, what is our basic style-guide) for contributors. 

@animeshsingh @castrojo since you've both mentioned this pain point, what do you think of the above?

castrojo:
I've used that tool to [contribute a tweet](https://github.com/kubernetes-sigs/contributor-tweets/blob/master/tweets/08-31-2020-slack-channel-tip.tweet) and while it is awkward compared to direct access, it beat the alternative of not being able to do it at all. 

Also even if it clunky it's transparent and every tweet has a record of submission and review. I can take a stab at doing a draft set of twitter guidelines, want me to draft something up for review?

theadactyl:
That would be fantastic -- whichever solution we use, that would be helpful
to have.

Also interested in hearing from @bogby about how challenging that
automation would be to implement.

---------- Forwarded message ---------

castrojo:
The automation seems to want it's own repository, so PRing a draft in here is probably putting the cart before the horse, however I have taken the K8s instructions, modiified them and mashed them together here: https://gist.github.com/castrojo/c02dcda1f3a630e524edcd5138a3b9aa

The automation and submissions instructions are straightforward and can be changed/adapted, the meat will be the policies themselves, which I'll inline here for discussion/feedback.:


# Kubeflow Twitter Guidelines

[...]

PR #492, opened 2021-02-26
[WIP] Initial blog guidelines

castrojo:
Here's a first stab at blog guidelines, please do not consider merging at this time as this is draft and needs discussion. It is an amalgamation of the K8s and Knative blog guidelines. 

Notes to our future selves: 
- The [twitter guidelines](https://github.com/kubeflow/community/issues/487) are also in flight and this will need to sync with that. 
- The website repo has it's own technical instructions for submitting posts so I linked to those instead of replicating them here.
- As the community guidelines get concensus, we might want to consider combining the blog and twitter guidelines into one document that is guidelines content only, and then refer the technical instructions to whichever place those end up being. This would avoid duplication of rules across multiple documents. Howerver one document per property isn't a bad place to start.

### DIFF
--- a/guidelines/blog-guidelines.md
+++ b/guidelines/blog-guidelines.md
@@ -1,6 +1,6 @@
 # Kubeflow Blog Guidelines
 
-The Knative blog is owned by $group and run by the $other_group. 
+The Kubeflow blog is owned by $group and run by the $other_group. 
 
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@castrojo did you make any progress on this? Or should we close it?

theadactyl:
Opening this back up -- we should make some updates and add to kubeflow/blog as part of the review process. cc @james-jwu @zijianjoy

PR #498, opened 2021-03-12
Add Kubeflow versioning proposal

knkski:
Adds a proposal for a strict release schedule for Kubeflow versions


proposals/kubeflow-versioning.md
```diff
@@ -0,0 +1,54 @@
...
+
+In brief, this proposal would switch the release cycle to this:
+
+![Kubeflow Versioning](diagrams/kubeflow-versioning.png)
+
+The changes this would introduce are:
+
+- A new version of Kubeflow would be released every 2 months
+- The deadline would be strict; if something is not ready for a particular release, it will be bumped to the next one
+- Each version would be supported (i.e. bugs/security issues fixed) for 4 months, except for for yearly Extended Support (ES) versions that would be supported for 16 months
```
RFMVasconcelos, line 17: `for` repeated

RFMVasconcelos: I am not sure the community has a concept of `supported` well defined yet but I think it makes sense to add a flag of `supportable`, meaning that organizations that offer Kubeflow support can have that as a reference.

proposals/kubeflow-versioning.md
```diff
@@ -0,0 +1,54 @@
...
+
+- The release cycle of individual Kubeflow components
+
+## Motivation
+
+- Strict release schedule
+  - It is not easy to plan around the current release schedule, as it does not have hard deadlines.
+- Bumping components to subsequent releases
+  - Kubeflow components must not block releases for this proposal to work, otherwise Kubeflow releases can't be strict.
+- 2 month release cycle
```
RFMVasconcelos, line 30: For this to work there might be the need to create a `rolling leadership scheme` or something that ensures there is accountable leadership and capacity in the community to do the work.

RFMVasconcelos (commented):
Thank you for this writing @knkski after the last community meeting! 

I left some comments. I think that this might be a good way to align component released with platform releases, and having a predictable release cadence would most certainly help organizations plan work around those releases. 

Let's see if we can find a path for the community to be able to deliver on this, if everyone agrees it's the right path.

chris-sanders:
This is a good start, I agree that something with a longer term bug-fix support will be necessary for Enterprise adoption. Anything that is required to be upgraded more than 1/year is, in my experience, in danger of getting itself eliminated from many environments. See Kubernetes itself, which is moving support to provide 1 year of fixes.

However, waiting for an annual release to adopt new features is also a large ask for a fast moving team. Having a fast iteration 'stable' and a long running ES release that won't get new features but does still get fixes for issues seems like a good mix to me.

I don't see anything about upgrades, just to be explicit, upgrades should be guaranteed for consecutive versions as well as from an ES to the next ES.

proposals/kubeflow-versioning.md
```diff
@@ -0,0 +1,54 @@
...
+- 4 month support cycle
+  - This length of time is enough that there are always 3 supported versions of Kubeflow, assuming a 2 month release cycle. This number of supported versions matches Kubernetes, where it has seemed to work well.
+- 16 month Extended Support cycle
+  - This length of time is long enough that distributors and consumers of Kubeflow that are not interested in a rapid release cycle or immediate new features are not forced to continuously upgrade Kubeflow versions.
+- YY.MM versioning scheme
+  - This scheme is proposed instead of semantic versioning due to the fact that each individual Kubeflow component may or may not have breaking changes in its release, and that information cannot be captured accurately in a single version number. It also helps enforce a strict release schedule, as releasing say 21.04 in May or June of 2021 is not very palatable.
```
zijianjoy, line 37: I am interested to know whether there is any alternatives: SEMVER is widely adopted by many products: https://semver.org/. The benefit of SEMVER is that MAJOR version bump is breaking change, while other version upgrade can be straightforward. It is preferable for big apps with many dependencies, it helps them to determine when and how to upgrade dependencies.

Having this YY.MM.N schema is basically replicating the information of release timestamp: this release time information can be found on kubeflow release note. And 4 months of support for any version is relatively short for company to act.

Taking the `Component Versioning Example` as example: Kubeflow release essentially is a collection of versioned components in a package. Alternative to YY.MM.N: We can upgrade the kubeflow SEMVER based on the Biggest change of components: If there is at least a component with MAJOR version bump, then we bump Kubeflow MAJOR version. Otherwise check whether there is MINOR version bump in components, if yes, then bump Kubeflow MINOR version.

## Exmaple

| Kubeflow Release |  A  |  B  |  C  |
| :--------------- | :-: | :-: | :-: |
| 1.1.0            | 1.0.0 | 0.5.0 | 1.3.0 |
| 1.2.0            | 1.1.0 | 0.6.0 | 1.4.0 |
| 1.2.1            | 1.1.0 | 0.6.1 | 1.4.1 |
| 2.0.0           | 2.0.0 | 0.6.1 | 1.5.0 |
| 2.1.0            | 2.0.0 | 0.7.0 | 1.5.1 |

Bobgy:
Thank you @knkski for the well thought proposal!
I love the spirit and I can certainly feel how it addresses some Kubeflow user problems.

However, I feel that an important missing piece is how practical it is that we can fulfill the proposal.

According to [Kubeflow distributions proposal](https://github.com/kubeflow/community/blob/master/proposals/kubeflow-distributions.md):

> ### What is a Kubeflow Distribution
> 
> A Kubeflow distribution is an opinionated bundle of Kubeflow applications optimized for a particular use case or environment.
> 
> ### Ownership & Development
> 
> Going forward new distributions of Kubeflow should be developed outside of the Kubeflow GitHub org. This ensures
> 
>   * Accountability for the distribution
>   * Insulates Kubeflow from the success or failure of the distribution
>   * Avoid further taxing Kubeflow's overstretched engprod resources(see[kubeflow/testing#737](https://github.com/kubeflow/testing/issues/737))
> 
> The owners of existing distributions should work with the respective WG/repository/org owners to come up with appropriate transition plans.

Therefore, Kubeflow doesn't decide its release cycle & support range by itself, but rather all (active) WGs and distribution owners agreed upon a certain release cycle and they try to get a release out at the same time. Also, each of them decide the support range by themselves.

Distribution owners develop and operate outside of Kubeflow org, so there's no way for us to enforce a release cadence and support range for them. Also, there might not be a one-size-fit-all answer to these numbers.

What's worse, I think there were some miscommunication in the past that, e.g. istio-dex distribution wasn't maintained for Kubeflow 1.2, we didn't know that and we only got many requests saying it doesn't work. So to move forward with the proposal, I think we need more discussions on:

* the promise we make is achievable for most distributions
* the efforts to achieve this promise are worth the opportunity cost that may be spent on developing Kubeflow
* an official mechanism to qualify distributions

I think getting all of the proposal in-place take some fairly large amount of efforts. Given that WG formation and letting each WG run on their own hasn't started long, and many distributions have just started to move off kubeflow/manifests (it's an explicit non-goal in KF 1.3 release). I feel that we have not reached a point to make these promises yet. (I totally agree with the spirit though, so when the temporary problems are resolved, I'd love to discuss more.)

Therefore, to summarize, I'm especially concerned about:

* 2 month release cycle (I don't feel all distributions and WGs are comfortable with the schedule, considering the amount of efforts each release needed)
* support cycles related proposals --> I believe distributions should be responsible for clarifying their own support cycle, we cannot decide for them.

The other proposals sound reasonable to me, but I'd say the release manager still has the power to override strict rules with reasonable arguments.

I wonder how other people, especially distribution owners think about this?

proposals/kubeflow-versioning.md
knkski: Thanks for the response. To be clear, the meat of this proposal is in the regularized release cadence; it would work just as well with semantic versioning.

That being said, I think the difficulty with semantic versioning here is that Kubeflow consists of enough components that there's likely to always be a breaking change in one of them. This would render semantic versioning semi-useless, since the answer to "is it a breaking change" would always be "yes". I'm also not sure how optional components would be handled. If an optional Kubeflow component included a breaking change, would that bump the major Kubeflow version?

> And 4 months of support for any version is relatively short for company to act.

I definitely agree, and that's why the proposal includes Extended Support releases that would allow companies to upgrade Kubeflow on a more relaxed, once a year schedule.

knkski:
@Bobgy: Thanks for the response. I guess to start, I'm interested in hearing more about this:

> Therefore, Kubeflow doesn't decide its release cycle & support range by itself

That was my impression before bringing this up in the community meeting on March 9th. However @yanniszark seemed to be saying that wasn't the case? Could we get some clarity on this?

> ... all (active) WGs and distribution owners agreed upon a certain release cycle and they try to get a release out at the same time. Also, each of them decide the support range by themselves.
>
> Distribution owners develop and operate outside of Kubeflow org, so there's no way for us to enforce a release cadence and support range for them. Also, there might not be a one-size-fit-all answer to these numbers.

To be clear, this proposal is not attempting to implement a release schedule that any individual component has to follow. Each Kubeflow component would still have its own release schedule. This proposal is more aimed at ensuring that there is a regular point at which all of the individual Kubeflow components can be guaranteed to integrate well together.

thesuperzapper:
@Bobgy @knkski while the release cadence may need to be changed from this proposal (2 months is quite fast), I really like some of the ideas, and think we should consider them:
1. `YY.MM` versioning:
     * Kubeflow is a collection of distinctly versioned applications, SEMVER principals cant sensibly applied to the project as a whole, see @knkski's [comment for a good description](https://github.com/kubeflow/community/pull/498/files#r595395792)
     * makes it obvious how old/new a version is for users
2. Decoupling of releases from distros:
     * There is no reason for Kubeflow itself to wait on every distro (and not all distros need to support every version)
     * Note, we need to define a testing phase of each release, which distros can provide feedback within
2. Defined support windows:
     * Enterprises typically don't update software quickly, so there is a strong need for extended support versions
     * I love the idea of effectively having "rapid" and "stable" channels 

Therefore, I think we should keep discussing something like this proposal.

proposals/kubeflow-versioning.md
zijianjoy: Sounds good on providing a more relaxed, once a year schedule for Extended Support releases. 👍 

About breaking changes and SEMVER: I think your argument of breaking changes in each Kubeflow release makes sense. However, I hope that not every Kubeflow release contains major component upgrade: That means we need to maintain a long list of upgrade guide across the cloud providers. For example: We need to maintain a document about `How do you upgrade from v1 to v2, v2 to v3, v3 to v4, etc.`, we don't hope that Kubeflow components bring in breaking changes as frequently as monthly. So my suggestion is to schedule a `breaking change Kubeflow release` in annual or semi-annual frequency, where Kubeflow components can sign up for their breaking changes release.

My opinion towards optional components: the breaking change of optional component will probably be taken into account too, as long as they are listed under kubeflow/manifest.

PR #512, opened 2021-04-23
Add DGL Operator proposal to kubeflow community

ryantd:
This is Xiaoyu Zhai, from Qihoo 360 AI Infra. Currently, our team is working on DGL Operator to make DGL distributed training easier on Kubernetes. And I am glad to introduce our proposal.

Looking forward to having you guys any feedback.


terrytangyuan (commented):
Great to see this! cc'ing the Kubeflow Training WG leads to review this.

/cc @zw0610 @carmark

proposals/dgl-operator-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+      template:
+        spec:
+          containers:
+          - image: dgloperator/graphsage:v0.1.0
+            name: dgl-graphsage
+```
+
+- The `cleanPodPolicy` can be optionally configured as `Always` / `Never` / `OnFailure` / `OnCompletion`, indicating whether to delete the pod when the task is terminated. Preferred `Never` in debugging and `OnCompletion` in production.
+
+- The content of `Launcher` and `Worker` are followed the `PodTemplateSpec`. Users can be free to add more native key-values according to the spec.
```
terrytangyuan, line 74: Looks like we can reuse the common spec in https://github.com/kubeflow/common

proposals/dgl-operator-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Container Image
+
+A requirement on user's container image is that it includes DGL and related libraries, and our supporting tools (`dglrun` and util scripts).
+
+SSH is not needed (or used).
+
+### Resulting Launcher
+
+The resulting launcher resembles ones in MPI Operator very much. Like, `kubectl-delivery` makes sure all the worker pods are ready and download the `kubectl` (we just did some minor changes on Env and ip config processing).
```
terrytangyuan, line 84: Perhaps we can consider moving the common parts from MPI Operator ([kubectl-delivery controller](https://github.com/kubeflow/mpi-operator/tree/master/pkg/controllers/kubectl_delivery)) to kubeflow/common so DGL operator can use it as well?

proposals/dgl-operator-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+3. Create worker `Pods` that reaches the desired replicas. The command is set to sleep forever.
+4. Create launcher `Pods`. And `initContainer` (`kubectl-delivery`) in `launcher` pod takes the responsibility to:
+    1. wait for all worker pods are ready
+    2. `kubectl` binary is delivered to an `emptyDir` volume
+5. Take off main container of the launcher `Pod`, and run the `dglrun` command and arguments.
+    1. `dglrun`: A must-have binary file that can take the control of DGL distributed training.
+    2. The launcher is very light weight.
+6. When `DGLJob` finishes, it will clean all workers.
+
+![DGL Operator-wise Diagram](diagrams/dgl-operator-1.png)
```
terrytangyuan, line 215: This is very similar to MPI operator's design. Any thoughts on how to refactoring and reuse what's in MPI operator?

proposals/dgl-operator-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Alternatives Considered
+
+### DGL Operator vs MPI Operator
+
+Currently, for distributed partitioning, DGL uses `ParMETIS` to split big graphs up and store in parallel on each worker `Pods`, which is empowered by MPI. For distributed training, DGL uses PyTorch's `DDP` in dense parameters updating, and self-developed methods in other scenarios.
+
+So in this manner, a fully-informed user, who understand any details about DGL and Operator philosophy, can hack all DGL workloads (distributed partitioning & training) on MPI Operator. However, this hack approach is not a good idea, because 1) user need to revise `ipconfig` processing in DGL library to adapt MPI Operator's `hostfile`; 2) user may need involve intrusive code in partition and train script.
+
+Therefore, DGL Operator will follow the community, to provide native DGL distributed training experience. There is no need for users to change any code, transferring the workload from machines to Kubernetes.
```
terrytangyuan, line 244: Good to see this has been taken into consideration. I understand that there can be specific differences that are framework-specific but how much of this could be common and reused?

proposals/dgl-operator-proposal.md
ryantd: Ongoing development is based on pure Kubebuilder v2, not engaged in Kubeflow common spec definition or job controller. But we are glad to reuse them.

proposals/dgl-operator-proposal.md
ryantd: Agreed, sounds great

proposals/dgl-operator-proposal.md
ryantd: Leave my thoughts at your last comment (https://github.com/kubeflow/community/pull/512#discussion_r619407527)

proposals/dgl-operator-proposal.md
ryantd: Thank Terry for reading this consideration, which is the most significant concern in my thinking. Here are some points that I want to discuss with you.

Firstly, let me add one more thing,

**Support 2 partitioning ways**: the operator-wise design part in this proposal is the most ideal implementation for supporting fully distributed partitioning and training. However, distributed partitioning is an experimental new-coming feature, and I think it is a work-around way of DGL recommended, which is not included in DGL library currently. Moreover, the standalone partitioning method is included in DGL library, so I want to support both distributed and standalone in "DGL Operator".

Then, I think if we (all allreduce-way operators) reused MPI Operator as much as possible, MPI Operator may need,

1. Imporve the customerization of the `hostfile` generating formats. DGL Operator's `ipconfig` format: [ip] [dgl server port] [worker num] [host/pod name] [gpu num]
2. Improve the customization of the path of config_volume
3. How to fulfill standalone partitioning without any training workloads' waiting?
   1. Idea 1: At first, use an initContainer to invoke a `Partitioner` Pod. And then take off all training worker Pod. So we may need to find a way of injecting this initContainer.
   2. Idea 2: Use job status to control the workflow of `Partitioner` / `Worker` Pod's starting phase and releasing phase

ryantd:
The proposal is updated
1. add standalone partition mode
2. illustrate the differences of implementation between standalone and distributed partitioning
3. fix typos

gaocegege:
/cc @carmark @zw0610 @Jeffwan 

The proposal LGTM, It's clean and elegant. One concern is that if we should maintain a new operator or support it in mpi-operator.

carmark:
Great proposal, @ryantd .

@gaocegege , I do prefer reuse the mpi-operator and do some extensions.

proposals/dgl-operator-proposal.md
```diff
@@ -0,0 +1,284 @@
...
+Currently, for the scope of this proposal, we won't be addressing the method for serving the model.
+
+## API
+
+### Custom Resource Definition
+
+The CRD yaml example is as following:
+
+```yaml
+apiVersion: qihoo.net/v1alpha1
```
Jeffwan, line 31: I assume apiVersion will be changed to kubeflow fashion?

proposals/dgl-operator-proposal.md
```diff
@@ -0,0 +1,284 @@
...
+          containers:
+          - image: dgloperator/graphsage:v0.1.0
+            name: dgl-graphsage
+```
+
+- The `cleanPodPolicy` can be optionally configured as `Always` / `Never` / `OnFailure` / `OnCompletion`, indicating whether to delete the pod when the task is terminated. Preferred `Never` in debugging and `OnCompletion` in production.
+
+- The content of `Launcher` and `Worker` are followed the `PodTemplateSpec`. Users can be free to add more native key-values according to the spec.
+
+- The `partitionMode` can be configured as `distributed` / `standalone`, indicating how to partition the graph. `standalone` uses DGL API `dgl.distributed.partition_graph`, while `distributed` uses DGL recommended work-around solution [ParMETIS](https://docs.dgl.ai/guide/distributed-preprocessing.html#distributed-partitioning).
```
Jeffwan, line 77: If user choose to use `distributed` but only provides 1 worker. Does it run under standalone eventually?

proposals/dgl-operator-proposal.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Alternatives Considered
+
+### DGL Operator vs MPI Operator
+
+Currently, for distributed partitioning, DGL uses `ParMETIS` to split big graphs up and store in parallel on each worker `Pods`, which is empowered by MPI. For distributed training, DGL uses PyTorch's `DDP` in dense parameters updating, and self-developed methods in other scenarios.
+
+So in this manner, a fully-informed user, who understand any details about DGL and Operator philosophy, can hack some DGL workloads (distributed partitioning & training) on MPI Operator. But this hack approach is not a good idea, because 1) user need to revise `ipconfig` processing in DGL library to adapt MPI Operator's `hostfile`; 2) user may need involve intrusive code in partition and train script.
+
+However, there is a way to reuse MPI Operator as much as possible, MPI Operator may need,
```
Jeffwan, line 279: @terrytangyuan @carmark

Can you help evaluate if these can be changed in MPI-Operator to support DGL use case?

Jeffwan:
/hold

Looks like there's way to support in MPI-Operator. If that's the case, I think the preferred option is to check possibility and efforts we need to make in mpi-operator to support this case.  @ryantd If you like to add more details on this, that would be great.

proposals/dgl-operator-proposal.md
ryantd: 🤔 , yes, but a `ParMETIS` standalone. The 'standalone' I used in the proposal context is meaning `DGL API` standalone.

Do you think "`partitionMode` can be configured as `DGL-API` / `ParMETIS`", would be better?

proposals/dgl-operator-proposal.md
ryantd: Yes, apiVersion will be changed to kubeflow fashion

ryantd:
> @ryantd If you like to add more details on this, that would be great.

@Jeffwan . Sure, and let me put more details on this.

## basic requirements
1. custom `hostfile` generating formats. DGL Operator's ipconfig format: [pod ip] [dgl server port] [dgl server num] [host/pod name] [gpu num]
2. custom `ContainerPort`
3. custom `Env`

## standalone partition (DGL-API) needs
1. a `Partitioner` Pod
    1. ideas:
        1. `Worker-0` will be looked as a `Partitioner`, will be terminated after finished the dispatching job
        2. `Worker-1` to `Worker-N` will work on the training later
        3. Use env to recognize which one is `Partitioner`, and `supporting tools` can do the actual partitioning things
2. a `partition-driver` initContainer, waiting for `Partitioner` Pod to finish the partitioning job
    1. To be solved: how to inject any custom initContainer logic into MPI Operator philosophy
image

So far, these are the questions I have thought of. I am glad that DGL Operator can merge into MPI Operator, but I don't want to make MPI Operator deal with more diverged issues and potentially bigger and bigger.

And personly, I feel like, if we **only** support the distributed partitioning (`ParMETIS` way), it is more easily merged into MPI Operator.

proposals/dgl-operator-proposal.md
ryantd: more details at https://github.com/kubeflow/community/pull/512#issuecomment-827414177

gaocegege:
@Bobgy Do we have a process to determine if we should accept external contributions?

Bobgy:
Yes, @theadactyl will soon send out a PR for process on how we decided for paddle operator.

And I will use paddle operator to evaluate the time needed for adding a repo.

Jeffwan:
@ryantd Do you have an idea about next steps? Do you think support this case in mph-operator is a reasonable direction to go? If not, any technical blockers?

ryantd:
@Jeffwan Sorry for the late reply

> Do you think support this case in mpi-operator is a reasonable direction to go?

After this period of consideration, I think the migration is not a reasonable direction.

Because,
1. I think DGL Operator should be looked as a GNN focused Operator. TF, PyTorch, and Horovod are intended to solve CNN/RNN problems. There are some GNN exclusive topics,
    1. **Partition**. Currently, you must have a Partition workload to do loading and partitioning.
    2. **DGL-Server, DGL-Sampler, and DGL-Trainer in Training**. Currently, these 3 parts are included in the training workload. But, in the near future, DGL-Server, DGL-Sampler, and DGL-Trainer may have their own workload (Pod).
2. If the migration happens, MPI Operator needs to support the Partitoiner workload. And we want the Partitioner workload to be created by Operator, rather than in users' YAML file. So I must put the Partitioner logic into MPI Operator, while I think it is the most unreasonable point.
3. Distributed DGL is built on the top of PyTorch's allreduce approach (DDP) to solve dense parameters updating, while sparse parameters updating is DGL self-developed.

So in this context, I think DGL Operator should be independent.

ryantd:
@Jeffwan Sorry for the late reply

> Do you think support this case in mpi-operator is a reasonable direction to go?

**Summary: After the period of consideration, I think I should provide 3 options to introduce my thinking, and let the community and WG determine.**

----

## Graph Partitioning

The most important thing is graph partitioning. DGL has 2 approaches to partition, one is single-thread (DGL built-in API), the other one is multi-process (ParMETIS, not in DGL).

In the K8s+Operator context, DGL Operator wants to apply these and support 3 `partitionMode`s,
- Single Pod, single-thread partitioning, _will call SSP later_
- Single Pod, multi-thread partitioning (native ParMETIS version), _will call SMP later_
- Multi Pod, single-thread partitioning (may need some adoptions in ParMETIS, still in research), _will call MSP later_

You can see SSP and SMP need a single Partitioner Pod. And MSP may don't need the Partitioner Pod, can partition in parallel in Worker Pods. So, **SSP and SMP cannot apply MPI manner**[1], **MSP can apply MPI manner**.

## Training

Each DGL distributed training workload has 3 components, DGL-Server, DGL-Sampler, and DGL-Trainer. DGL-Trainer will do the allreduce thing, but only use PyTorch's DDP to solve dense parameters updating, while sparse parameters updating is DGL self-developed. Currently, these 3 parts are included in the training workloads (Worker Pods, default one DGL-Server, one DGL-Sampler, and one DGL-Trainer). But, in the near future, DGL-Server, DGL-Sampler, and DGL-Trainer may have their own workload (Pod).

So, **if we will not upgrade to "training workloads support any number of sub-workloads", training can apply fully MPI manner**.

## Conclusion
In view of the above discussions, I can give 3 options
- Option 1: Entirely merging into MPI Operator
- Option 2: Let DGL training be a two-stage Operator mechanism (**90% support all user cases**)
- Option 3: Be an independent Operator (**Fully support all user cases**)

### Option 1
In order to entirely merge into MPI Operator, and apply fully MPI manner. We should only support **MSP and non-upgraded training workload**, and remove SSP, SMP, and upgraded training workload.

### Option 2
In order to 90% support all user cases, and maximized merging into MPI Operator. DGL Operator will be an upper layer of MPI Operator, only do SSP and SMP part. And MPI Operator will do the same things in Option 1.

Something like [Wang Zhang's idea](https://docs.google.com/document/d/1x1JPDQfDMIbnoQRftDH1IzGU0qvHGSU4W6Jl4rJLPhI/edit#heading=h.e33ufidnl8z6)

### Option 3
Be an independent Operator. DGL Operator may need to be looked at as a GNN focused Operator. TF, PyTorch, and Horovod are intended to solve CNN/RNN problems.

## Appendix

[1]: Put Partitioner logic manually into MPI Operator can make SSP and SMP apply MPI Operator manner. But I think this will break the specificity of MPI Operator.

**This comment is up-to-date, may be different from the proposal. And proposal will be updated ASAP.**

zw0610:
How thrilling to have a GNN training operator in Kubeflow commuting!

Regarding the three partition mode, the SSP mode is more like the [single-client mode](https://github.com/tensorflow/community/blob/master/rfcs/20200306-single-client-parameter-server.md) in TensorFlow PS/Worker training if we can merge `Launcher` and `Partitioner`. From my understanding, the `main` container in `Partitioner` can be converted into an `initContainer` in the `Launcher` pod in replace of `watcher-loop-partitioner` as long as the partitioner can quit before the workers start training. Please do correct me if I'm wrong.

The SMP mode, which also needs a single partitioner Pod, can follow the same fashion mentioned above. However, I'm concerned how the output of `pm_dglpart` shall be distributed and synchronized to worker pods. But with `kubectl` installed in the `Launcher` pod, the worst scenario is to use `kubectl cp`.

The MSP mode is a regular MPIJob case just like Xiaoyu mentioned.

If this suggestion is workable, option 1 should be feasible to all user cases. If I missed something somewhere, please ping me at any time.

terrytangyuan:
@zw0610 @ryantd Given that you have discussed this separately offline, could either of you provide a summary of your discussions here as well so that everyone will be on the same page?

zw0610:
Sure, we will keep the community updated after the scheduled discussion
this week.

proposals/dgl-operator-proposal.md
```diff
@@ -0,0 +1,384 @@
...
+The CRD yaml example is as following:
+
+```yaml
+apiVersion: qihoo.net/v1alpha1
+kind: DGLJob
+metadata:
+  name: dgl-graphsage
+  namespace: dgl-operator
+spec:
+  cleanPodPolicy: Running
```
cheimu, line 61: Hi Xiaoyu and comminuty : )  I don't know if it is probably better to use runPolicy field https://github.com/kubeflow/common/blob/master/pkg/apis/common/v1/types.go#L176 as other operators did instead of directly using cleanPodPolicy ?

For example tfjobs https://github.com/kubeflow/training-operator/blob/master/pkg/apis/tensorflow/v1/types.go#L55; 
pytorchjob https://github.com/kubeflow/training-operator/blob/master/pkg/apis/pytorch/v1/types.go#L53. 

For standalone operator, this works fine; but for all-in-one training-operator, all of controllers use the same base job controller, where the clean up logic try to read runPolicy.cleanPodPolicy.  Like mpi-operator migrating into training-operator, the fields mismatched will be an headache.

varodrig:
@ryantd  should we close this item since it's 2 years old?
cc @terrytangyuan

terrytangyuan:
Yes, feel free to submit a KEP to kubeflow/trainer if you are still interested. @ryantd

Issue #518, opened 2021-06-11
Restructure of Kubeflow Slack Channels


thesuperzapper:
Hey all, I believe the Slack would be much easier for new (and existing) users if we instituted a standard set of "prefixes" for the channel names. 

Prefixes are very standard practice on Slack and allow you to effectively "group" channels into sections.

__Channels Layout:__
* `#announcements`
   * CHANNEL - for community announcements
   * DEFAULT - yes
   * NOTE - must be the original `#general` (due to free slack limitations)
* `#general`
   * CHANNEL - for all general discussions.
   * DEFAULT - yes 
* `#feature-requests`
   * CHANNEL - for people to suggest/discuss new ideas for Kubeflow
   * DEFAULT - yes 
* `#job-postings`
   * CHANNEL - for people to post Kubeflow related jobs
   * DEFAULT - no 
* `#release`
   * CHANNEL - for people to discuss the next release of Kubeflow
   * DEFAULT - no 
* `#distro-*`
   * PREFIX - for Kubeflow distributions
* `#external-*`
   * PREFIX - for software/things NOT owned by a Kubeflow WG
* `#kubeflow-*`
   * PREFIX - for software/things owned by a Kubeflow WG
* `#platform-*`
    * PREFIX - for cloud/hosting platforms (not distributions)
* `#users-*`
    * PREFIX - for specific groups of users (e.g countries, regions)
* `#sig-*`
   * PREFIX -  for official Kubeflow Special Interest Groups
* `#wg-*`
   * PREFIX - for official Kubeflow Working Groups

johnugeorge:
Adding Kubeflow prefix looks redundant to me. 
Also, it is confusing to have Kubeflow-operator-tensorflow mapped to tf-operator, Is there real need for this change at this point?

thesuperzapper:
@johnugeorge thanks for your comments!

### Re: the `kubeflow-*` prefix

Adding this prefix has a few main benefits:
* Allows us to distinguish things are "officially part of Kubeflow"  (that is, maintained" by a Kubeflow WG) 
* It groups all the component discussions together on the side menu
* `#kubeflow-pipelines` has stronger brand recognition than `#pipelines` (as does `#kubeflow-notebooks` vs `#notebooks`)

### Re: the `kubeflow-operator-*` prefix

There are many training operators, so if they have different prefixes, they appear all over the side-menu, making it hard for users to understand how they relate to each-other and Kubeflow.

[...]

PR #521, opened 2021-08-09
remove archived channels and fix channels' names

haoxins:


terrytangyuan (approved):

thesuperzapper:
@haoxins there are actually quite a few more channel renames than you have reflected here, if your game, do you want to update your PR to reflect them all?

Also, note that lots of channels were archived (so would need to be removed from this list).

__Here is the issue discussing the slack renames:__
- https://github.com/kubeflow/community/issues/518

Bobgy:
/lgtm
/Approve

Thank you for the contribution!

PR #522, opened 2021-08-16
feat(pytorch): Add elastic proposal

gaocegege:
Signed-off-by: cegao <cegao@tencent.com>

### DIFF
--- a/proposals/pytorch-elastic-proposal.md
+++ b/proposals/pytorch-elastic-proposal.md
@@ -141,7 +141,36 @@ One built-in port named `kubeflow-rdzv-port` is introduced for `rendezvous`.
 
 ### Reconciliation
 
+`JobController.ReconcilePods` should be refactored. Now the pods are returned by `GetPodSlices`. For example, if `spec.Replicas` is 3, the PodSlices may look like: `[[0],[1],[2]]`. It is not expected when elastic training is enabled.
 
+```go
+// ReconcilePods checks and updates pods for each given ReplicaSpec.
+// It will requeue the job in case of an error while creating/deleting pods.
+func (jc *JobController) ReconcilePods(
+	job interface{},
+	jobStatus *apiv1.JobStatus,
+	pods []*v1.Pod,
+	rtype apiv1.ReplicaType,
+	spec *apiv1.ReplicaSpec,
+	replicas map[apiv1.ReplicaType]*apiv1.ReplicaSpec) error {
+  ...
+	numReplicas := int(*spec.Replicas)
+	var masterRole bool
+  ...
+	podSlices := jc.GetPodSlices(pods, numReplicas, logger)
+	for index, podSlice := range podSlices {
+		if len(podSlice) > 1 {
+			logger.Warningf("We have too many pods for %s %d", rt, index)
+		} else if len(podSlice) == 0 {
+			logger.Infof("Need to create new pod: %s-%d", rt, index)
+      ...
+		} else {
+			...
+		}
+	}
+	return nil
+}
+```
 
 ### Resulting Spec
 
@@ -179,3 +208,87 @@ spec:
 - KUBEFLOW_RDZV_PORT will be open for every pod even though workers except worker-0 do not use it.
 
 ## Alternatives Considered
+
+### API/CRD
+
+[TorchElastic operator](https://github.com/pytorch/elastic/blob/master/kubernetes/api/v1alpha1/elasticjob_types.go) implemented by @jeffwan puts the new fields under `PyTorchJobSpec`.
+
+Personally, prefer keeping it in `common.ReplicaSpec` since other Jobs may also need it.
+
+```diff
+// PyTorchJobSpec is a desired state description of the PyTorchJob.
+type PyTorchJobSpec struct {
+	// RunPolicy encapsulates various runtime policies of the distributed training
+	// job, for example how to clean up resources and how long the job can stay
+	// active.
+	//+kubebuilder:validation:Optional
+	RunPolicy common.RunPolicy `json:"runPolicy"`
+
++	// minReplicas is the lower limit for the number of replicas to which the training job
++	// can scale down.  It defaults to nil.
++	// +optional
++	MinReplicas *int32 `json:"minReplicas,omitempty"`
++	// upper limit for the number of pods that can be set by the autoscaler; cannot be smaller than MinReplicas, defaults to nil.
++	// +optional
++	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// A map of PyTorchReplicaType (type) to ReplicaSpec (value). Specifies the PyTorch cluster configuration.
+	// For example,
+	//   {
+	//     "Master": PyTorchReplicaSpec,
+	//     "Worker": PyTorchReplicaSpec,
+	//   }
+	PyTorchReplicaSpecs map[common.ReplicaType]*common.ReplicaSpec `json:"pytorchReplicaSpecs"`
+}
+
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen=true
+// ReplicaSpec is a description of the replica
+type ReplicaSpec struct {
+	// Replicas is the desired number of replicas of the given template.
+	// If unspecified, defaults to 1.
+	Replicas *int32 `json:"replicas,omitempty"`
+
+	// Template is the object that describes the pod that
+	// will be created for this replica. RestartPolicy in PodTemplateSpec
+	// will be overide by RestartPolicy in ReplicaSpec
+	Template v1.PodTemplateSpec `json:"template,omitempty"`
+
+	// Restart policy for all replicas within the job.
+	// One of Always, OnFailure, Never and ExitCode.
+	// Default to Never.
+	RestartPolicy RestartPolicy `json:"restartPolicy,omitempty"`
+}
+```
+
+### Autoscaler Integration
+
+Three fields should be added in CustomResourceDefinition:
+
+```yaml
+    scale:
+      specReplicasPath: .spec.pytorchReplicaSpecs.Worker.replicas
+      # Should we have a total replicas?
+      statusReplicasPath: .status.replicaStatuses.Active
+      labelSelectorPath: .status.labelSelector
+```
+
+`LabelSelector` should be introduced into `common.ReplicaStatus`.
+
+```diff
+type ReplicaStatus struct {
++	// LabelSelector is the selector for the replica.
++	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
+
+	// The number of actively running pods.
+	Active int32 `json:"active,omitempty"`
+
+	// The number of pods which reached phase Succeeded.
+	Succeeded int32 `json:"succeeded,omitempty"`
+	// The number of pods which reached phase Failed.
+	Failed int32 `json:"failed,omitempty"`
+}
+```
+
+Then `PyTorchJob` has the `scale` subResource, then it can work with Autoscaler. The only problem is that, `PyTorchJob` already has the minReplicas and maxReplicas fields. They are used to generate command. The Autoscaler resource needs them, too. Thus users may need to define them again.
+ 
--- a/proposals/pytorch-elastic-proposal.md
+++ b/proposals/pytorch-elastic-proposal.md
@@ -1,0 +1,144 @@
+_Authors:_
+
+* [Ce Gao(@gaocegege)](https://github.com/gaocegege)
+
+_Status_
+
+- 2021-08-16 Draft v1
+
+## Abstract
+
+[TorchElastic](https://pytorch.org/docs/1.9.0/distributed.elastic.html), which was open sourced over a year ago in the pytorch/elastic github repository, is a runner and coordinator for PyTorch worker processes. it has been part of PyTorch core since 1.9.0. This proposal is to support such feature with the help of PyTorchJob. 
+
+## Background
+
+### TorchElastic Design
+
+TorchElastic provides the approach that makes distributed PyTorch fault-tolerant and elastic.
+
+To launch a fault-tolerant job, run the following on all nodes:
+
+```python
+python -m torch.distributed.run
+        --nnodes=NUM_NODES
+        --nproc_per_node=TRAINERS_PER_NODE
+        --rdzv_id=JOB_ID
+        --rdzv_backend=c10d
+        --rdzv_endpoint=HOST_NODE_ADDR
+        YOUR_TRAINING_SCRIPT.py (--arg1 ... train script args...)
+```
+
+To launch an elastic job, run the following on at least MIN_SIZE nodes and at most MAX_SIZE nodes.
+
+```python
+python -m torch.distributed.run
+        --nnodes=MIN_SIZE:MAX_SIZE
+        --nproc_per_node=TRAINERS_PER_NODE
+        --rdzv_id=JOB_ID
+        --rdzv_backend=c10d
+        --rdzv_endpoint=HOST_NODE_ADDR
+        YOUR_TRAINING_SCRIPT.py (--arg1 ... train script args...)
+```
+
+The command needs to be run on every node. The training job will start When at least min total number of nodes have joined. The command runs a local elastic agent on the node, which is used to launch and manage underlying worker processes. The agent on the node assigns `RANK`, `LOCAL_RANK` and so on for every worker.
+
+## Goals
+
+A Kubeflow user should be able to run elastic training using PyTorch. This proposal is centered around a Kubernetes operator for PyTorch. A user should be able to run both single node and distributed elastic training jobs with PyTorch.
+
+Besides this, users should be able to run non-elastic training jobs as before.
+
+## Design and Implementation
+
+We introduce the design and implentation based on all-in-one operator.
+
+## API/CRD
+
+```diff
+// PyTorchJobSpec is a desired state description of the PyTorchJob.
+type PyTorchJobSpec struct {
+	// RunPolicy encapsulates various runtime policies of the distributed training
+	// job, for example how to clean up resources and how long the job can stay
+	// active.
+	//+kubebuilder:validation:Optional
+	RunPolicy common.RunPolicy `json:"runPolicy"`
+
+	// A map of PyTorchReplicaType (type) to ReplicaSpec (value). Specifies the PyTorch cluster configuration.
+	// For example,
+	//   {
+	//     "Master": PyTorchReplicaSpec,
+	//     "Worker": PyTorchReplicaSpec,
+	//   }
+	PyTorchReplicaSpecs map[common.ReplicaType]*common.ReplicaSpec `json:"pytorchReplicaSpecs"`
+}
+
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen=true
+// ReplicaSpec is a description of the replica
+type ReplicaSpec struct {
+	// Replicas is the desired number of replicas of the given template.
+	// If unspecified, defaults to 1.
++	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
+
++	// minReplicas is the lower limit for the number of replicas to which the training job
++	// can scale down.  It defaults to nil.
++	// +optional
++	MinReplicas *int32 `json:"minReplicas,omitempty"`
++	// upper limit for the number of pods that can be set by the autoscaler; cannot be smaller than MinReplicas, defaults to nil.
++	// +optional
++	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// Template is the object that describes the pod that
+	// will be created for this replica. RestartPolicy in PodTemplateSpec
+	// will be overide by RestartPolicy in ReplicaSpec
+	Template v1.PodTemplateSpec `json:"template,omitempty"`
+
+	// Restart policy for all replicas within the job.
+	// One of Always, OnFailure, Never and ExitCode.
+	// Default to Never.
+	RestartPolicy RestartPolicy `json:"restartPolicy,omitempty"`
+}
+```
+
+Two fields are added in `common.ReplicaSpec`: `minReplicas` and `maxReplicas`. They acts as MIN_SIZE and MAX_SIZE in the elastic example above.
+
+## Command
+
+```yaml
+apiVersion: "kubeflow.org/v1"
+kind: "PyTorchJob"
+metadata:
+  name: "pytorch-dist-mnist"
+spec:
+  pytorchReplicaSpecs:
+    # There is no master in elastic training jobs.
+    Worker:
+      minReplicas: 3
+      MaxReplicas: 5
+      restartPolicy: OnFailure 
+      template:
+        spec:
+          containers: 
+            - name: pytorch
+              image: <image>
+              command: "python -m torch.distributed.run --rdzv_backend=c10d --rdzv_endpoint=$KUBEFLOW_RDZV_HOST:$KUBEFLOW_RDZV_PORT --nnodes=$KUBEFLOW_MIN_SIZE:$KUBEFLOW_MAX_SIZE --nproc_per_node=1 xxx.py"
+```
+
+There are four environment variables here: `KUBEFLOW_RDZV_HOST`, `KUBEFLOW_RDZV_PORT`, `KUBEFLOW_MIN_SIZE` and `KUBEFLOW_MAX_SIZE`.
+
+## Alternatives Considered
+One alternative considered for the CRD spec is shown below:
+```yaml
+apiVersion: "kubeflow.org/v1alpha1"
+kind: "PyTorchJob"
+metadata:
+  name: "example-job"
+spec:
+  backend: "gloo"
+  masterPort: "23456"
+  worldSize: 3
+  container:
+  - image: pytorch/pytorch:latest
+```
+The idea was the number of replicas for worker and masters could be derived from the `worldSize` given there would only be one master. It was decided against based on the fact that it deviates from a regular replicaSpec and provides less customization.
--- a/proposals/pytorch-elastic-proposal.md
+++ b/proposals/pytorch-elastic-proposal.md
@@ -116,7 +116,7 @@ spec:
     # There is no master in elastic training jobs.
     Worker:
       minReplicas: 3
-      MaxReplicas: 5
+      maxReplicas: 5
       restartPolicy: OnFailure 
       template:
         spec:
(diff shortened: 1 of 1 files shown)


proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+python -m torch.distributed.run
+        --nnodes=MIN_SIZE:MAX_SIZE
+        --nproc_per_node=TRAINERS_PER_NODE
+        --rdzv_id=JOB_ID
+        --rdzv_backend=c10d
+        --rdzv_endpoint=HOST_NODE_ADDR
+        YOUR_TRAINING_SCRIPT.py (--arg1 ... train script args...)
+```
+
+The command needs to be run on every node. The training job will start When at least min total number of nodes have joined. The command runs a local elastic agent on the node, which is used to launch and manage underlying worker processes. The agent on the node assigns `RANK`, `LOCAL_RANK` and so on for every worker.
```
alculquicondor, line 43: by node you mean worker pod?

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+	Template v1.PodTemplateSpec `json:"template,omitempty"`
+
+	// Restart policy for all replicas within the job.
+	// One of Always, OnFailure, Never and ExitCode.
+	// Default to Never.
+	RestartPolicy RestartPolicy `json:"restartPolicy,omitempty"`
+}
+```
+
+Two fields are added in `common.ReplicaSpec`: `minReplicas` and `maxReplicas`. They acts as MIN_SIZE and MAX_SIZE in the elastic example above.
```
alculquicondor, line 104: Do these fields make sense for every other operator?

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen=true
+// ReplicaSpec is a description of the replica
+type ReplicaSpec struct {
+	// Replicas is the desired number of replicas of the given template.
+	// If unspecified, defaults to 1.
++	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
+
++	// minReplicas is the lower limit for the number of replicas to which the training job
```
alculquicondor, line 84: Can users specify `MinReplicas` or `MaxReplicas` but not both?
How is `nil` interpreted?
Also, in k8s we prefer the word `null` instead of `nil`, as users might be using a language different than go to build their configurations.

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+```yaml
+apiVersion: "kubeflow.org/v1"
+kind: "PyTorchJob"
+metadata:
+  name: "pytorch-dist-mnist"
+spec:
+  pytorchReplicaSpecs:
+    # There is no master in elastic training jobs.
+    Worker:
+      minReplicas: 3
```
alculquicondor, line 117: if I use `minReplicas` and `maxReplicas` then `replicas` should be `nil`?

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+              command: "python -m torch.distributed.run --rdzv_backend=c10d --rdzv_endpoint=$KUBEFLOW_RDZV_HOST:$KUBEFLOW_RDZV_PORT --nnodes=$KUBEFLOW_MIN_SIZE:$KUBEFLOW_MAX_SIZE --nproc_per_node=1 xxx.py"
+```
+
+There are four environment variables here: `KUBEFLOW_RDZV_HOST`, `KUBEFLOW_RDZV_PORT`, `KUBEFLOW_MIN_SIZE` and `KUBEFLOW_MAX_SIZE`. The environment variables will be set by the operators.
+
+## Operator
+
+### Environment Variables
+
+`SetPodEnv` in `pkg/controller.v1/pytorch/pytorch.go` should be changed. There is no need to set `RANK`, `WORLD_SIZE`, `MASTER_ADDR`, `MASTER_PORT` if TorchElastic is used. `KUBEFLOW_RDZV_HOST`, `KUBEFLOW_RDZV_PORT`, `KUBEFLOW_MIN_SIZE` and `KUBEFLOW_MAX_SIZE` Should be set instead.
```
alculquicondor, line 134: what determines if a PyTorchJob is elastic or not? whether `minReplicas` and `maxReplicas` are different than `nil`?

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+One built-in port named `kubeflow-rdzv-port` is introduced for `rendezvous`.
+
+### Reconciliation
+
+`JobController.ReconcilePods` should be refactored. Now the pods are returned by `GetPodSlices`. For example, if `spec.Replicas` is 3, the PodSlices may look like: `[[0],[1],[2]]`. It is not expected when elastic training is enabled.
+
+```go
+// ReconcilePods checks and updates pods for each given ReplicaSpec.
+// It will requeue the job in case of an error while creating/deleting pods.
+func (jc *JobController) ReconcilePods(
```
alculquicondor, line 149: Is this code new code or existing code?

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+
+`KUBEFLOW_RDZV_HOST` will be set to `<name>-worker-0`, `KUBEFLOW_RDZV_PORT` will be set to 29500 by default. `KUBEFLOW_MIN_SIZE` and `KUBEFLOW_MAX_SIZE` will be set to `${pytorchjob.spec.replicas[worker].minReplicas}` and `${pytorchjob.spec.replicas[worker].macReplicas}`.
+
+### Ports
+
+One built-in port named `kubeflow-rdzv-port` is introduced for `rendezvous`.
+
+### Reconciliation
+
+`JobController.ReconcilePods` should be refactored. Now the pods are returned by `GetPodSlices`. For example, if `spec.Replicas` is 3, the PodSlices may look like: `[[0],[1],[2]]`. It is not expected when elastic training is enabled.
```
alculquicondor, line 144: What should `GetPodSlices` return for elastic jobs instead?

How does the controller decide how many pods to create?

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+    env:
+    - name: MASTER_PORT
+      value: "23456"
+    - name: KUBEFLOW_RDZV_HOST
+      value: ${pytorchjob.metadata.name}-worker-0
+    - name: KUBEFLOW_MIN_SIZE
+      value: "${pytorchjob.spec.replicas[worker].minReplicas}"
+    - name: KUBEFLOW_MAX_SIZE
+      value: "${pytorchjob.spec.replicas[worker].macReplicas}"
+    command: "python -m torch.distributed.run --rdzv_backend=c10d --rdzv_endpoint=$KUBEFLOW_RDZV_HOST:$KUBEFLOW_RDZV_PORT --nnodes=$KUBEFLOW_MIN_SIZE:$KUBEFLOW_MAX_SIZE --nproc_per_node=1 xxx.py"
```
alculquicondor, line 197: Could the operator set a default `command` and then users can use the `args` to append more arguments and the python file?

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+	Template v1.PodTemplateSpec `json:"template,omitempty"`
+
+	// Restart policy for all replicas within the job.
+	// One of Always, OnFailure, Never and ExitCode.
+	// Default to Never.
+	RestartPolicy RestartPolicy `json:"restartPolicy,omitempty"`
+}
+```
+
+### Autoscaler Integration
```
alculquicondor, line 264: Is this part of the "alternatives considered"?

proposals/pytorch-elastic-proposal.md
gaocegege: No, just a node which manages underlying worker processes.

proposals/pytorch-elastic-proposal.md
gaocegege: > Can users specify MinReplicas or MaxReplicas but not both?

They can set both.

> How is nil interpreted?

They are optional. Users need to set min + max or just replicas.

gaocegege: > Also, in k8s we prefer the word null instead of nil, as users might be using a language different than go to build their configurations.

OK SGTM

proposals/pytorch-elastic-proposal.md
gaocegege: In the current design, I think so.

proposals/pytorch-elastic-proposal.md
gaocegege: https://github.com/kubeflow/common/blob/master/pkg/controller.v1/common/pod.go#L315

proposals/pytorch-elastic-proposal.md
gaocegege: I think we need to support it, but I am not sure how it affects API design. As you know, there is no built-in resource in Kubernetes which has minReplicas and maxReplicas except Autoscaler. 

I was thinking if we should put them in the PyTorchJob CRD. Thus make it in the alternatives considered

proposals/pytorch-elastic-proposal.md
gaocegege: Ref https://pytorch.org/docs/stable/distributed.html

```
python -m torch.distributed.launch --nproc_per_node=<X> --nnodes=<Y>
```

There are Y nodes,  X workers per node. Thus the number of worker processes should be X * Y.

In Kubernetes, we will get Y pods, X processes per pod.

proposals/pytorch-elastic-proposal.md
gaocegege: Personally, I think so. But we should discuss it further /cc @kubeflow/wg-training-leads

proposals/pytorch-elastic-proposal.md
zw0610: > Can users specify MinReplicas or MaxReplicas but not both?

Are you asking whether user has to specify at least one of [MinReplicas, MaxReplicas]?

Jeffwan:
Nice proposal. I will leave some comments tomorrow.

proposals/pytorch-elastic-proposal.md
alculquicondor: oh these are nodes from Pytorch's perspective, not k8s. Maybe say Pytorch node. But I think Pod would be more clear as this document refers to the infrastructure more than the framework.

proposals/pytorch-elastic-proposal.md
alculquicondor: > Are you asking whether user has to specify at least one of [MinReplicas, MaxReplicas]?

Yes, I'm asking whether specifying ONLY MinReplicas or MaxReplicas is valid. If it's not valid, it should be part of the API comment. But this can be left to the code review.

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,295 @@
...
+apiVersion: "kubeflow.org/v1"
+kind: "PyTorchJob"
+metadata:
+  name: "pytorch-dist-mnist"
+spec:
+  pytorchReplicaSpecs:
+    # There is no master in elastic training jobs.
+    Worker:
+      minReplicas: 3
+      MaxReplicas: 5
```
andreyvelich, line 119:
```suggestion
      maxReplicas: 5
```

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,295 @@
...
+
+There are four environment variables here: `KUBEFLOW_RDZV_HOST`, `KUBEFLOW_RDZV_PORT`, `KUBEFLOW_MIN_SIZE` and `KUBEFLOW_MAX_SIZE`. The environment variables will be set by the operators.
+
+## Operator
+
+### Environment Variables
+
+`SetPodEnv` in `pkg/controller.v1/pytorch/pytorch.go` should be changed. There is no need to set `RANK`, `WORLD_SIZE`, `MASTER_ADDR`, `MASTER_PORT` if TorchElastic is used. `KUBEFLOW_RDZV_HOST`, `KUBEFLOW_RDZV_PORT`, `KUBEFLOW_MIN_SIZE` and `KUBEFLOW_MAX_SIZE` Should be set instead.
+
+`KUBEFLOW_RDZV_HOST` will be set to `<name>-worker-0`, `KUBEFLOW_RDZV_PORT` will be set to 29500 by default. `KUBEFLOW_MIN_SIZE` and `KUBEFLOW_MAX_SIZE` will be set to `${pytorchjob.spec.replicas[worker].minReplicas}` and `${pytorchjob.spec.replicas[worker].macReplicas}`.
```
andreyvelich, line 137:
```suggestion
`KUBEFLOW_RDZV_HOST` will be set to `<name>-worker-0`, `KUBEFLOW_RDZV_PORT` will be set to 29500 by default. `KUBEFLOW_MIN_SIZE` and `KUBEFLOW_MAX_SIZE` will be set to `${pytorchjob.spec.replicas[worker].minReplicas}` and `${pytorchjob.spec.replicas[worker].maxReplicas}`.
```

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Ports
+
+One built-in port named `kubeflow-rdzv-port` is introduced for `rendezvous`.
```
andreyvelich, line 141: Will this port be equal `29500` ?

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,295 @@
...
+    name: worker
+    env:
+    - name: MASTER_PORT
+      value: "23456"
+    - name: KUBEFLOW_RDZV_HOST
+      value: ${pytorchjob.metadata.name}-worker-0
+    - name: KUBEFLOW_MIN_SIZE
+      value: "${pytorchjob.spec.replicas[worker].minReplicas}"
+    - name: KUBEFLOW_MAX_SIZE
+      value: "${pytorchjob.spec.replicas[worker].macReplicas}"
```
andreyvelich, line 197:
```suggestion
      value: "${pytorchjob.spec.replicas[worker].maxReplicas}"
```

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,295 @@
...
+    # KUBEFLOW_RDZV_PORT is set to 29500 by default in TorchElastic.                
+    - containerPort: 29500
+      name: kubeflow-rdzv-port
+      protocol: TCP
+
+```
+
+## Limatations
+
+- KUBEFLOW_RDZV_PORT will be open for every pod even though workers except worker-0 do not use it.
```
andreyvelich, line 209: Can we add and expose this port only for Worker-0 pod ?

proposals/pytorch-elastic-proposal.md
andreyvelich: We might need to explain how we will refactor this part.

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,295 @@
...
+    Worker:
+      minReplicas: 3
+      MaxReplicas: 5
+      restartPolicy: OnFailure 
+      template:
+        spec:
+          containers: 
+            - name: pytorch
+              image: <image>
+              command: "python -m torch.distributed.run --rdzv_backend=c10d --rdzv_endpoint=$KUBEFLOW_RDZV_HOST:$KUBEFLOW_RDZV_PORT --nnodes=$KUBEFLOW_MIN_SIZE:$KUBEFLOW_MAX_SIZE --nproc_per_node=1 xxx.py"
```
Jeffwan, line 126: Let's make sure it's user's choice on rendezvous backends?  We don't want to manage this part, right?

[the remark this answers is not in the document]
Jeffwan: Considering the user experience, I would suggest to help them "concatenate command". envs like `KUBEFLOW_RDZV_HOST ` would be blackbox to them. We can do following way

1. Check commands contains `rdzv` commands. 
 -> If so, check all `backend` `endpoint` exist, if not, fail validation
 -> if not.  find the index to insert the commands. This is a little bit tricky and it could be problematic.

gaocegege: I think so. There are three backends now: static, c10d, etcd. And users can also implement their own backend like redis and so on. Users can specify it manually. If they use c10d, we can set rdzv endpoint for them. If they use etcd, they can set the endpoint by themselves.

proposals/pytorch-elastic-proposal.md
gaocegege: Maybe we can, do you mean we deal with it with a custom condition loop?

proposals/pytorch-elastic-proposal.md
gaocegege: It is 29500 by default, while TorchElastic supports custom port.

proposals/pytorch-elastic-proposal.md
gaocegege: I will illustrate more about it.

proposals/pytorch-elastic-proposal.md
gaocegege: SGTM will add more details.

proposals/pytorch-elastic-proposal.md
gaocegege: SGTM, will update it.

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,295 @@
...
+    Worker:
+      minReplicas: 3
+      maxReplicas: 5
+      restartPolicy: OnFailure 
+      template:
+        spec:
+          containers: 
+            - name: pytorch
+              image: <image>
+              command: "python -m torch.distributed.run --rdzv_backend=c10d --rdzv_endpoint=$KUBEFLOW_RDZV_HOST:$KUBEFLOW_RDZV_PORT --nnodes=$KUBEFLOW_MIN_SIZE:$KUBEFLOW_MAX_SIZE --nproc_per_node=1 xxx.py"
```
zw0610, line 126: So users who launches a pytorch elastic job has to add `--rdzv_endpoint=$KUBEFLOW_RDZV_HOST:$KUBEFLOW_RDZV_PORT --nnodes=$KUBEFLOW_MIN_SIZE:$KUBEFLOW_MAX_SIZE` with these exact environment variables name in the command? This seems a bit user unfriendly, compared to a regular pytorch job where pytorch will read the environment variables directly (instead of from the launching arguments).

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,295 @@
...
+
+- KUBEFLOW_RDZV_PORT will be open for every pod even though workers except worker-0 do not use it.
+
+## Alternatives Considered
+
+### API/CRD
+
+[TorchElastic operator](https://github.com/pytorch/elastic/blob/master/kubernetes/api/v1alpha1/elasticjob_types.go) implemented by @jeffwan puts the new fields under `PyTorchJobSpec`.
+
+Personally, prefer keeping it in `common.ReplicaSpec` since other Jobs may also need it.
```
Jeffwan, line 218: Agree on this direction

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,295 @@
...
+
+```diff
+// PyTorchJobSpec is a desired state description of the PyTorchJob.
```
Jeffwan, line 221: API side. Let's take this into consideration. We want to support user specified pods to scale in as an optional field to give more granular control
https://github.com/kubeflow/mpi-operator/issues/410

proposals/pytorch-elastic-proposal.md
gaocegege: Yes, I think so. Prefer the latter.

proposals/pytorch-elastic-proposal.md
gaocegege: Maybe users want to set their own entrypoint in the command, I think. Thus it may be better to keep command here, WDYT

proposals/pytorch-elastic-proposal.md
andreyvelich: Do we provide APIs to specify this port other than 29500 ?

proposals/pytorch-elastic-proposal.md
andreyvelich: Yes, for example. If we don't want to have additional ports.

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,296 @@
...
+* [Jiaxin Shan(@jeffwan)](https://github.com/jeffwan) - Bytedance
+* [Wang Zhang(@zw0610)](https://github.com/zw0610) - Tencent
+
+_Status_
+
+- 2021-08-16 Draft v1
+
+## Abstract
+
+[TorchElastic](https://pytorch.org/docs/1.9.0/distributed.elastic.html), which was open sourced over a year ago in the pytorch/elastic github repository, is a runner and coordinator for PyTorch worker processes. it has been part of PyTorch core since 1.9.0. This proposal is to support such feature with the help of PyTorchJob.
```
terrytangyuan, line 13: What's involved to support additional frameworks with elastic capabilities?

gaocegege:
Ref https://github.com/pytorch/pytorch/issues/67616

Still cannot tolerate agent failures with PyTorch 1.10. I will dive deep into it.

gaocegege:
https://github.com/pytorch/pytorch/issues/67742

There is an upstream bug in PyTorch. It does not support SIGTERM when scaling down the workers. After the patch, the MVP works well. I will refine the proposal these days.

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,332 @@
...
+type ReplicaSpec struct {
+	// Replicas is the desired number of replicas of the given template.
+	// If unspecified, defaults to 1.
++	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
+
++	// minReplicas is the lower limit for the number of replicas to which the training job
++	// can scale down.  It defaults to null.
++	// +optional
++	MinReplicas *int32 `json:"minReplicas,omitempty"`
```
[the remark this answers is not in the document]
gaocegege: > Maybe we can abandon the Master replica type and move the Master-specific fields into a DDPPolicy.

In elastic training, we do not need master. We can keep min and max in ElasticPolicy, WDYT

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,332 @@
...
+* [Jiaxin Shan(@jeffwan)](https://github.com/jeffwan) - Bytedance
+* [Wang Zhang(@zw0610)](https://github.com/zw0610) - Tencent
+
+_Status_
+
+- 2021-08-16 Draft v1
+
+## Abstract
+
+[TorchElastic](https://pytorch.org/docs/1.9.0/distributed.elastic.html), which was open-sourced over a year ago in the [pytorch/elastic](https://github.com/pytorch/elastic) GitHub repository, is a runner and coordinator for PyTorch worker processes. it has been part of PyTorch core since 1.9.0. This proposal is to support such a feature with the help of PyTorchJob.
```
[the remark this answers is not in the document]
gaocegege: It is copied from the description of torch elastic. Torch elastic is the daemon process that manages the underlying trainer processes in the local machine.

proposals/pytorch-elastic-proposal.md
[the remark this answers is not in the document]
gaocegege: Yes, I refined the proposal, and used the built-in environment variables to do it.

gaocegege:
I refined the proposal, PTAL

There is a demo https://asciinema.org/a/446932 to help us understand the design.

proposals/pytorch-elastic-proposal.md
```diff
@@ -0,0 +1,436 @@
...
+	Active int32 `json:"active,omitempty"`
+
+	// The number of pods which reached phase Succeeded.
+	Succeeded int32 `json:"succeeded,omitempty"`
+	// The number of pods which reached phase Failed.
+	Failed int32 `json:"failed,omitempty"`
+}
+```
+
+Then `PyTorchJob` has the `scale` subResource, then it can work with Autoscaler. The only problem is that `PyTorchJob` already has the minReplicas and maxReplicas fields. They are used to generate commands. The Autoscaler resource needs them, too. Thus users may need to define them again.
```
alculquicondor, line 431: Would it make sense for the pytorch controller to create a HPA object in behalf of the user?

gaocegege: Sounds reasonable. Should we do it in PyTorchJob, or should we make a new abstraction, maybe ElasticJob, to manage the underlying PyTorchJob and the HPA?

alculquicondor: Maybe better start with just PyTorch and add the abstraction later if it ends up working well.

gaocegege:
I updated the details about HPA. BTW, I think we should upgrade the PyTorchJob APIVersion to v2beta1 since it is a breaking change.

alculquicondor:
What is breaking about it? If users don't set elastic configurations, wouldn't it work as before?

proposals/pytorch-elastic-proposal.md
5had3z: PyTorch 1.10 introduced [```torchrun```](https://pytorch.org/docs/stable/elastic/quickstart.html), so you may need to be flexible to accomidate <1.10 versions that use ```python -m torch.distributed.run``` and  >=1.10 with torchrun.

gaocegege:
I double-checked the CRD definition. We can keep compatibility with v1. common.JobStatus is changed but it does not block v1 if we upgrade the kubeflow/common version. Thus we can do it in v1.


```diff
type ReplicaStatus struct {
+	// LabelSelector is the selector for the replica.
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
	// The number of actively running pods.
	Active int32 `json:"active,omitempty"`
	// The number of pods which reached phase Succeeded.
	Succeeded int32 `json:"succeeded,omitempty"`
	// The number of pods which reached phase Failed.
	Failed int32 `json:"failed,omitempty"`
}
```

alculquicondor:
Right, API changes are breaking only if you remove or rename a field. I think the proposed changes look backwards compatible.

gaocegege:

Could you please get another review?

https://github.com/kubeflow/training-operator/pull/1453 The PR is ready, and the coverall coverage increased (+7.1%) to 15.252%, PyTorch related test coverage is increased from 0% to 80%

proposals/pytorch-elastic-proposal.md
gaocegege: Yeah, now we use the built-in environment variables PET_* to do it. Then I think we do not have the problem.

theadactyl:
@kubeflow/wg-training-leads could we get an "/approved" for this?

terrytangyuan (approved):

gaocegege:
I am closing this since it is outdated.

Issue #495, opened 2021-03-02
Proposal: form the "Technical Steering Committee"


thesuperzapper:
## Goals
* Ensure Kubeflow gets back-on-track to implement the community-accepted [governance structure](https://docs.google.com/document/d/1HKB662Ju6URVdzw0Neq9OFnxM6RMXFu4Dd04q9E3kGk/edit)

## Background:

Back on 2019-05-02, the community accepted a governance structure proposal:
https://docs.google.com/document/d/1HKB662Ju6URVdzw0Neq9OFnxM6RMXFu4Dd04q9E3kGk

This governance structure defined a future group called the "Technical Steering Committee" which was intended to be a diverse and community-elected leadership group for Kubeflow.

In order to facilitate the creation of the "Technical Steering Committee", the governance structure defined the following __interim__ governance groups. (NOTE: these groups were not intended to be permanent)

__Working Groups:__
* Purpose: the primary venue for technical decision-making in the project
* Disbanded: if "Technical Steering Committee" imposes a governance change

__Technical Advisory Council:__
* Purpose: a higher authority to resolve escalations from working groups
* Disbanded: upon the creation of the "Technical Steering Committee"
   * NOTE: it seems this group was accidentally disbanded __before the formation__ of the "Technical Steering Committee"

__Project Steering Group:__
* Purpose: to bootstrap the "Technical Steering Committee" by creating all required administrative structures, (like how voting will work)
* Disbanded: upon the creation of the "Technical Steering Committee"

## Steps:

Clearly we have gone a bit off-track during 2020, but I believe we can quickly move to form the "Technical Steering Committee" by undertaking the following actions:

1. The current "Project Steering Group" finalises the roadmap to form the initial "Technical Steering Committee"
   * NOTE: The governance document states __"Bootstrapping initial TSC"__ and __"Lead creation of TSC Charter"__ under  the "Project Steering Group" definition
2. The current "Project Steering Group" presents their proposed roadmap to the community
3. The roadmap is followed, leading to the formation of the "Technical Steering Committee"
   * I believe a __reasonable timeframe__ for the transition from PSG to TSC is: __2-3 Months__
   * Meaning the TSC would be formed and operational around __May 2021__

thesuperzapper:
Hey @kubeflow/project-steering-group, here is my proposal which was spawned out of the discussion in kubeflow-discuss:
https://groups.google.com/g/kubeflow-discuss/c/WlAtv8yBwGE

[...]

PR #524, opened 2021-08-31
Adding Kubeflow Conformance Program Proposal

james-jwu:
The document can also be found at:

https://docs.google.com/document/d/1a9ufoe_6DB1eSjpE9eK5nRBoH3ItoSkbPfxRA0AjPIc/edit?resourcekey=0-IRtbQzWfw5L_geRJ7F7GWQ#heading=h.d5ye3ctn3kvn

### DIFF
--- a/proposals/kubeflow-conformance-program-proposal.md
+++ b/proposals/kubeflow-conformance-program-proposal.md
@@ -1,0 +1,110 @@
+# Kubeflow Conformance Test
+
+James Wu (james-jwu@)  
+2021-07-29
+
+# Overview
+
+The [Kubeflow Project](https://github.com/kubeflow) is currently managed by different [Working Groups](https://github.com/kubeflow/community/blob/master/wg-list.md) whose composition represents a broad spectrum of industry and community. The Kubeflow trademark is owned by Google.   
+The [Kubeflow Brand Guidelines](https://github.com/kubeflow/community/blob/master/KUBEFLOW_BRAND_GUIDELINES.pdf) was published in Mar. 2021. The guideline is broadly applicable to usage of Kubeflow trademark by products, events and publications. While the brand guidelines provide general guidance, it does not prescribe the definition of Kubeflow and what makes a distribution/application "Kubeflow" vs. not "Kubeflow".  
+This document aims to define conformance criteria for the following usage of Kubeflow trademark:
+
+-  For Kubeflow distribution - "Certified Kubeflow"
+-  For Kubeflow application - TBD
+
+The goal is to ensure these special usages of Kubeflow trademark meet common standards of interoperability, increase cohesiveness of the Kubeflow platform, promote customer confidence, and reduce the burden of Kubeflow maintainers.
+
+# Kubeflow Distribution
+
+A conformant Kubeflow Distribution is certified to provide a set of core functionalities and API integration options. A core set of these features are listed in the "[Pipeline tests](#heading=h.71p7gyy4tf5d)" and "[Metadata tests](#heading=h.wc7nytwk964f)'' sections.  
+At the high level, the tests certifies that Kubeflow Pipeline and Metadata are part of the distribution, and exposes a standard set of APIs. Pipeline and Metadata and the binding "glue" for other Kubeflow components. With open standards for certifying Kubeflow Applications (see next section), the Kubeflow conformance criteria gives more flexibility to Kubeflow distributors to provide customization.   
+The tests will be designed in a way similar to [Kubernetes conformance program](https://github.com/cncf/k8s-conformance).  
+The tests will be implemented in stages.  
+Conformant distributions is entitled to refer to the distribution as "Certified Kubeflow". The distribution can be listed under a partner page under the Kubeflow project. The naming of the distribution still needs to follow [Kubeflow Brand Guidelines](https://github.com/kubeflow/community/blob/master/KUBEFLOW_BRAND_GUIDELINES.pdf).
+
+## Example
+
+Company X creates a distribution of Kubeflow and plans to name it "X Kubeflow Service". Company X tries to certify for Kubeflow conformance, which entails:
+
+-  Install the distribution
+-  Runs the Kubflow conformance test suite
+-  Submits the test log as a PR to Open Usage Commons (OUC) for approval.
+-  Upon approval and following trademark guidelines, Company X changes the distribution name to "X service for Kubeflow", or "X Platform (Certified Kubeflow)"
+-  Optionally, the distribution may be listed in a catalog on Kubeflow website
+
+The conformance tests make sure "X service for Kubeflow" supports KFP and Metadata. Company X may include more applications (e.g. TFJob, Katib) in the distribution, but it does not affect the conformance standing of "X service for Kubeflow".
+
+## Pipeline tests
+
+-  Pipeline runtime
+   -  V1 conformance
+      -  The tests use a designated version of KFP SDK to compile a set of pipelines and submit it to the distribution under test.
+
+   -  V2 conformance
+      -  The tests uses a designated version of KFP SDK that compiles pipeline to [IR](https://docs.google.com/document/d/1PUDuSQ8vmeKSBloli53mp7GIvzekaY7sggg6ywy35Dk/edit) (Intermediate Representation)
+      -  The IR is submitted to the pipeline server to exercise and verify different KFP features.
+
+   -  Below is a categorization of the features and is not meant to be exhaustive.
+      -  Artifact and parameter passing
+      -  Caching
+      -  Executors: container / importer / resolver
+      -  Control flow features: ParallelFor, conditional, exit handler
+      -  A subset of Kubernetes features (e.g. secrets, volume)
+
+-  Pipeline management
+   -  Pipeline template management
+   -  Pipeline run management (Get / Delete / Cancel / Archive / Retry)
+   -  Recurring runs
+
+## Metadata tests
+
+-  Metadata lineage - pipeline runs must produce the correct lineage graph including
+   -  Context
+   -  Execution
+   -  Artifact
+   -  Event
+
+-  Metrics - verifies metrics artifacts are produced
+-  Metadata APIs (future work)
+
+## First version of conformance
+
+The first version of conformance will be limited to V1 Pipeline Runtime conformance.
+
+# 
+
+# Kubeflow Application
+
+Kubeflow Application certification verifies the application-under-test can integrate well with other Kubeflow Applications. Metadata generation is automatic when Kubeflow Application conforms to standard Kubeflow Pipelines component interface.  
+Kubeflow Application is entitled to refer to the application as "Kubeflow Native" or <TBD>. The application may be listed under an application catalog (to be created) under Kubeflow project. The naming of the application still needs to follow [Kubeflow Brand Guidelines](https://github.com/kubeflow/community/blob/master/KUBEFLOW_BRAND_GUIDELINES.pdf).  
+NOTE: existing projects in Kubeflow org do not require certification.
+
+## Example
+
+Company X creates a Kubernetes Custom Resource for model training, and wishes to certify the feature for Kubeflow Application. Company X needs to:
+
+-  Create a Kubeflow Pipelines component for launching the custom resource, with inputs and outputs appropriately defined using parameters and artifacts. The component may be published as a Python function or YAML.
+-  Runs conformance tool against the Python source, by specifying the source file, and the component function (in the case of Python function).
+-  Submits the test results to OUC for approval.
+-  Upon approval, Company X may name the component "X Training for Kubeflow", "X Training <TBD>"
+
+## Test principles
+
+Kubeflow Application conformance test verifies the component function under test conforms to the Kubeflow Pipelines component definition. 
+
+Proposed CLI:
+
+```
+$ kfp conformance verify-component --file=my_component.py --component_function=my_component
+
+$ kfp conformance verify-component - file=my_component.yaml
+```
+
+The tests verify the component integrates well with the following:
+
+-  Kubeflow Pipeline integration: a well defined component interface ensures the Kubeflow Application under test plays well with other Kubeflow Applications. The test will not try to verify functionality or code quality.
+-  Metadata: Kubeflow Pipelines automatically records the input/output parameters and artifacts in metadata. The tests verify the component interfaces. Kubeflow Application candidates can optionally emit metadata, either by using output_metadata mechanism (to be explained), or some other mechanism added to KFP in the future. Kubeflow Application candidates are encouraged to log additional metadata to MLMD but are not required to do so.
+
+## References
+
+-  Prior [discussion](https://groups.google.com/g/kubeflow-discuss/c/d6whgEgror8) in Kubeflow community on Kubeflow conformance


proposals/kubeflow-conformance-program-proposal.md
```diff
@@ -0,0 +1,111 @@
...
+
+Company X creates a distribution of Kubeflow and plans to name it "X Kubeflow Service". Company X tries to certify for Kubeflow conformance, which entails:
+
+-  Install the distribution
+-  Runs the Kubflow conformance test suite
+-  Submits the test log as a PR to Kubeflow Trademark Team for approval.
+-  Upon approval and following trademark guidelines, Company X changes the distribution name to "X service for Kubeflow", or "X Platform (Certified Kubeflow)"
+-  Optionally, the distribution may be listed in a catalog on Kubeflow website
+
+The conformance tests make sure "X service for Kubeflow" supports KFP and Metadata. Company X may include more applications (e.g. TFJob, Katib) in the distribution, but it does not affect the conformance standing of "X service for Kubeflow".
```
andreyvelich, line 36: Let's continue our discussion on the Kubeflow Community Meeting.

Why including Training Operators and Katib in the distribution conformance test is optional ?
Since these components are in Kubeflow core, does it make sense to verify that distribution is working with the core functionality and the APIs ?

From my point of view, Kubeflow distribution should support all Kubeflow core functionality to provide complete Cloud Native MLOps experience for the users.
In addition to that, it helps contributions to get a feedback from the distribution owners.

What do others think @kubeflow/wg-training-leads @kubeflow/wg-automl-leads @jbottum ?

andreyvelich (commented):
Thank you for driving this @james-jwu!
This is very helpful.

proposals/kubeflow-conformance-program-proposal.md
gaocegege: As you know, some platforms adopt a subset of Kubeflow components, should we have different conformance tests for different components?

nakfour: @andreyvelich I agree with you. Only including pipelines and metadata in the KF certification will confuse both developers and users. We specifically list multiple components as core components in KF (https://www.kubeflow.org/docs/components/) and just certifying one will cause confusion with no clear direction. @james-jwu are we only certifying pipelines because we have plans for other components similar to KFServing graduating?  It feels like this community is not transparent in a lot of strategic KF direction.

james-jwu: One of the premises of KF Distribution Certification is that Kubeflow endeavored to be neutral and not endorse one distribution over another.  In addition, Kubeflow Distributions allows customization by distro vendors and provides the flexibility of choosing components to include in the distro. As Ce said, not all platforms adopt the full set of components currently under the Kubeflow project. With the introduction of Kubeflow Application Certification (proposed name), we envision many projects outside of Kubeflow project contributing the the Kubeflow ecosystem. KFServing’s graduation is one example where the certification program can easily accommodate projects not directly under Kubeflow.

The point about basing the tests on Pipeline and Metadata is not that they are the totality of “Kubeflow”.  Pipeline and Metadata provides the API and common layer that binds all the components. [quote from PR] “_Pipeline and Metadata are the binding "glue" for other Kubeflow components. With open standards for certifying Kubeflow Applications (see next section), the Kubeflow conformance criteria gives more flexibility to Kubeflow distributors to provide customization._” I also want to point to the long [discussion](https://groups.google.com/g/kubeflow-discuss/c/d6whgEgror8) we had last year, which raised some excellent points about why pipeline and metadata could be good choices for the minimal common layer that brings cohesion to Kubeflow ([[1]](https://groups.google.com/g/kubeflow-discuss/c/d6whgEgror8/m/yw1lHf5nBgAJ), [[2]](https://groups.google.com/g/kubeflow-discuss/c/d6whgEgror8/m/U9MYsyRTBgAJ)).
The point about providing a complete Cloud Native MLOps experience is well taken. I think most distro vendors will likely include most of the components under https://www.kubeflow.org/docs/components/ anyway. Forcing all distros to include all components is IMO not necessarily going to promote the growth of Kubeflow.

andreyvelich: > As Ce said, not all platforms adopt the full set of components currently under the Kubeflow project.

That's true, but one of the biggest advantages of using Kubeflow is that our users can deploy necessary components or deploy Kubeflow as an end-to-end MLOps platform. From the recent [Kubeflow User Panel](https://youtu.be/nFXL48x0Lk0), we saw that users want to have both options, right @jbottum ?
Conformance Program for each Kubeflow Core component might help to make distribution compatible with the Kubeflow Core.

As I can see from the thread discussion, the problem is that other Kubeflow components are loosely coupled.
Our future work should be focused on how to improve it, isn't it ?
For example, we are building integration between Katib and distributive Training Operators.

As well from the thread, I can point this message:
```
However, I imagine that the value of Kubeflow is offering a solution to the MLOps and I'd perhaps suggest even to the ResearchOps side of things, both of which are not a solved problem just yet.
```

Can Pipelines and Metadata offer end to end MLOps and ResearchOps infrastructure without other Kubeflow components ?

>  I think most distro vendors will likely include most of the components under https://www.kubeflow.org/docs/components/ anyway. 

Are we planing to add all components which pass Kubeflow Application Certification under https://www.kubeflow.org/docs/components ?

nakfour: @james-jwu I think I have a better understanding of why only pipelines and metadata are only included in the conformance test. You are basically saying to get the certified KF logo at a minimum get pipelines and metadata to pass conformance versus asking for all components to get certified. I don't agree with your statement "_Pipeline and Metadata are the binding "glue" for other Kubeflow components_" we have Data Scientists users that perceive Jupyter Notebooks as the "glue". From a devops/mlops perspective yes Pipelines are the "glue".  I have seen other communities offer levels or tiers of certification or conformance, for example Kubeflow conformance would be pipelines and metadata and Kubeflow+ would pipelines, metadata and all other components. I am not clear on what Kubeflow Application conformance will cover. Would it be possible to have a two tier conformance testing? I am concerned that excluding some Kubeflow components from conformance testing will slow adoption and cause user and developer confusion.

james-jwu: Andrey, Juana, thanks for your feedback. You have some great points, and I think we can address them in an iterative way. One thing we learned from the Kubernetes Conformance program is to start small. The initial Kubernetes conformance tests ([v1.10](https://github.com/cncf/k8s-conformance/blob/master/docs/KubeConformance-1.10.md)) had only 174 tests, and [later versions](https://github.com/cncf/k8s-conformance) expanded that set significantly. We think Kubeflow Conformance will also evolve over time, and adapt to the needs of Kubeflow community, distributors and users.

I think starting the conformance program with the base interop layer (i.e. pipeline + metadata) sets a good path to an inclusive Kubeflow ecosystem, where distributors have assurance of the interoperability of components/applications that are bundled into the distro. We will probably learn a lot in the process. The feedback and information gathered from this effort can guide us to the next steps, which could expand the conformance scope.

With the initial scope of the conformance program, I think it is totally okay to produce a Certified distro that in its description has something along the lines of “Supports MLOps components under the official Kubeflow project: [https://www.kubeflow.org/docs/components](https://www.kubeflow.org/docs/components).”

The idea about 2 tiers of conformance is interesting. I agree that Notebook is an essential tool for data scientists. Tests for Notebook seem non-specific to Kubeflow, so the right scope of the tests is a little unclear to me at this point. What constitutes “core components” may warrant a discussion as well in the context of conformance.

RE - “Are we planning to add all components which pass Kubeflow Application Certification under https://www.kubeflow.org/docs/components” - the answer is no. Certified Application will have their own catalog, not to be confused with components living in the Kubeflow project.

nakfour: @james-jwu thank you very much for taking our input into consideration. I think it makes sense to start small with Pipelines and Metadata and plan for more components to be included later. As long as we are clear to our Users and Developers that more components testing will be included in the future and its part of the roadmap. Is it possible to add a sentence or section in the document reflecting the plan to include more KF components to conformance testing in later stages? Thanks

thesuperzapper:
@james-jwu @Bobgy @theadactyl Thanks for the effort on this!

Firstly, I think we should discuss certifying "Kubeflow Distributions" and "Kubeflow Applications" separately.

Secondly, I am worried these proposals effectively define "Kubeflow" as "Kubeflow Pipelines", when Kubeflow has always aimed to be "the ML ecosystem for Kubernetes".

## Thoughts on "Kubeflow Application" certification:

I strongly agree that we should "certify" which applications "work with Kubeflow", but 
we must be careful of how we do this, otherwise we will undermine our goal of being 
"the ML ecosystem for Kubernetes".

This proposal defines a "Certified Kubeflow Application" as supporting "Kubeflow Pipelines" and "Kubeflow Pipelines Metadata" (which ARE important parts of Kubeflow), however, this definition is not compatible with Kubeflow being "the ML ecosystem for Kubernetes", as not every user will need or want Kubeflow Pipelines.

A more broad definition of a "Certified Kubeflow Application" would be that it:
1. "runs completely inside a K8S cluster"
2. "if it has a UI, it is accessible from Central Dashboard"

Therefore, I propose:
1. The @kubeflow/wg-pipeline-leads consider using the existing proposal as their "Kubeflow Pipelines Conformance Program"
2. The Kubeflow community uses a more broad definition of a "Certified Kubeflow Application" (possibly using the definition above)

## Thoughts on "Kubeflow Distribution" certification:

I strongly agree that we need to protect the Kubeflow brand, and ensure some minimum expectations of a "Kubeflow Distribution". However, if Kubeflow is "the ML ecosystem for Kubernetes", we should certify distributions based off their "ability to support each component", rather than the inclusion of a specific component like Kubeflow Pipelines.

Once we agree on what a "Certified Kubeflow Application" is, it's pretty clear that a "Certified Kubeflow Distribution" should be able to run any of those applications (along with some other sanity checks).

Therefore, I propose the following "tests" for a "Certified Kubeflow Distribution":

### 1 - "Distinct Project Test"
- Maintains all distribution __documentation__ separate from the Kubeflow Org
- Maintains all distribution __code__ separate from the Kubeflow Org
- Maintained by a __separate entity__ to the Kubeflow Org

### 2 - "Active Project Test"
- Offers upgrades to `Kubeflow 1.X` within `6` months of release
- Actively resolves issues raised by their users _(need to clarify what metric is used here)_

### 3 - "Certified Applications Test"
- Allow users to install ANY "Certified Kubeflow Application"
- Provides the "Central Dashboard" component

## Closing thoughts

I would encourage the working group leads to consider their own Certification programs, which can further verify that a distribution has "good support" for your specific application.

We need to carefully consider our actions here, if we don't do this right, we might accidentally define Kubeflow as something other than "the ML ecosystem for Kubernetes", and irreparably fragment the project.

proposals/kubeflow-conformance-program-proposal.md
thesuperzapper: I have raised some very similar concerns in my comment here: https://github.com/kubeflow/community/pull/524#issuecomment-916568247

@andreyvelich @gaocegege @johnugeorge @nakfour @james-jwu

james-jwu: @nakfour Sounds good. I will provide an update after collecting more feedback.

james-jwu:
@thesuperzapper Thanks for putting a lot of thought into this topic. Kubeflow website summarizes the project as “The Machine Learning Toolkit for Kubernetes”. I completely agree with you that the conformance test should promote that message.

For “Certified Kubeflow Application'', the definition proposed above - a) runs in k8s, and b) UI is accessible from Central Dashboard - is a weaker test than the RFC. The 2 criteria do not really speak strongly of ML, and are too general in my opinion to stand as the test for an ML application.

On the proposed “Certified Distribution” tests: the conformance tests are not designed to enforce distribution channels, nor does it strictly enforce quality. I understand the WGs had agreed on some of these points (particular #1) before. The Kubernetes Conformance tests is a very narrow subset of the functional tests over the API surface, and we intend to follow that model. As I responded to @nakfour, we want to start small in the initial program, and are open to including other Kubeflow components in the test suite in a future version of Kubeflow conformance.

Bobgy:
I personally think the "Distinct Project Test" is valuable in the long term, maybe we can consider including them to the conformance program gradually in the future.
However, for current proposal, as @james-jwu pointed out, we don't want to disrupt existing distributions too much.

Bobgy:
IIUC, Active project test is somewhat implied -- a distribution owner needs to run the test suite and submit test logs to be certified. I don't think someone will do so for inactive projects.

proposals/kubeflow-conformance-program-proposal.md
james-jwu: I added a few point we discussed in the previous community meeting. PTAL. @nakfour @thesuperzapper @animeshsingh @jbottum . The summary is:

- Kubeflow distribution
  - Added statement about expanding the conformance scope in the future
  - Added information about conformance versioning and expiration
  - Added out-of-scope issues for conformance tests: product quality/supportability, distribution development/channel
- Kubeflow application
  - Added the requirement that Kubeflow Application must be Kubernetes Application through self-attestation on deployment mechanism.

thesuperzapper:
@Bobgy @james-jwu 

Right now, Kubeflow has a number of apps that would NOT meet the second `Kubeflow Native Application` criterion, specifically the `"have a KFP Component"`, and `"automatically generates KFP metadata"` requirements.

Taking a step back, I think the purpose of `Kubeflow Native Application` should be about what EXTERNAL things `"work with Kubeflow"`, not what apps are `"part of Kubeflow"`.

We (as a community) should already "know" what apps are `"part of Kubeflow"` (in a core sense).

__Therefore, I propose:__
1. __(Seperatly) the community agrees on what defines the core `"parts of Kubeflow"`.__
    - A starting point might be:    
       - "Pipelines UI"
       - "Pipelines Metadata"
       - "Notebook Spawner UI" 
       - "Notebooks Controller"
       - "Central Dashboard UI",
       - "Profile Controller"
       - "KFAM API"
       - (NOTE: "KServe" is not included, as its now a separate project)
2. __Replace `"Kubeflow Native"` certification with a `"works with Kubeflow"` certification.__
    - We could initially use the proposed criteria:
       - "has a KFP Component"
       - "automatically generates KFP metadata"
       - "is a Kubernetes Application" 
           - (perhaps this is not needed anymore, especially if we want to certify apps like: "BigQuery", "Redshift")

terrytangyuan:
Second @thesuperzapper's comment above. More specifically:
1. We should not overuse/mislead the word "native" and invent a new concept. Either "works with Kubeflow" or "integrates with Kubeflow" is more intuitive to me.
1. I agree that the community should have a stake in what defines the core "parts of Kubeflow" instead of an explicit requirement from the steeling committee that only consists of 3 Googlers (related to https://github.com/kubeflow/community/issues/495).
1. The conformance program sounds more specific to KFP conformance program instead of the entire Kubeflow community projects. I am concerned that this may frustrate our existing contributors to other Kubeflow subprojects. Even though we may expand the scope in the future, the fact that the first 1.0 version doesn't include many other important subprojects is confusing. To me "A conformant Kubeflow Distribution is certified to provide a set of core functionalities and API integration options" feels like all other subprojects (many of them have huge industries adoptions) are not important.

I am cc'ing other WG leads to comment as well: @kubeflow/wg-automl-leads @kubeflow/wg-deployment-leads @kubeflow/wg-manifests-leads @kubeflow/wg-notebooks-leads @kubeflow/wg-pipeline-leads @kubeflow/wg-serving-leads @kubeflow/wg-training-leads

james-jwu:
@thesuperzapper  I think our proposals are conceptually similar:

- You 1st point about what makes up Kubeflow should be defined by "a set of core functionalities and API integration options" (~line 22), hence the responsibility of “Kubeflow Distribution” certification. One difference is that I would give much higher priority to the API over UI, partly because of the lack of UI tests in Kubernetes conformance model, partly because UI tests are usually finicky. 

- “Certified Kubeflow Application” (I am open to other naming) covers your 2nd point about “works with Kubeflow”.

The RFC also grandfathers in all the current Kubeflow components under “Kubeflow Application” certification (~line 111). I feel the “Kubeflow Application” designation might be important to current KF components.

james-jwu:
A new update to the RFC is uploaded. I think it addressed the comments and concerns raised so far.

**Major updates:**
- KF Distribution Certification
  - Proposed to include test contributions for all WGs, with some guidelines. I think we can make a lot of progress this way in the first version.
- KF Application Certification: 
  - Reworded the guiding principle to ".... verifies that the application under test integrates well with Kubeflow."
  - Removed the clause "existing KF components is automatically certified". Components in KF project participates in distribution certification, but not application certification.
  - Left in the "k8s application" requirement for discussion
@thesuperzapper @terrytangyuan @andreyvelich @nakfour @jbottum @animeshsingh

thesuperzapper:
@james-jwu its looking a lot better, here are my comments for the "Certified Kubeflow" section

----
Firstly, "X service for Kubeflow" probably shouldn't be one of the "protected" names.

Instead, I think being "certified" should entitle you to:
1. Append `(Certified Kubeflow)` to the end of your distribution name:
    - example 1: `SuperKF (Certified Kubeflow)`
    - example 2: `AI-IS-FUN (Certified Kubeflow)`
2. Display the "certified" logo:
     - (for example, similar to the K8S one)
     - 

----

Secondly, I want to draw everyone's attention to the [Certified Kubererntes](https://www.cncf.io/certification/software-conformance/) principals:
1. __Consistency:__ Users want consistency when interacting with any installation of Kubernetes.
2. __Timely updates:__  To remain certified, vendors need to provide the latest version of Kubernetes yearly or more frequently, so you can be sure that you’ll always have access to the latest features the community has been working hard to deliver.
3. __Confirmability:__ Any end user can confirm that their distribution or platform remains conformant by running the identical open source conformance application (Sonobuoy) that was used to certify.

I note that we are explicitly excluding `Product quality and supportability` in the latest revision, I thought we were trying to align with the "Certified Kubernetes" program?

PR #537, opened 2021-11-22
Improve the handbook with feedback from the 1.4 Release Retro

kimwnasptd:
I'm augmenting the release handbook to include discussion points from the 1.4 release retro we had  https://github.com/kubeflow/community/blob/master/releases/retrospectives/release-1.4.md 

Points addressed:
1. Release team selection process
2. Encouragement to have docs PRs alongside feature PRs
3. Add more details in the Timeline, regarding RCs
4. Explicitly mention that we should also have a tracking issue for the release

I also think that we should look into breaking the handbook into some smaller files, like `roles`, `timeline` etc. It's currently a huge 300+ lines and will keep getting bigger.

/cc @annajung 
/cc @shannonbradshaw 
/cc @jbottum

### DIFF
--- a/releases/handbook.md
+++ b/releases/handbook.md
@@ -194,7 +194,11 @@ Criteria for timeline that the team needs to consider
 - Kubecon dates - let’s not hard block on events, but keep them in mind since we know community members might get doublebooked.
 - Associated events (aka. AI Day at Kubecon, Tensorflow events) - we want to keep them in mind.
 
-**Success Criteria:** Release team selected, schedule sent to kubeflow-discuss, all release team members have the proper permissions and are meeting regularly.
+**Success Criteria:**
+* Release team selected
+* Schedule sent to kubeflow-discuss
+* All release team members have the proper permissions and are meeting regularly
+* A tracking issue for the release has been created in kubeflow/community
 
 
 ### Development (10 weeks)
--- a/releases/handbook.md
+++ b/releases/handbook.md
@@ -199,6 +199,22 @@ Criteria for timeline that the team needs to consider
 * All release team members have the proper permissions and are meeting regularly
 * A tracking issue for the release has been created in kubeflow/community
 
+#### Release Team Selection
+
+The leads of the current release team will be selecting their successors for
+the next release. This should also take place before the current release team
+holds the release retrospective, in order for the new team to be able to
+provide feedback.
+
+The process for selecting shadows and members will be different though. In this
+case we want to welcome as many members as possible to join the release team.
+The release team should make an outreach via the available channels,
+mailing-list, slack, GitHub, and ask for volunteers that would like to
+participate in the release team.
+
+We currently don't have a limit on how many consecutive times someone can be a
+release manager, but it is hightly recommended that this role is rotated. This
+will allow us to further evolve our process and handbook.
 
 ### Development (10 weeks)
 
--- a/releases/handbook.md
+++ b/releases/handbook.md
@@ -205,7 +205,10 @@ Criteria for timeline that the team needs to consider
 Normal development in the different WGs and in the <https://github.com/kubeflow/manifests> repo.
 
 **Success Criteria:**
-* (Optional but encouraged): Issues tracking new features should also provide information on whether the docs should be updated for that feature
+* (Optional but encouraged): Issues tracking new features should also have a
+  corresponding docs PR that gets developed in parallel with the feature.
+  This will help keep track of PRs at risk during the Feature Freeze phase as
+  well as evaluate the docs effort early.
 
 
 ### Feature Freeze (2 weeks)
--- a/releases/handbook.md
+++ b/releases/handbook.md
@@ -155,26 +155,25 @@ The manifests repo will be following the release process below:
 
 
 ## Timeline
-
-| Week | Events |
-| --- | --- |
-| 1 | Development |
-| 2 |   |
-| 3 |   |
-| 4 |   |
-| 5 |   |
-| 6 |   |
-| 7 |   |
-| 8 |   |
-| 9 |   |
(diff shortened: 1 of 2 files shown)


releases/handbook.md
```diff
@@ -155,26 +155,25 @@ The manifests repo will be following the release process below:
...
-| 17 | Release |
+| **Week** | **Who** | **What** |
+| -------- | ------- | -------- |
+| Week 0 | Release Manager | [Preparation](#preparation) |
+| Week 1 | Release Manager | Start of Release Cycle |
+| Week 1 | Community | [Development](#development-10-weeks) |
+| Week 10 | Release Team | [Feature Freeze](#feature-freeze-2-weeks) |
+| Week 10 | Manifest WG | [rc.0 Released](#feature-freeze-2-weeks) |
+| Week 10 | Docs Lead | [Documentation](#documentation) |
+| Week 10 | Manifest WG | [rc.1 Released](#feature-freeze-2-weeks) |
```
annajung, line 166: According to the [handbook](https://github.com/kubeflow/community/blob/master/releases/handbook.md#feature-freeze-2-weeks), I believe rc.1 is released 2 weeks after the feature freeze? 

I think rc.1 should happen in week 12 since there is a schedule for rc.0 during week 10 already.

releases/handbook.md
```diff
@@ -194,15 +193,38 @@ Criteria for timeline that the team needs to consider
 - Kubecon dates - let’s not hard block on events, but keep them in mind since we know community members might get doublebooked.
 - Associated events (aka. AI Day at Kubecon, Tensorflow events) - we want to keep them in mind.
 
-**Success Criteria:** Release team selected, schedule sent to kubeflow-discuss, all release team members have the proper permissions and are meeting regularly.
+**Success Criteria:**
+* Release team selected
+* Schedule sent to kubeflow-discuss
```
annajung, line 198: Can we rephrase this to "send out release timeline proposal PR to kubeflow-discuss to receive feedback from the community"? and maybe during the week 1 phase, we have success criteria of "Send out the finalized release schedule to kubeflow-discuss"?

releases/handbook.md
```diff
@@ -194,15 +193,38 @@ Criteria for timeline that the team needs to consider
...
+* All release team members have the proper permissions and are meeting regularly
```
annajung, line 199: I think it would be helpful to list where the new release team members need to be added for permission. For example, the release GitHub team needs to be updated https://github.com/kubeflow/internal-acls/pull/511 and you mentioned a new member need to join kubeflow-discuss mailing group. Can we add those here and is there anything else that we can specify here?

releases/handbook.md
kimwnasptd: Nice catch! Yes you are right, it should happen on week 12 which is before the Manifests Testing starts.

I think I'll need to rephrase this sentence in the [Feature Freeze (handbook)](https://github.com/kubeflow/community/blob/master/releases/handbook.md#feature-freeze-2-weeks)
> On the last day of feature freeze, cut a new vX.Y-rc.1 RC tag on the vX.Y-branch release branch.

To make it clear that:
1. `Feature Freeze` does not stop at that time, it continues through the subsequent phases
2. We cut the RC1 because we are preparing for `Manifests Testing`, and not because `Feature Freeze` stops [which doesn't happen]. We want the Manifests leads to use an RC before they start their testing

kimwnasptd: Oh and also I think we'll need to fix the week 10 -> week 12 in this one as well
https://github.com/kubeflow/community/tree/master/releases/release-1.5#timeline

I'll push a commit with this as well

releases/handbook.md
kimwnasptd: I've mentioned that everyone needs to be part of the `internal-acl` repo, although it's a hyperlink, in the list of checkboxes for the [preparation phase](https://github.com/kubeflow/community/blob/master/releases/handbook.md#preparation).

I'll make it more explicit by adding sub-checks in that step to note that release team mambers need to be:
1. In the internal-acl repo and github team
2. Part of the `kubeflow-discuss` mailing list, in order to have access to the assets (drive, notes, recordings) of the release team

WDYT?

releases/handbook.md
kimwnasptd: Since there are the following checkboxes in the [preparation phase](https://github.com/kubeflow/community/blob/master/releases/handbook.md#preparation) above the success criteria:
> [ ] Publish draft schedule to kubeflow-discuss, with actual dates
> [ ] Get lazy consensus on the release schedule from the WG leads

WDYT about just replacing `Schedule sent to kubeflow-discuss` with your proposed `Send out the finalized release schedule to kubeflow-discuss` sentence there?

Currently the Preparation phase does not have a deadline, but from your comment about week 1 I understand you'd like to start defining this phase more concretely?

releases/handbook.md
annajung: That sounds good!

releases/handbook.md
annajung: Ah okay, if it's already in the preparation phase I think it's okay to leave it as it is. No need to rephrase. I wanted to make sure that the release timeline was proposed to the community for feedback before finalizing. :)

releases/handbook.md
annajung: again, totally missed the preparation phase. 

I think the only item missing is being part of the `kubeflow-discuss`. I think just adding that should be sufficient!

releases/handbook.md
annajung: I believe https://github.com/kubeflow/community/tree/master/releases/release-1.5#timeline will be fixed with https://github.com/kubeflow/community/pull/538.

PR #283, opened 2019-08-04
Put calendar events in code

scottilee:
This puts calendar events for the [Kubeflow calendar](https://calendar.google.com/calendar/embed?src=kubeflow.org_7l5vnbn8suj2se10sen81d9428%40group.calendar.google.com&ctz=America%2FLos_Angeles) in YAML so they can be defined and updated in one place.

The associated Python script parses the YAML and creates the events on a Google calendar.

This is in reference to https://github.com/kubeflow/community/issues/274.

@jlewi I need your help modifying this so it works with Prow or some CI to automatically update the calendar as well as associated things like hooking it up to the Kubeflow calendar. And of course, the code could be better so I'd appreciate comments on that as well.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/community/283)

PR #575, opened 2022-09-21
add `Kubeflow Community Call (Europe/APAC)`

thesuperzapper:
As discussed in the `Kubeflow Community Call (US West/APAC)` community meeting on 2022-09-20, due to low attendance, we will temporarily trial moving the `5:30pm San Fransisco` timeslot to `8:00am London` time.

This will result in the Kubeflow Community Meeting alternating weekly each Tuesday between `8:00am San Fransisco` and `8:00am London`.

__NOTE:__ the next call is still at `2022-09-27 @ 8:00am San Fransisco`, and the first meeting at the new time will be `2022-10-04 @ 8:00am London`

[kubeflow-discuss thread](https://groups.google.com/g/kubeflow-discuss/c/nxY_2FL1zes)

### DIFF
--- a/calendar/calendar.yaml
+++ b/calendar/calendar.yaml
@@ -6,7 +6,8 @@
 #   the list).
 #   name - The name of the meeting.
 #   date - The date of the first meeting (month/day/year).
-#   time - The start and end of the meeting in PST time zone.
+#   time - The start and end of the meeting.
+#   timezone - The timezone in IANA Time Zone Database format (default: 'America/Los_Angeles').
 #   frequency - How often the meeting takes place.
 #   until - Optional the last day to repeat until
 #   attendees: See: https://developers.google.com/calendar/v3/reference/events/insert#python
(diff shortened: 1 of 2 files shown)


thesuperzapper:
@Bobgy @jbottum I would appreciate a review/merge on this so it will update the calendar invites.

calendar/calendar.yaml
```diff
@@ -38,6 +41,29 @@
       Meeting Host Link: https://zoom.us/s/83583392870
   organizer: autobot@kubeflow.org
 
+- id: kf002
```
james-jwu, line 44: Can you increment this number instead (kf033 will be the next id)? I am not sure what will happen if you reuse an old event id.

thesuperzapper: @james-jwu I actually verified that an event with ID `kf002` does not exist on that calendar.

james-jwu: It existed is a previous commit: https://github.com/kubeflow/community/pull/283

Just out of caution can we use an unused event id?

thesuperzapper: Updated to `kf033` in https://github.com/kubeflow/community/pull/575/commits/49a21eb09838f5275fc44f1f962685b3f7498d27

PR #595, opened 2023-01-10
Add cncf-transition folder

james-jwu:
This folder will be used to host transition related files.


annajung:
Thanks James!
/lgtm

Issue #594, opened 2023-01-10
Add a doc for Kubeflow versioning policy


annajung:
Currently, there is no documentation on the versioning policy for each Kubeflow component and the Kubeflow release. Add a document consolidating all versioning policies from the WG to provide an overview of the versioning scheme used in the project.

cc @jbottum @james-jwu @theadactyl

james-jwu:
PR for creating cncf-transition folder: https://github.com/kubeflow/community/pull/595

PR #597, opened 2023-01-17
Add Kubeflow versioning doc

annajung:
Signed-off-by: Anna Jung (VMware) <antheaj@vmware.com>

fixes https://github.com/kubeflow/community/issues/594

- Add doc for Kubeflow versioning policy

cc interim steering committee @james-jwu @jbottum @theadactyl 
cc WG leads @kimwnasptd @thesuperzapper @zijianjoy @andreyvelich @johnugeorge

### DIFF
--- a/cncf-transition/versioning.md
+++ b/cncf-transition/versioning.md
@@ -1,0 +1,18 @@
+# Kubeflow Versioning Policy
+
+[Kubeflow Working Groups](https://github.com/kubeflow/community/blob/master/wg-list.md) (WG) follows
+ [semantic versioning](https://semver.org/) terminology expressed as X.Y.Z format, where X is the major version,
+ Y is the minor version, and Z is the patch version.
+
+For pre-release artifacts, `rc` and a number incrementing from 0 are appended to the end of the version to indicate
+ that it is a pre-release. For example, `X.Y.Z-rc.0`.
+
+In addition, some Kubeflow WGs leverage feature stages `alpha` and `beta` before releasing a stable version.
+ For example, `X.Y.Z-alpha.0`.
+
+For more details, see specification of each working group version policy and release process
+- [Katib](https://github.com/kubeflow/katib/tree/master/docs/release#release-process)
+- Manifests
+- Notebooks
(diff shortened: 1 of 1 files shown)


cncf-transition/versioning.md
```diff
@@ -0,0 +1,18 @@
...
+For pre-release artifacts, `rc` and a number incrementing from 0 are appended to the end of the version to indicate
+ that it is a pre-release. For example, `X.Y.Z-rc.0`.
+
+In addition, some Kubeflow WGs leverage feature stages `alpha` and `beta` before releasing a stable version.
+ For example, `X.Y.Z-alpha.0`.
+
+For more details, see specification of each working group version policy and release process
+- [Katib](https://github.com/kubeflow/katib/tree/master/docs/release#release-process)
+- Manifests
+- Notebooks
```
terrytangyuan, line 16: Are there policies for these as well?

annajung: I couldn't find any release process / version policies documented for these two WGs. Hoping @kimwnasptd can chime in if there is one available.

I decided to still list them here without a link to indicate that there is no specific policy set for these two WGs.

terrytangyuan: It looks like this list is for the ones with specific process though.

james-jwu: Should we file issues to Manifest and Notebooks WG to document their versioning policy?

annajung: Got it - I can understand the confusion. I'll go ahead and remove Manifest and Notebooks from the list for now to address @terrytangyuan 's concern and also file the issues in Manifest and Notebook to ensure their process is documented which addresses @james-jwu's comment. 

Once the docs are available for Manifest and Notebok, we can add them back to this list

jbottum:
thanks @annajung for moving this forward.  The versioning policies and formats for each WG are different, but I propose those differences are acceptable.   I agree that notebooks and manifests can be added later.    This 2nd commit looks like a good incremental step, which can be updated as the WG make changes/updates to their policies, so I think this looks good to me.

terrytangyuan (approved):
/lgtm

Issue #602, opened 2023-01-31
The Invitation to Kubeflow workspace in the Readme is no longer available


Syulin7:
Can someone help update it?

https://github.com/kubeflow/community/blob/13d18b3f8d84267f225edf1af2e4743a16a94caf/slack/README.md?plain=1#L15

thesuperzapper:
I recommend we update it with a link to the kubeflow website, which will always have a working link.

https://www.kubeflow.org/docs/about/community/#kubeflow-slack

thesuperzapper:
@Syulin7 I recommend we update it with a link to the kubeflow website, which will always have a working link.

https://www.kubeflow.org/docs/about/community/#kubeflow-slack

thesuperzapper:
@james-jwu will need to approve since its in the root of the repo.

PR #604, opened 2023-02-01
Update join Slack link to the non expiring link

mameshini:


jbottum:
/lgtm

thanks!

thesuperzapper:
@mameshini @jbottum can we please update it to `https://www.kubeflow.org/docs/about/community/#kubeflow-slack` (rather than using the invited link directly).

As I was saying in https://github.com/kubeflow/community/issues/602#issuecomment-1411298281, if we change how we do Slack inviting in the future, we will forget about this link otherwise.

Syulin7:
> @mameshini @jbottum can we please update it to `https://www.kubeflow.org/docs/about/community/#kubeflow-slack` (rather than using the invited link directly).
> 
> As I was saying in [#602 (comment)](https://github.com/kubeflow/community/issues/602#issuecomment-1411298281), if we change how we do Slack inviting in the future, we will forget about this link otherwise.

Agree. I would like to update it.

mameshini:
Updated and merged [#605] into master

PR #606, opened 2023-02-02
Update the link for Slack signup page

mameshini:
Update the link for Slack signup page


mameshini:
This is already addressed

PR #631, opened 2023-06-12
delete PM meeting from calendar.yaml

jbottum:
Deleting Product Management team from Kubeflow calendar as we are not using it.


jbottum:
@theadactyl would you please approve this ?

thesuperzapper:
I don't know why this was ever needed, because that event has an `until` date, so is not active on the calendar.

/close

PR #629, opened 2023-05-21
Add Manifests WG call to common calendar

kimwnasptd:
This PR introduces the following changes:
1. Add the Manifests WG meeting in the common Kubeflow calendar
    * The meeting links, recording and drive are maintained by Manifests WG
    * The meeting will now be happening on a bi-weekly basis, rather than monthly
    * The meeting day will be Thursday instead of Tuesday
2. Remove the Model Management SIG call which conflicts with manifests 
    * We've given a notice 2 months ago but got no response #619 
    * The last meeting of this SIG was in 7 Feb 2021

cc @kubeflow/wg-manifests-leads @annajung @juliusvonkohout 
/assign @james-jwu

PR #634, opened 2023-06-22
Add Notebooks WG to community calendar

kimwnasptd:
Follow-up from https://github.com/kubeflow/community/pull/629#issuecomment-1602075037

@james-jwu I've fixed the two problematic entries:
* Release team meeting also had the `(CET)` in it, cc @DnPlas 
* Manifests WG meeting had `(EET)

And explicitly added the `timezone` field. 

Then I also added an entry for the Notebooks WG recurring call. After that we should be able to run the script as expected

/cc @james-jwu


kimwnasptd:
The calendar is up to date and I see the meetings. Thanks for the help @james-jwu!

james-jwu:
You are welcome!

PR #664, opened 2023-11-29
Update the kubeflow-steering-committee-election-proposal.md

akgraner:
Made the following changes 

Updated the tool we are going to use to CIVS which is the tool Elekto is based upon, but it's easier to administer. 

Updated the timeline to take account for the October and November industry events that many of the active community members participate in, the holiday season that takes place as well as vacation time during November, December and January.  

Added a testimonial phase to the elections to give community members the opportunity to endorse nominees ahead to the election date. 

This should allow us to move this process from a proposal to process and then we can kick off the election process and open nominations.


james-jwu:
Thanks! 
/lgtm
/approve

PR #667, opened 2023-12-02
Create eligible-candidates-and-voters-2023-ksc.md

akgraner:
added the eligible candidates and voters file so we can add the names on 4 December before the exception phase begins.


Davidnet:
I will to this merge the list first thing morning 4th

PR #715, opened 2024-04-24
Migrate community call to CNCF Zoom

thesuperzapper:
This PR migrates from the community calls from my personal Zoom account to the CNCF Zoom account.

We share the host password for this new account using OnePassword.

We need to warn people that the link to join has changed.

### DIFF
--- a/calendar/calendar.yaml
+++ b/calendar/calendar.yaml
@@ -25,25 +25,25 @@
 # Current meetings
 ####################################################################################################
 - id: kf032
-  name: Kubeflow Community Call (US East/EMEA)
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@jbottum @kubeflow/kubeflow-steering-committee, this will update the community call to use the new CNCF Zoom account.

jbottum:
@thesuperzapper thanks for pushing this along.   we need to get this done even if the 1st week is bumpy.   I am +1 on this but would like other steering committee members to ack

terrytangyuan (approved):
/lgtm

PR #725, opened 2024-06-18
Merge security and manifests WG meeting in a cal entry

rimolive:


juliusvonkohout:
CC @kimwnasptd @akgraner for review

juliusvonkohout:
@kimwnasptd are you alright with the organizer change and should we switch to the new Zoom as well?

thesuperzapper (changes requested):
As it says at the top of this file, never remove a recurring meeting from this file otherwise it will end up in the calendar forever. 

You have to move it to the legacy section and set its end date in the past.

rimolive:
> As it says at the top of this file, never remove a recurring meeting from this file otherwise it will end up in the calendar forever.
> 
> You have to move it to the legacy section and set its end date in the past.

My apologies to that. I made the changes as requested.

rimolive:
cc @kubeflow/kubeflow-steering-committee @juliusvonkohout @kimwnasptd

juliusvonkohout:
/lgtm

I will probably add the new zoom bridge in a follow up PR.

thesuperzapper:
I just saw that the proposed meeting is named `Kubeflow Platform Meeting (Manifests + Security WG)`.

It's hard to imagine that "Kubeflow Platform" should not also involve the distribution owners? I am not sure we should be calling the meetings "Kubeflow Platform" if they are only referring to Manifests/Security WGs.

juliusvonkohout:
> I just saw that the proposed meeting is named `Kubeflow Platform Meeting (Manifests + Security WG)`.
> 
> It's hard to imagine that "Kubeflow Platform" should not also involve the distribution owners? I am not sure we should be calling the meetings "Kubeflow Platform" if they are only referring to Manifests/Security WGs.

Then lets ask @andreyvelich @terrytangyuan @jbottum what they think. Based on what they decide we could change it.

andreyvelich:
From my perspective, this meeting should be the main place for users if they have questions related to Kubeflow Platform components. For example, questions about Profile Controller, KFAM, Central Dashboard, etc.

Distributions should certainly join these calls since they are main consumers of Kubeflow Manifests.

juliusvonkohout:
> From my perspective, this meeting should be the main place for users if they have questions related to Kubeflow Platform components. For example, questions about Profile Controller, KFAM, Central Dashboard, etc.
> 
> Distributions should certainly join these calls since they are main consumers of Kubeflow Manifests.

Can you approve it then or do we have to make changes?

andreyvelich (commented):
/lgtm
/assign @johnugeorge @jbottum @terrytangyuan @james-jwu

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #731, opened 2024-07-04
Move the Pipelines WG PST PM meeting to PST AM

rimolive:
As discussed in the Pipelines WG meeting, we will be meeting every 2 weeks only on PST AM.

cc @kubeflow/pipelines

### DIFF
--- a/calendar/calendar.yaml
+++ b/calendar/calendar.yaml
@@ -46,26 +46,11 @@
       Meeting Host Link: https://zoom.us/s/99152427566
   organizer: autobot@kubeflow.org
 
-- id: kf025
-  name: Kubeflow Pipelines Community Meeting (PST PM)
-  date: 08/05/2020
-  time: 5:30pm-6:10pm
-  frequency: every-4-weeks
-  video: https://meet.google.com/jkr-dupp-wwm
-  attendees:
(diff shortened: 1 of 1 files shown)


calendar/calendar.yaml
```diff
@@ -509,4 +494,20 @@
...
+  video: https://meet.google.com/jkr-dupp-wwm
+  until: 07/03/2024
+  attendees:
+    - email: kubeflow-discuss@googlegroups.com
+    - email: kubeflow-pipelines@google.com
+  description:
+    - |
+      Notes & Agenda: http://bit.ly/kfp-meeting-notes
+      Join at: https://meet.google.com/jkr-dupp-wwm
+  organizer: chensun
```
zijianjoy, line 513: Can we remove this meeting to avoid duplication?

rimolive: This is part of instructions to remove recurring meetings. We cannot just remove it from the file, but move to the bottom of the file and set a `until` field with a date in the past. If we don't do that, the calendar entry will never be removed.

zijianjoy: I see, thank you for the reminder!

zijianjoy:
/lgtm
/approve

Thank you for this PR!

PR #744, opened 2024-08-28
docs: expand on DCO using manual sign-off

tarilabs:
@hbelmiro wdyt?
I had this gotchas when DCO didn't pass despite I've used the hook (in a project, the Name was different).

I'm also not sure the email of commit "must" match the github primary email, but that's for another time.

### DIFF
--- a/dco-signoff-hook/README.md
+++ b/dco-signoff-hook/README.md
@@ -5,7 +5,7 @@ You can also configure it globally (for every repo on your machine) by copying t
 
 You can also sign off your contributions manually by doing ONE of the following:
 * Use `git commit -s ...` with each commit to add the sign-off or
-* Manually add a `Signed-off-by: Your Name <your.email@example.com>` to each commit message.
+* Manually add a `Signed-off-by: Your Name <your.email@example.com>` to each commit message; please note that Name and Email of sign-off must match the commit's Author, in other words the `user.name` and `user.email` of the git configuration used when creating the commit. Using `git commit -s ...` does this automatically.
 
 The email address must match your primary GitHub email. You do NOT need cryptographic (e.g. gpg) signing.
 * Use `git commit -s --amend ...` to add a sign-off to the latest commit, if you forgot.


ederign (approved):

tarilabs:
bump @terrytangyuan @andreyvelich could this be merged, please? (I'm going through some backlog of pending PRs...) thanks

dco-signoff-hook/README.md
```diff
@@ -5,7 +5,7 @@ You can also configure it globally (for every repo on your machine) by copying t
 
 You can also sign off your contributions manually by doing ONE of the following:
 * Use `git commit -s ...` with each commit to add the sign-off or
-* Manually add a `Signed-off-by: Your Name <your.email@example.com>` to each commit message.
+* Manually add a `Signed-off-by: Your Name <your.email@example.com>` to each commit message; please note that Name and Email of sign-off must match the commit's Author, in other words the `user.name` and `user.email` of the git configuration used when creating the commit. Using `git commit -s ...` does this automatically.
```
terrytangyuan, line 8: I am not sure if this is too verbose here for an alternative step. We should recommend the automatic way instead of the manual way.

tarilabs: thank @terrytangyuan , the automatic way is listed just above this bulletpoint.

this is imho nice "gotchas" for those situation where you have to resolve something manually; it happen a couple of times to me, and in the first occurence I lost time digging for this information, so my hope is this will spare some time for others facing potentially same issues when those rare occurences of having to manually signing off

andreyvelich:
Thank you for this PR @tarilabs!
Can we move this doc to the Kubeflow website: https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits ?
It would be nice to unify docs for our users and reduce number of places they need to research.
 cc @kubeflow/kubeflow-steering-committee

dco-signoff-hook/README.md
terrytangyuan: Were there any error messages when the name/email mismatches? I wonder if we should defer this to the DCO check itself since I'd imagine there are many other gotchas. This one talks about the name but there can be a mismatch in primary email as well

terrytangyuan:
> Can we move this doc to the Kubeflow website: https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits ?

The website currently already links back to this doc which has more details.

andreyvelich:
> > Can we move this doc to the Kubeflow website: https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits ?
> 
> The website currently already links back to this doc which has more details.

But this doc has very little information. @terrytangyuan Is there any problem to just move it in Kubeflow website ?

dco-signoff-hook/README.md
tarilabs: > This one talks about the name but there can be a mismatch in primary email as well

this comment actually refers to the commit Author, as a whole, making reference to the Name and Email explicitly. 


> I wonder if we should defer this to the DCO check itself since I'd imagine there are many other gotchas.

I've added this comment after some little frustration because I didn't find this information easily, and my hope would have been to spare others of the same.

If we think however this is resulting in the doc instead being too verbose, we can close this PR.

terrytangyuan: Ok if DCO check doesn't provide much information then this would be useful to have. Thanks for clarifying.

terrytangyuan:
This file has 13 lines of instructions, which could take 2/3 of the page or so (and that section is at the top of the page on the website). If contributors are already familiar with DCO, they don't need the details. 

There are also instructions in this README that are specific to the [commit hook](https://github.com/kubeflow/community/blob/master/dco-signoff-hook/prepare-commit-msg) so I think it would be better for code and instruction to stay close to each other.

andreyvelich:
@terrytangyuan Why it will be the 2/3 page ?
I can see that this file has 316 lines already: https://github.com/kubeflow/website/blob/master/content/en/docs/about/contributing.md?plain=1.

We can always create dedicated page in the website if we feel that this is additional information that users should read.

I just think that having unify place for docs will be easier for us to maintain, and easier for users to search.

Any thoughts @StefanoFioravanzo @hbelmiro ?

terrytangyuan:
> Why it will be the 2/3 page

I meant that if we move these lines to that page, it would take up 2/3 of my screen before I can read the next sub-section "Follow the code of conduct".

tarilabs:
I believe this would have been meaningful addition

PR #752, opened 2024-09-16
Add KSC contacts and remove delegated authority section

terrytangyuan:
@kubeflow/kubeflow-steering-committee


terrytangyuan:
Yeah we don't have plan for those. We can add later if needed.

PR #785, opened 2024-11-04
Introduce rule to allow 1 vote only from the same organization

james-jwu:
This proposal was discussed together with #786 (1 seat per organization) in KSC, where majority felt the 2 rules will encourage diversity of the KSC committee.

2 members from the same organization on KSC may give that organization the power to veto any governance change proposals, under the 70% rule for passing such proposals. In an ideal world, each KSC member should act and vote individually and according to the best interest of the Kubeflow project. However, KSC recognize that an organization may have strong influence their representatives in KSC. This rule is designed to prevent one organization from being able to reject proposals.

In the case where KSC is made up of members from 3 organizations (i.e. organizations A, A, B, B, C), then the 3 votes will need to be unanimous to pass a proposal.

In the case where 2 members from the same organization have different opinions, this rule does not provide any guidance on resolving the conflict.

If #786 is passed, this rule will not have any effect to the 2026 KSC and thereafter, since the KSC will have representation from 5 different organizations. This rule may still be kept to in case new proposal revert #786.

### DIFF
--- a/KUBEFLOW-STEERING-COMMITTEE.md
+++ b/KUBEFLOW-STEERING-COMMITTEE.md
@@ -72,8 +72,11 @@ Members of KSC may abstain from a vote. Abstaining members will only be consider
 Issues that impacts the KSC governance requires a special decision process. Issues include:
 - Changes to the KSC charter
 - KSC voting rules
+- Election rules
 
 The issue may pass with 70% of the members (rounded up) of the committee supporting it.
+
+At most 1 KSC member from the same organization may cast a vote. Example: if KSC is made up of employees from organizations A, A, B, C, D, then orgnaization A may cast 1 vote through either employee. There would be a total of 4 votes, and 3 votes (70% rounded up) is required to pass a proposal. This rule is designed to remove organization A's ability to defeat a proposal that is supported by all other KSC members.
 
 ### Results
 
--- a/KUBEFLOW-STEERING-COMMITTEE.md
+++ b/KUBEFLOW-STEERING-COMMITTEE.md
@@ -76,7 +76,9 @@ Issues that impacts the KSC governance requires a special decision process. Issu
 
 The issue may pass with 70% of the members (rounded up) of the committee supporting it.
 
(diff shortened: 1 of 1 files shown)


thesuperzapper:
I personally don't think this rule would be of benefit to the community.

My two major concerns are:

1. If the company makeup of the KSC was `A A B B C`, achieving 70% would require a unanimous vote because 2/3 == 67% (rounded up)
2. The members of the KSC are individuals, and are not strictly there to represent a specific company.
    - For example, two people from the same company might disagree on some rule and want to vote differently.
    - In cases like this, how would you even allocate that company's vote? There is no higher body to mediate this.

I also think that this rule slightly misses the point of the 70% special vote. The intention was to make it more difficult to make changes to governance rules.

That is, it's a feature, not a bug, that it requires 4/5 KSC members to agree on changes. This allows any two people to block such changes (by either voting against or just abstaining).

jbottum:
I am +1 as is.   Note - if others would like the two KSC members from the same company to have their vote count as 0.5 on special votes, I would support that refinement.

thesuperzapper:
Also, if we choose to adopt this, we need to address how exactly these "special votes" are run, so that there is no ambiguity on what to do when there is a disagreement between members of the same org.

My proposal is that we still have all members cast an explicit vote, but down weighted by the proportion of seats that company has.

E.g. if the company has 2 seats, their members get 0.5, but if they have 1 seat, they get 1.0.

james-jwu:
Have KSC members from the same company weighted and aggregates to 1 (i.e. 0.5 in case of 2 members) is a good idea. +1 to that.

terrytangyuan:
+1 to both editions

terrytangyuan (approved):

james-jwu:
I amended the proposal by introducing weighted votes from the same company.

andreyvelich (commented):
Thank you for this @james-jwu!
+1 for this rule

johnugeorge:
+1

andreyvelich:
Thanks everyone for your feedback 🎉 

/lgtm

PR #816, opened 2025-02-06
chore: add trademark disclaimer

pboyd:
According to [kubeflow/website #3664](https://github.com/kubeflow/website/issues/3664) and [clomonitor.io](https://clomonitor.io/projects/cncf/kubeflow#community) this readme should have a trademark disclaimer.

PR #817, opened 2025-02-07
chore: revert legal disclaimer

pboyd:
PR #816 did not have the intended effect, so it should be reverted.

Signed-off-by: Paul Boyd <pboyd@redhat.com>


varodrig:
@pboyd  agree that needs to be fixed and updated. but why to  removed the source code instead of update it with the fix?
can you add more clarifications?

pboyd:
Hi @varodrig, thanks for the reply. The correct fix isn't in the code, so I can't submit a PR for it. The website URL in the repo metadata needs to point to www.kubeflow.org, I don't have access to update it myself, but there are more details in #818. It should only a take a second, if you have access.

PR #887, opened 2025-07-29
tenzen-y steps down from WG AutoML leadership

tenzen-y:
REF: https://github.com/kubeflow/katib/pull/2561

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -10,7 +10,6 @@ aliases:
     - andreyvelich
     - gaocegege
     - johnugeorge
-    - tenzen-y
   wg-deployment-leads:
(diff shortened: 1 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -10,7 +10,6 @@ aliases:
     - andreyvelich
     - gaocegege
     - johnugeorge
-    - tenzen-y
```
andreyvelich, line 13: You might need to remove yourself from [the `wgs.yaml`,](https://github.com/kubeflow/community/blob/master/wgs.yaml#L110) and run `make generate`

tenzen-y: done

Issue #809, opened 2025-01-29
GSOC 2025 - Mentor Project Planning


rareddy:
Hello Kubeflow Community, 

Kubeflow has officially submitted to be part of Goole Summer Code 2025. We do hope to get selected as one of the participating organizations this year. We had a very successful GSOC 2024. I thank all the Mentors and Students for making it successful.

I am looking for

- Signing up Mentors. If you would like to be a mentor this season, please reach out to me on CNCF Slack or leave your Name and git ID here. I will invite all the 2024 mentors by default, so please use this form only if you are a new mentor.

- Project Ideas for Kubeflow components. If you submit an idea, you must be a mentor (Kubeflow member and/or committer) or if you are a student submitting a proposal it must be approved by one of the component leads or a committer on that component.

Make sure your Project Ideas include ALL of the required information below:

1. Project title/description
2. Detailed description of the project. Provide as many details as possible for students to read and understand without asking too many questions about the project.
3. Expected outcomes
4. Skills required/preferred
5. Possible mentors
6. Component
7. Difficulty  (Low, Medium, Expert)
8. If you already have Git issue for it, please include that here.
9. Expected size of the project (90, 175 or 350 hours. Please do NOT use any arbitrary number, choose one close to list given)

Leave a single comment here for each proposal. As we approve them, we will collate all the projects at https://www.kubeflow.org/events/gsoc-2025/. If you have questions, please come and ask in CNCF Slack #kubeflow-gsoc-participants.

**All mentors MUST be assigned to one or more projects.** 

The deadline to submit all the ideas is **Feb 15th, 2025**.  So, please take some time and submit them. Thank you all for your support.

Shekharrajak:
@rareddy , I would like to be a mentor and interested on [spark-operator](https://github.com/kubeflow/spark-operator) project. I will be sharing my ideas/project proposals soon.

Electronic-Waste:
@rareddy Hi, I was a GSoC student last year, and also a new maintainer in Kubeflow. Benefit from the GSoC project a lot, I would like to apply for the mentor of [training-operator](https://github.com/kubeflow/training-operator) and [katib](https://github.com/kubeflow/katib) this year. I'm sure it will help more new contributors like me to get engaged in Kubeflow and  the world of open source.

As for the project ideas, not yet. But I'll put them here once they are ready.

cc @kubeflow/wg-automl-leads @kubeflow/wg-training-leads

juliusvonkohout:
[...]

Issue #833, opened 2025-03-10
Add more adopters and list them on the https://www.kubeflow.org/ front page


juliusvonkohout:
@jbottum @akgraner @chasecadet @kubeflow/kubeflow-steering-committee @kubeflow/release-team do you know other companies we can list to later on expose this on the main entrypoint of the website ?
From Kubeflow talks I remeber PepsiCo, Cern, Roche Pharmaceutical (@kromanow94 @MaxKavun ) and a few other high-profile companies.

Also Redhat @rimolive @varodrig and canonical @kimwnasptd @DnPlas  could be added.

"As part of this initiative, please could someone drive the discussion to update https://www.kubeflow.org/ front page with companies using Kubeflow projects?
We can start with opening an issue in kubeflow/website and find UI/UX engineers to propose something."

- [x] Telia
- [ ] DHL https://github.com/kubeflow/community/pull/828
- [ ] Canonical
- [ ] RedHat
- [ ] PepsiCo
- [ ] Cern
- [ ] Roche
- [ ] Microsoft
- [ ] AWS
- [ ] Google

juliusvonkohout:
> There are a lot of adopters for individual components, e.g. trainer: https://github.com/kubeflow/trainer/blob/master/ADOPTERS.md

Yes, this should also be used.

thesuperzapper:
The `kubeflow/community` adopters file has been updates with links to the component ADOPTERS files:
- https://github.com/kubeflow/community/blob/master/ADOPTERS.md

But we currently lack ADOPTERS files to link for the following components:
- Kubeflow Model Registry
- Kubeflow Notebooks
- Kubeflow Pipelines

I imagine that many orgs will want to list themselves in the specific components they use, we should get adopters files set up for those components and update the links on the `kubeflow/community` one.

mahdikhashan:
also dkube: https://www.linkedin.com/products/one-convergence-devices-dkube/

akeed:
Wrong name in Telecommunications - Telia:

Telia, Sources: [GitHub ADOPTERS.md](https://github.com/kubeflow/community/blob/master/ADOPTERS.md) (Official adopter list), TODO Kubecon 2024, Contacts: **Akeed Kadir**i, Sarah Berenji, Tarek Abouzeid

should be

[...]

PR #891, opened 2025-07-31
KEP-2676: Support Istio ambient in Kubeflow

orfeas-k:
Add proposal for supporting Istio ambient in Kubeflow. 

## Context
This has been discussed for more than a year in the Kubeflow community (see kubeflow/manifests#2676). Currently, there is an ongoing GSOC project working on implementing this (see
https://github.com/kubeflow/community/issues/809#issuecomment-2633370759), which will be greatly facilitated by agreeing on expected changes.

## What is this proposal about?
This proposal:
* defines the changes needed to make Kubeflow components ambient-compatible
* outlines a setup that uses Istio in ambient and enables all the existing Authorization and Authentication use cases.

Ref kubeflow/manifests#2676

cc @kimwnasptd @juliusvonkohout @thesuperzapper


proposals/2676-istio-ambient.md/README.md
```diff
@@ -0,0 +1,217 @@
...
+    - Aligns Kubeflow more with the official, vendor-neutral direction of Kubernetes networking.
+
+You can read more about Ambient's benefits at istio.io[^1] and in their performance analysis[^6].
+
+### Goals
+This KEP aims to define:
+* A high-level architecture of Kubeflow that uses Istio ambient and showcases how to:
+  * Use HTTPRoutes instead of VirtualServices
+  * Enable the user login flow
+  * Use waypoints to enforce namespace isolation (L7 and L4 traffic)
```
juliusvonkohout, line 31:
```suggestion
  * Use waypoints to enforce namespace isolation (L7 and L4 traffic)
  * Provide Zero overhead namespaces, so without actual user workloads the Pods, CPU and memory used should be 0 (after the seaweedfs PR is merged). So no waypoint proxies are allowed in kubeflow profile namespaces.
```

proposals/2676-istio-ambient.md/README.md
```diff
@@ -0,0 +1,217 @@
...
+#### 2. Ingress traffic
+In sidecar mode, Kubeflow deploys one Istio gateway deployment for serving ingress traffic but multiple Istio gateways that attach to it. In ambient, one k8s gateway will be created. It is configured to serve traffic for all namespaces and is of `gatewayClassName: istio`. The spawned deployment from that k8s gateway is the component that actually serves the traffic. 
+
+For routing, the migrated HTTPRoutes are attached to the single ingress k8s gateway created.
+
+#### 3. Securing the `kubeflow` namespace
+To enable authorization for Kubeflow components, the whole `kubeflow` namespace is included in the mesh. 
+
+##### Waypoints
+To get the best performance out of ambient mode[^6][^7], services are enrolled to a waypoint only when the L7 features are required. For the Kubeflow control plane, only authorizationPolicies targeting `ml-pipeline` and `ml-pipeline-ui` make use of L7 attributes. Thus, a waypoint is deployed in the `kubeflow` namespace while only the services of those two components are enrolled to the waypoint.
```
juliusvonkohout, line 82: Are you sure that notebooks do not use L7 ? I mean everyone with the kubeflow-userid header uses L7 so i guess almost all components in Kubeflow.

proposals/2676-istio-ambient.md/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### 4. Securing the user namespaces
+Similarly to the `kubeflow` namespace, all user namespaces are included in the mesh in order to enable authorization enforcement, while sidecars are disabled.
+
+##### One waypoint per namespace
+User namespaces are secured via an AuthorizationPolicy that is created dynamically by kubeflow's `profile-controller`. Most of the rules in this AuthorizationPolicy have L7 features[^8]. In sidecar mode, this AuthorizationPolicy is enforced to all traffic in the namespace by having no `selector`. This is due to the nature of those namespaces, where services are created dynamically. In Istio ambient mode, the only way to enforce a namespace-wide L7 AuthorizationPolicy is by proxying traffic to the whole namespace through the waypoint and then attaching the AuthorizationPolicy to the waypoint. This results in the authorization rules being enforced for all traffic proxied by the waypoint. This means that **a separate waypoint needs to be deployed in every user namespace**.
```
juliusvonkohout: No, we want to have exactly the opposite. Never ever waypoint proxies in user namespaces. Also the istio documentation and my talk with @kimwnasptd  and the istio/solo.io founder at Kubecon supports that you do not need them in user namespaces. See https://istio.io/latest/docs/ambient/usage/waypoint/#configure-resources-to-use-a-cross-namespace-waypoint-proxy. That is the most essential part that you do not have waypoint proxies there and do not destroy zero overhead user  namespaces.

proposals/2676-istio-ambient.md/README.md
```diff
@@ -0,0 +1,217 @@
...
+>Typically, the authorization policy would be applied to a service so the destination would be "the thing my policy is attached to".
+>...
+>I don't think there is a shorthand expression for "all services in X namespace".
+
+This leaves no other way to enforce an L7 AuthorizationPolicy to all services in a namespace, where services are being created dynamically.
+
+Note that the introduced resource overhead affects cases with a large number of **idle** namespaces. For namespaces that are being actively used, the resource footprint is still greatly reduced, given that there are no sidecar containers injected.
```
juliusvonkohout, line 113: This sounds really problematic. So i would rather stay with Istio-cni until this is possible. Follow up of https://github.com/kubeflow/community/pull/891/files#r2248473959

proposals/2676-istio-ambient.md/README.md
orfeas-k: [Yes](https://github.com/kubeflow/manifests/blob/master/applications/jupyter/jupyter-web-app/upstream/overlays/istio/authorization-policy.yaml#L7-L11). All those components trust the ingress gateway and the headers it provides. Thus, they accept all requests from the ingress gateway without any filtering done at the AuthorizationPolicy, like kfp components do ([api server](https://github.com/kubeflow/manifests/blob/master/applications/pipeline/upstream/base/installs/multi-user/istio-authorization-config.yaml#L47-L49))

proposals/2676-istio-ambient.md/README.md
orfeas-k: I don't think we should block on this. Of course, providing zero overhead namespaces without any namespaced resources should be the north star we 're after. The fact though that Istio ambient does not enable us to do so should not be a reason for Kubeflow to not be ambient compatible. Both istio-cni and ambient mode have their pros and cons and Kubeflow should provide administrators the option to decide on their own which mode suits them. For example, clusters with few namespaces but hundreds of workloads per namespace would benefit greatly from using ambient, contrary to clusters with hundreds of idle namespaces and a few workloads per namespace.

As mentioned in the proposal, instead of blocking, I suggest:
1. Implement code that does not provide enforce any high-level architecture. That means that it will be configurable enough to support both waypoint per namespace and cross-namespace waypoint architectures. If Istio ambient starts supporting L7 namespace isolation using a single cross-namespace waypoint, then the code is already compatible or at minimum needs a few tweaks rather than the whole "sidecar -> istio ambient & gateway API" migration effort.
2. Provide an `ambient` overlay in manifests. The Istio-cni will still be the default one but providing an overlay enables users and distributions to try ambient and experience benefits and downsides. This can greatly help the next point. 
3. Ask feature from Istio to support L7 namespace isolation using a single cross-namespace waypoint. I 've already asked them in Slack and plan to bring this up in their meetings too. However, I suspect that they have a good reason this is not supported right now. If there is adoption of Istio ambient by kubeflow users, then this puts us in a better position by providing valid use cases for it.

proposals/2676-istio-ambient.md/README.md
orfeas-k: (replied in https://github.com/kubeflow/community/pull/891#discussion_r2269208373)

proposals/2676-istio-ambient.md/README.md
orfeas-k: This is also related to https://github.com/kubeflow/community/pull/891#discussion_r2269208373.

proposals/2676-istio-ambient.md/README.md
kimwnasptd: Agree with all of @orfeas-k's points.

While having waypoints per namespace will have a different resource footprint from CNI, still this is up to the vendors to decide if they are OK with or not rather than not supporting it in Kubeflow.

The example can keep the CNI, but the project should still give the option to end users to have Kubeflow with K8s Gateway API and ambient.

proposals/2676-istio-ambient.md/README.md
juliusvonkohout: I was talking about the jupyterlab itself in the customer namespace. They should be created on demand based on the notebook custom resource.

proposals/2676-istio-ambient.md/README.md
juliusvonkohout: Yes,  i agree, lets not make istio ambient the default until this is possible. So for the time being istio-cni is the default and istio-ambient opt-in overlay of istio-cni.

proposals/2676-istio-ambient.md/README.md
orfeas-k: Do you refer to the AuthorizationPolicy created by the `profile-controller`? This is created dynamically when a new user namespace is created and it also allows/denies traffic to jupyterlab server pods. This looks like [this](https://github.com/orfeas-k/kubeflow-istio-ambient-demo/tree/main/use-case-c-programmatic-access-using-k8s-token#namespace-wide-l7-authorization), which indeed has L7 rules.

proposals/2676-istio-ambient.md/README.md
orfeas-k: I **think** technically they 'll need to be two separate overlays but let's keep this implementation detail for implementation time. We all agree that Istio-cni will be the default one, provided in the example.

juliusvonkohout: I just want to keep a single istio folder. In there we also have the legacy insecure overlay to disable istio-cni. https://github.com/kubeflow/manifests/tree/master/common/istio/istio-install/overlays/insecure. please lets not create multiple istio versions to maintain, but one base version with overlays.

orfeas-k: This was brought up in the Istio community meeting by @ca-scribner and it looks like they don't see any blockers to this. It was not an intentional omission, just something that hasn't been implemented or decided on. There was nothing brought up that was a hard blocker, more just challenges to overcome. It would though be a big effort, so before they actually action it they'd want to see more than a single person requesting the feature.

I filed a feature request here https://github.com/istio/istio.io/issues/16781. What we need to do now is to:
1. get more consensus/demand, which can only be done by getting more people to use the ambient overlay (at least from Kubeflow's side)
2. bring it up again in a future community meeting once the demand is obvious

I 'm updating the proposal with this information to make sure it is documented.

proposals/2676-istio-ambient.md/README.md
orfeas-k: I think this is not needed since it's not an actual goal of this proposal.

monotek:
> Thank you @monotek. Do you mind adding Staffabse to the Kubeflow Adopters ? https://github.com/kubeflow/community/blob/cb085e4311e48b906cec9c6c1653bb94928cbf9b/ADOPTERS.md? See also #833


We're still evaluating kubeflow. Not sure yet if we'll keep using it.


> @monotek we have just merged ambient here [kubeflow/manifests#3246](https://github.com/kubeflow/manifests/pull/3246). Maybe you want to evaluate it. CC @madmecodes

That's likely not enough because the notebook controller should be able to set waypoint labels to the notebook service too.

christian-heusel:
@kubeflow/kubeflow-steering-committee this KEP got closed by the stale bot even though the people that raised this are still interested in driving this forward (we just discussed this in the Notebooks WG Call due to a proposed change in the dashboard https://github.com/kubeflow/dashboard/pull/267). Could you clarify on the status of the proposal and make sure that it is being properly shepherded? 🤗 

cc @deusebio

deusebio:
Yes, as I mentioned yesterday during the call, we have definitive interest seeing this feature consolidated upstream, since in the past year we've kept on working, re-opening pull-requests, rebasing and addressing comments in the various PRs that we have opened, naively hoping this could have happened for 26.03 already. The work we have done goes well beyond just the PRs opened there, since we have also already integrated those changes into our solution as an experimental feature to test and validate that the integration does indeed work as expected. 

If anyone want to give it a go, you can already either: 
1. [deploy Charmed Kubeflow](https://documentation.ubuntu.com/charmed-kubeflow/latest/how-to/install/general-installation/), also following the guidance [here](https://documentation.ubuntu.com/charmed-kubeflow/latest/how-to/manage/migrate-ambient/) on how to migrate/enable ambient
2. Or you can also install Charmed Kubeflow with ambient using [this terraform module](https://github.com/canonical/charmed-kubeflow-solutions/tree/main/modules/kubeflow-ambient) and following the guidance [here](https://documentation.ubuntu.com/charmed-kubeflow/latest/how-to/install/install-terraform/). 

This should already provide an easy path to anyone that is also interest on this topic and they would like to give it a go to get started. 

We *strongly* believe this would be a very sensible feature that we hope it can be of broader interest for the community. Beside providing support to ambient (that is more efficient and easy to manage than side-car), this also brings the integration with the Gateway API, that would also make a step forward to make Kubeflow more general and compliant with the new way to manage routing in K8s. But having said this (which are our considerations why we believe this is a useful feature), at this point I'd really like to have clearer indication whether there is indeed interests and commitment from upstream to get this in. I really wouldn't like to keep on working on this any further to then figure out that the feature was not of interest or will never make it to a release.

monotek:
Please not.
Managing k8s resources with Terraform almost certainly lead to configuration drift.

Why can't we just have plain k8s resources or a Helm chart, which we can use in a real gitops tool like FluxCD, independent from a service mesh or anything else? :cry:

deusebio:
Let's not drift conversation. There is no proposal to use Terraform in the context of the changes discussed here in the KEP.

I just provided a way to consume Charmed Kubeflow with ambient if anyone wanted to give it a go and look around.

juliusvonkohout:
Yes, we can get this merged @deusebio and I definitely have the permissions to do so. But there must be someone dedicated working on it.

Issue #783, opened 2022-04-11
Kubeflow component integration with ML Metadata


richardsliu:
/kind feature

**Why you need this feature:**
Kubeflow currently doesn't have a unified metadata/artifact management story beyond what's supported in KFP. For example, the concept of a "ML experiment" exists in training and hyperparameter tuning, but there is no way to track it across separate Kubeflow components. Having unified metadata tracking allows users to aggregate things like:
* Experiment runs
* Datasets
* Metrics
* Trained artifacts
* Hyperparameter configurations
* etc

Originally Kubeflow covered this through the [Metadata](https://github.com/kubeflow/metadata) project but it has since been archived. There were some additional discussions around this, found in issue kubeflow/kubeflow#4955.

It would be great to revisit this problem and see if we can propose a unified interface for metadata and artifact storage, possibly by using [ML metadata](https://github.com/google/ml-metadata).

**Describe the solution you'd like:**

One problem with the original Kubeflow metadata project is that it comes with its own storage backend using MySQL, which makes it heavy-weight. We do not need to re-implement the storage backend since MLMD already solves that problem. Instead, we can make MLMD an optional installation, and write to it directly. This is what KFP is currently doing, see [this link](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/driver.go) for the code.

If we can define a unified data model and interface, it should be possible to build a light-weight library on top of ML metadata. It can be an optional import for training jobs and hyperparameter tuning jobs.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

tarilabs:
thanks @andreyvelich for bringing this back into attention

> One problem with the original Kubeflow metadata project is that it comes with its own storage backend using MySQL, which makes it heavy-weight

I'm not sure I understood this from the original posting 🤔 isn't MLMD backed by MySQL(/MariaDB/PostgreSQL) too?

juliusvonkohout:
As far as I know we want to rewrite or migrate away from ml-metadata.

/close

Issue #238, opened 2019-03-05
Create a kubeflow/metadata repository


zhenghuiwang:
It looks like metadata management should be a kubeflow component of its own. We need a repository e.g. `kubeflow/metadata` for its design proposals, api and schema docs, and implementations.

animeshsingh:
**metadata** gives it a much wider connotation. Better would be to call it **metadata-store**. Also we would love to collaborate on this one

mpvartak:
Few thoughts:

- Having a KubeFlow "native" metadata repo will make this metadata implementation the de-facto implementation for metadata in the KF ecosystem, whether/not it turns out to be the best implementation. 

- A KF metadata repo lowers the incentive for 3rd party developers to build a metadata store that will work with KubeFlow because it's not clear why a user should use a third-party vs. a pre-bundled metadata store. 

- As a parallel, there could have been a "Serving" repo in KF but there isn't a pre-specified serving component. This has allowed multiple serving components like Seldon, OpenVino, TFServing to be available. Having an official metadata store will essentially makes KF less open as a platform/framework

- Operationally, I understand the need to organize metadata efforts, but we should consider its impact on current and potential integrators into KF and signaling going forward. 

- Have a single spec for metadata seems more valuable compared to a single implementation repo as it will allow for multiple implementations.

zhenghuiwang:
@mpvartak Thanks for bringing up the concerns! Here are my clarifications.

> * Having a KubeFlow "native" metadata repo will make this metadata implementation the de-facto implementation for metadata in the KF ecosystem, whether/not it turns out to be the best implementation.
> * A KF metadata repo lowers the incentive for 3rd party developers to build a metadata store that will work with KubeFlow because it's not clear why a user should use a third-party vs. a pre-bundled metadata store.

The purpose of this repository is not to have everyone use one single implementation but accommodate  multiple implementations. Users should be able to choose one implementation when installing the Kubeflow cluster, because it is impossible to have everyone uses the same database.

> * As a parallel, there could have been a "Serving" repo in KF but there isn't a pre-specified serving component. This has allowed multiple serving components like Seldon, OpenVino, TFServing to be available. Having an official metadata store will essentially makes KF less open as a platform/framework

Serving is a relative independent component. The interface is clear: this is my model file, serve it. But metadata is about understanding the data produced by multiple components in KF. So we have to define the meaning and interface of metadata ourselves.

[...]

PR #892, opened 2025-08-01
KEP-897: First-Class MLflow Integration for Experiment Tracking in Kubeflow

mprahl:
GitHub issue: #897

Instead of building a new experiment tracking backend inside Kubeflow, the KEP proposes that Kubeflow deeply integrate with MLflow as a strong open-source option with an active community. The proposal focuses on making MLflow Kubernetes-native for Kubeflow through donation of the Kubernetes plugins, alignment with Kubeflow Profiles and multi-tenancy, a supported MLflow image and deployment path, and a UI strategy based on either launching out to MLflow or embedding it in the dashboard.


proposals/892-experiment-tracking/README.md
```diff
@@ -0,0 +1,514 @@
...
+
+Model Registry's current authorization model gates access at the API level through a proxy rather than at the individual
+resource or namespace level. This approach prevents the same Model Registry instance from being shared across teams that
+require isolation. This limitation conflicts with Kubeflow Pipelines' multiuser mode, which is the default deployment
+strategy for the Kubeflow platform. MLMD has the same limitation, so adding multi-tenancy support to Model Registry
+would address this gap for all of Kubeflow's metadata needs.
+
+This proposal introduces an optional/configurable authorization mode for Model Registry that leverages Kubernetes
+subject access review without adding custom resource definitions or modifying the existing API concepts. Instead, it
+maps Kubernetes RBAC concepts to the existing REST API entities at the namespace level.
```
tarilabs: I'm supportive of implementation solution(s) which do _not_ impose in MR to mix in code multiple cross-cutting concerns, ie "business logic" (data transformation and queries) and "authorizations" (tenancy) 👍 
It is a general enough driving principle and best-practice, but thought I'd add it here as well :)

andreyvelich:
Thank you for driving this @mprahl! Please can you create a tracking issue under `kubeflow/community`, so you can get the KEP number ?

It would be also good to also mention the history as I mentioned here: https://github.com/kubeflow/community/issues/783
https://github.com/kubeflow/community/issues/238

As previously discussed in
- https://github.com/kubeflow/community/issues/783
- https://github.com/kubeflow/community/issues/238

proposals/892-experiment-tracking/README.md
```diff
@@ -0,0 +1,514 @@
+# KEP-892: Centralize Experiment Tracking in Kubeflow
+
+## Summary
+
+This proposal aims to resolve the current fragmented and limited experiment tracking experience by expanding the
+**Kubeflow Model Registry** into a unified, centralized metadata store. Currently, experiment tracking is scattered
+across components like Kubeflow Pipelines (which requires pipeline execution for tracking) and Katib (limited to
```
andreyvelich: I am interested whether we should define the concept of Kubeflow Experiment outside of KFP and Katib ?
E.g. Experiment which sits on top of TrainJob, Katib Jobs, Pipelines, Spark Jobs.

proposals/892-experiment-tracking/README.md
```diff
@@ -0,0 +1,514 @@
...
+hyperparameter tuning). This leads to challenges such as **limited flexibility** for direct logging from Python scripts
+or Jupyter notebooks, a **fragmented user experience** across multiple interfaces, and **maintenance difficulties** due
+to reliance on the inactive MLMD project.
+
+The proposal tackles these issues by:
+
+- **Expanding Model Registry** into a central experiment tracking store for experiments, runs, metrics, and artifacts
```
andreyvelich: Would it be scope of Model Registry or we need another Kubeflow project for it?
If we decide to use Model Registry for it, we might want to find another name of this project.
cc @kubeflow/wg-data-leads

proposals/892-experiment-tracking/README.md
```diff
@@ -0,0 +1,514 @@
...
+  across all Kubeflow components.
+- Providing **MLFlow SDK compatibility**, enabling users to leverage familiar APIs while storing data in Kubeflow.
```
andreyvelich: Why MLFlow and not Weight & Biases: https://wandb.ai/site/ ? 
I found that these days more and more organizations adopt it.

proposals/892-experiment-tracking/README.md
```diff
@@ -0,0 +1,514 @@
...
+
+Currently, experiment tracking in Kubeflow is fragmented and tightly coupled to specific components. Kubeflow Pipelines
+requires users to run everything through pipelines to track experiments, while Katib's experiments are limited to
+hyperparameter tuning scenarios. This fragmentation creates several challenges:
+
+1. **Limited flexibility**: Users cannot log experiments directly from Python scripts, Jupyter Notebooks, or other
+   Kubeflow components without forcing everything through pipelines. This restriction often drives users to seek
+   solutions outside the Kubeflow ecosystem.
+1. **Fragmented experience**: Users must navigate multiple interfaces to correlate run results and evaluation data,
+   depending on which Kubeflow component they use (e.g., Pipeline runs, Katib experiments, Training Operator results).
```
andreyvelich:
```suggestion
   depending on which Kubeflow project they use (e.g., Pipeline runs, Katib experiments, TrainJob results).
```

proposals/892-experiment-tracking/README.md
```diff
@@ -0,0 +1,514 @@
...
+1. **Implement multitenancy in Model Registry**: Add multitenancy support to Model Registry to enable isolation at the
+   Kubernetes namespace level like Kubeflow Pipelines does today. This resolves the multitenancy gap that exists with
+   MLMD today.
+1. **Differentiate Katib experiments in the Kubeflow dashboard**: Update the Kubeflow dashboard to clearly distinguish
+   between Katib's hyperparameter tuning experiments and general Kubeflow experiments, improving user experience and
+   reducing confusion.
+
+### Non-Goals
+
+1. **Decoupling Katib's experiments from their current implementation**: While the Kubeflow community can revisit
```
andreyvelich: I think, the concept of Katib Experiment and Pipeline is different.
In Katib, Experiment is just a CRD which defines HP Tuning job: https://github.com/kubeflow/katib/blob/master/examples/v1beta1/hp-tuning/random.yaml

Katib maintains its own database just to allow metrics collector to push metrics (e.g. accuracy, loss): https://www.kubeflow.org/docs/components/katib/reference/architecture/#katib-control-plane-components.

It is much more lightweight compare to KFP.

proposals/892-experiment-tracking/README.md
```diff
@@ -0,0 +1,514 @@
...
+
+### SDK Integration
+
+#### MLFlow SDK Compatibility
+
+- Implement MLFlow plugins to connect to the Model Registry backend
+- Enable users to leverage MLFlow's familiar APIs while storing data in centralized Kubeflow infrastructure
+- Support automatic logging for popular ML frameworks (TensorFlow, PyTorch, scikit-learn, etc.) through MLFlow's SDK
+
+#### Kubeflow SDK Enhancement
```
andreyvelich: It would be nice to have some code snippet on how those integration will look like.

proposals/892-experiment-tracking/README.md
```diff
@@ -0,0 +1,514 @@
...
+
+The expanded Model Registry will include the following domain models, heavily influenced by
+[kubeflow/model-registry#1224](https://github.com/kubeflow/model-registry/issues/1224#issuecomment-3068005968). This is
+meant to be a high-level design with further refinement being done in a Model Registry specific KEP or design document.
+
+1. **Experiments**: A group of related runs. This would replace experiments in Pipelines and could potentially map to
+   experiments in Katib.
+1. **Users**: A user to associate runs, metrics, artifacts, etc. for auditing. This will generally map to the Kubernetes
+   identity.
+1. **Runs**: An execution in the machine learning workflow. In Pipelines, this maps to a pipeline run. In Katib, this
```
andreyvelich: Shall we have representation of Run depending on the Kubeflow CRDs, i.e. ?
```
TrainJob
OptimizeJob
SparkJob
Pipeline/Workflow
```
I can imagine that users might want to create Experiment where they only have training job, or they might want to create more complex Experiment with multiple steps (e.g. data process, training, evals).

proposals/892-experiment-tracking/README.md
```diff
@@ -0,0 +1,514 @@
...
+   allowing us to move away from wrapping MLFlow without breaking user code. Furthermore, popular plugins such as from
+   [Microsoft Azure](http://docs.azure.cn/en-us/machine-learning/how-to-use-mlflow-configure-tracking) would cause
+   significant disruption for the MLFlow community if they were to change their SDK plugin stance, making such changes
+   unlikely.
+
+## Design Details
+
+### Model Registry Domain Models
+
+The expanded Model Registry will include the following domain models, heavily influenced by
```
andreyvelich: We should add schema/ API spec in this KEP.

andreyvelich (commented):
cc WGs to review 
@kubeflow/wg-pipeline-leads @kubeflow/wg-data-leads @kubeflow/wg-automl-leads @kubeflow/wg-data-leads @kubeflow/kubeflow-steering-committee @kubeflow/wg-manifests-leads @kubeflow/wg-notebooks-leads

proposals/892-experiment-tracking/README.md
tarilabs: +1, in some regards we are seeing the need of a "AI Asset Registry", but I'm also favourable of simply "Kubeflow Registry"

proposals/892-experiment-tracking/README.md
tarilabs: We can also make link to:
https://github.com/kubeflow/model-registry/issues/1224#issuecomment-3068005968

proposals/892-experiment-tracking/README.md
juliusvonkohout: Yes, we need to keep the multi tenancy.

proposals/892-experiment-tracking/README.md
juliusvonkohout: i would stick with model registry

andreyvelich: I like the idea of Kubeflow Registry, since it makes very clear for users that this project is designed for metadata storage, and not only for model artifacts. Alternatively, we can name it: Kubeflow Tracker/Tracking, or find better name.

juliusvonkohout: I think the plan is that kubeflow model registry is mlflow compatible and we can then use the mlflow sdk also for pipelines, trainer etc. and get rid of ml-metadata CC @franciscojavierarceo

franciscojavierarceo: Yes +1 on MLflow.

tarilabs: related:
- https://github.com/kubeflow/model-registry/issues/1225
- https://github.com/kubeflow/model-registry/issues/1405

proposals/892-experiment-tracking/README.md
mprahl: I think we are on the same page. The overall goal is that an experiment in Model Registry could contain a mix of experiment runs from Pipelines, TrainJob, SDK, etc. Each run would indicate the source of where they came from, but the experiments are not exclusive to a particular source.

When Pipelines is integrated in this mode, it wouldn't have its own experiments in its database like today. Instead it would pull the list of available ones from Model Registry. We'd still keep the existing Pipelines experiments concept for standalone installations of Pipelines though.

proposals/892-experiment-tracking/README.md
mprahl: @andreyvelich I do really like Weight & Biases but MLflow has a plugin architecture that allows the SDK to use a different backend (Model Registry in this case). So we could add integrations with Weight & Biases (e.g. export experiments and metrics to it from KFP) but we could not easily reuse their SDK, which is the reason MLflow is proposed here.

proposals/892-experiment-tracking/README.md
mprahl: @andreyvelich it would require more thought of how to integrate it, but at a high-level, the domain models seem to map well:

* Katib experiment -> Model Registry experiment
* Katib trail -> Model Registry run
* Katib trail parameters -> Model Registry run parameters
* Katib trial metrics -> Model Registry run metrics

So I think Katib's APIs could stay the same but just also create/reuse an experiment in Model Registry and then create a "run" in the Model Registry experiment per executed "trial".

I think for the purpose of this KEP, aligning on it being theoretically possible for Katib to export this metadata to the domain models proposed for Model Registry would be enough. The Katib maintainers/community can then decide if they'd like to integrate after the Model Registry implementation.

proposals/892-experiment-tracking/README.md
mprahl: I'll add this in `Kubeflow SDK Implementation` under `Design Details`. The example is from a proof of concept developed by @kramaranya!

proposals/892-experiment-tracking/README.md
mprahl: So a run could be just a single TrainJob or it could a Pipeline with multiple steps (each step is a nested run). I'm trying to model this after MLflow so it's agnostic to the source that logged it.

proposals/892-experiment-tracking/README.md
mprahl: I'll clarify that the linked comment includes APIs as well.

proposals/892-experiment-tracking/README.md
rareddy: @juliusvonkohout @andreyvelich we also worked towards removing the MLMD see https://github.com/kubeflow/model-registry/issues/865

This gives us flexibility to extend the schema model to fit the future needs of the registry to evolve into supporting the experiments and towards other enhancements.

+100 I am open for name change, I get that it is a lot of work but if that better represents where we are leading into it is a welcome change IMO.

dhirajsb: BTW, the name Model Registry is still valid since experiment tracking is just part of the Model lifecycle and needs to be stored in Model Registry to be able to track model lineage.

andreyvelich: I think, we should align on what does Experiment mean in AI lifecycle ?
For example, if we consider any type of MLOps activity as Experiment (e.g. data preparation, training, HPOs, evals), we might need to find better name.

proposals/892-experiment-tracking/README.md
andreyvelich: I think, the concept of Katib Experiment is misleading. Katib Experiment is a batch job (e.g. OptimizeJob) that has its own scheduler (e.g. Suggestion) which makes decision whether it needs to create more runs (e.g. Trials).
Ideally, we should refactor Katib API towards OptimizeJob where we have clear definition of hyperparameter optimization job.

@kubeflow/wg-training-leads @astefanutti @franciscojavierarceo Any thoughts ?

andreyvelich: Before Experiment API, we did have the StudyJob concept that fully aligned with the original Google Vizier paper.
Ref: https://github.com/google/vizier

proposals/892-experiment-tracking/README.md
andreyvelich: What are the benefits to use MLFlow SDK if eventually we will implement tracking API in the Kubeflow SDK ?

franciscojavierarceo: MLFlow is the dominant player in the market. https://clickpy.clickhouse.com/dashboard/mlflow

FWIW Weights and Biases is a close second but MLFlow has dominated the market for longer. https://clickpy.clickhouse.com/dashboard/wandb

mprahl: @andreyvelich the main benefits of the MLflow plugin are the Kubeflow SDK could choose to wrap the MLflow SDK to support the powerful [autolog](https://mlflow.org/docs/latest/ml/tracking/autolog/) feature and existing workflows leveraging the MLflow SDK would continue to work with minimal changes when pointing to a Model Registry API with experiment tracking.

andreyvelich: >MLFlow is the dominant player in the mark

@franciscojavierarceo Looking at those charts, `wandb` has the same number of downloads per month as `mlflow`.


>[autolog](https://mlflow.org/docs/latest/ml/tracking/autolog/) feature and existing workflows leveraging the MLflow SDK would continue to work with minimal changes when pointing to a Model Registry API with experiment tracking.

@mprahl How we could leverage the autolog feature if at the end we should create Kubernetes CRDs ?
IIUC, autolog monkey-patch the code in supported ML libraries to enable metrics logging when users run `.fit()` methods.

mprahl: @andreyvelich Model Registry isn't CRD based and they already added some of the APIs mentioned in this KEP here: https://github.com/kubeflow/model-registry/pull/1318.

The design details in the second half of the KEP touch on this but MLflow has a tracking store plugin to allow leveraging MLflow's SDK but leverage a different Python package to execute the backend API calls:
https://mlflow.org/docs/latest/ml/plugins/#storage-plugins

kramaranya: > What are the benefits to use MLFlow SDK if eventually we will implement tracking API in the Kubeflow SDK ?

To me it makes sense just to wrap MLflow SDK in the Kubeflow SDK due to several reasons. It's not feasible to reimplement everything MLflow already provides. 

Model Registry will provide a plugin https://github.com/kubeflow/model-registry/issues/1225 which will allow us to use Model Registry as a backend alongside MLflow SDK by just simply setting `tracking.set_tracking_uri(model_registry_uri)`.

Using the MLflow SDK also keeps us from being tightly coupled to Model Registry. If we went with the MR SDK instead, users would aways need to have Model Registry deployed just to do experiment tracking in the Kubeflow SDK. With MLflow, existing users can use the APIs they already know and still store their data in Model Registry if they want to.

Another big advantage, as @mprahl mentioned, is autologging. The autologging ecosystem is huge and we won't  reimplement all of that with the MR SDK.

Since we don't have a local Model Registry yet, the MLflow SDK gives us a way to do local experiment tracking - like for local execution mode in Trainer. MLflow works out of the box with file storage, SQLite etc, so users can easily switch between any storage backend:

```
# Local
tracking.set_tracking_uri("file:./mlruns")

# MR backend  
tracking.set_tracking_uri("modelregistry+https://model-registry.svc:8080")

# MLflow server
tracking.set_tracking_uri("http://mlflow-server:5000")
```

andreyvelich: I see, that makes sene!

>@andreyvelich Model Registry isn't CRD based and they already added some of the APIs mentioned in this KEP here: https://github.com/kubeflow/model-registry/pull/1318.

I am trying to understand how we will leverage MLFlow autolog feature in the concept of runs.
I understand that Model Registry has ability to orchestrate Experiment and Runs which are not CRD-based, but at the end users should be able to create Kubeflow CRDs (e.g. TrainJob, SparkJob, Workflow) within their Experiments.

franciscojavierarceo: @andreyvelich technically MLflow had 1 million more downloads last month than Weight and Biases.

Also I believe both proof of concepts were done with MLflow and I believe we intend on eventually trying to be compatible with both.

kramaranya: @andreyvelich the autologging happens inside your training code that runs within the job containers, not at the orchestration level.

So for example you create a TrainJob CRD, then Kubeflow starts containers based on that CRD and inside those containers your training code runs with MLflow autologging enabled, which captures metrics and parameters and sends them to the Model Registry backend.

The CRDs orchestrate where/how the code runs and autologging just captures what happens during execution, so they're separate layers that work together. And users don't need to modify their CRDs at all, experiment tracking is just another library in their training code.

BTW, this is similar to [the demo](https://cloud-native.slack.com/archives/C0742LDFZ4K/p1751045374441719) I shared earlier with MLflow SDK in Kubeflow SDK.

andreyvelich: I see, that looks great! I guess, in case of distributed training (when we execute train function across multiple nodes), we will enable `autolog()` only for the MASTER NODE, like this:
```python
if dist.rank() == 0:
  mlflow.autolog()
```

kramaranya: Exactly, but we still need to disable it on other nodes, so we can just do it this way :
```
rank = dist.get_rank()
mlflow.autolog(disable = rank != 0)
```

proposals/892-experiment-tracking/README.md
tarilabs: we could also consider a rename such as "Kubeflow registry" (or anything else which we find name convergence) _after_ this work is completed so that, given more capabilities are added/will be added (like the catalog, experiment, etc) we will have more data and make a better informed decision.

this is also with the other hat-on-head that with *our present focus on Graduation, making a name change at this time, might be also strategically challenging.

cc @mprahl @andreyvelich wdyt?

mprahl: I'm fine with postponing the renaming conversation. We can clarify through documentation and the UI as part of the delivery.

andreyvelich: Sure, we can rename it after the CNCF graduation, and include it as a Northstar in this KEP.
I just would like us to align on the future project scope and goals.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,584 @@
...
+hyperparameter tuning scenarios. This fragmentation creates several challenges:
+
+1. **Limited flexibility**: Users cannot log experiments directly from Python scripts, Jupyter Notebooks, or other
+   Kubeflow components without forcing everything through pipelines. This restriction often drives users to seek
+   solutions outside the Kubeflow ecosystem.
+1. **Fragmented experience**: Users must navigate multiple interfaces to correlate run results and evaluation data,
+   depending on which Kubeflow project they use (e.g., Pipeline runs, Katib experiments, Training Operator results).
+1. **No unified tracking**: Users cannot easily compare runs, share insights, or maintain consistent metadata across the
+   entire Kubeflow ecosystem.
+1. **Maintenance challenges**: The Kubeflow Pipelines dependency on MLMD creates technical debt and limits future
```
juliusvonkohout: ML-metadata also violates a lot of security best practices and breaks hard multi-tenancy.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,584 @@
...
+- The original Kubeflow Metadata project was later archived, leaving Kubeflow Pipelines’ MLMD usage as the de facto
+  approach and creating fragmentation across components (as noted in
+  [Issue #783](https://github.com/kubeflow/community/issues/783)).
+
+This KEP builds on that history by centralizing experiment tracking within Model Registry, aligning the data model
+across Kubeflow components, and decoupling Kubeflow Pipelines from a hard dependency on MLMD while allowing optional
+interoperability.
+
+### Goals
+
```
juliusvonkohout, line 100: Add per namespace/profile multi-tenancy as goal.

proposals/897-experiment-tracking/README.md
juliusvonkohout: and has a lot of CVEs
```
Scanning  ghcr.io/kubeflow/kfp-metadata-envoy:2.5.0
+----------+------+--------+-----+
| Critical | High | Medium | Low |
+----------+------+--------+-----+
|    0     |  0   |   28   |  7  |
+----------+------+--------+-----+
Scanning  ghcr.io/kubeflow/kfp-metadata-writer:2.5.0
+----------+------+--------+-----+
| Critical | High | Medium | Low |
+----------+------+--------+-----+
|    13    | 372  |  973   | 845 |
+----------+------+--------+-----+
Scanning  gcr.io/tfx-oss-public/ml_metadata_store_server:1.14.0
+----------+------+--------+-----+
| Critical | High | Medium | Low |
+----------+------+--------+-----+
|    0     |  0   |   41   |  18 |
+----------+------+--------+-----+
```

mprahl:
I'm closing this KEP because my team no longer has capacity to take this on. If others want to pursue this, feel free to fork the KEP and I'll be happy to review and advise. :smile:

juliusvonkohout:
@mprahl may we keep it open for now? Just to have it tracked.

The stalebot will close it anyway if there is no activity on this topic

andreyvelich:
I agree with @juliusvonkohout!

Maybe we should put out a call for contributors to help us add Experiment Tracking support via MLFlow for Kubeflow sub-projects.
This feels like a really important capability that many of our users are asking for, and moving it forward would have a big impact on usability and Kubeflow adoption.

cc @kubeflow/wg-training-leads @kubeflow/wg-pipeline-leads @kubeflow/kubeflow-steering-committee @kubeflow/wg-manifests-leads @kubeflow/wg-notebooks-leads @kubeflow/wg-data-leads @kubeflow/kubeflow-sdk-team @kubeflow/kubeflow-outreach-committee @jbottum

tarilabs:
> My only question is: in the short to medium term, what approach should we take to deliver the most value to users?

Very IMHO an SPI that is 1:1 to the MlFlow API in the short term.
I'm aware is very limiting and naive, but at least forces to identify where the boundary for this integration lies. In turn, it should indeed make it easier to "direct" contributors/GSoC students if they want to integrate W&B (found the https://github.com/kubeflow/community/pull/892#discussion_r2263805436 ! 😄 ) or other tracking system, next.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,340 @@
...
+backend. The user-facing model should stay simple:
+
+- users write to standard MLflow APIs
+- operators deploy a supported MLflow installation for Kubeflow
+- Kubeflow components discover and reuse that installation through common conventions
+- the Kubeflow dashboard either embeds the MLflow UI or links to it
+
+### Donated Kubernetes Plugins
+
+Kubeflow should accept donation of
```
franciscojavierarceo, line 138: Agreed.

kramaranya (commented):
This is great, thanks @mprahl for driving this! 
I do agree that Kubeflow should integrate with MLflow rather than build a competing backend, since many Kubeflow users already use MLflow for experiment tracking
cc @gigabyte132 @tmvfb

franciscojavierarceo:
I approve this KEP, great work @mprahl 

Collaborating with the MLFlow community would be wonderful. 👏

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+assemble their own integration.
+
+This direction also makes Kubeflow more relevant to the GenAI ecosystem. MLflow is investing heavily in GenAI experiment
+tracking, tracing, and agent observability, so a strong Kubeflow integration helps the platform stay aligned with where
+the broader open-source ML tooling ecosystem is going.
+
+This KEP is intentionally platform-level. Component-specific behavior should be described in follow-up KEPs such as the
+in-flight Kubeflow Pipelines proposal
+[KEP-12862](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md), and other
+Kubeflow components can follow the same shared MLflow conventions.
```
andreyvelich: How we can ensure that we don't diverge MLFlow integration across Kubeflow tools?
I would love us to reduce duplication efforts across integration between: Spark Operator, Trainer, Optimizer, and Pipelines

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+## Motivation
+
+Experiment tracking in Kubeflow is fragmented across component-specific experiences. A user may train from a Notebook,
+launch a job with Kubeflow Trainer, and orchestrate end-to-end workflows with Kubeflow Pipelines, yet still have no
+standard way to view those runs together using the same tracking backend and user experience.
```
andreyvelich, line 64: I think, it would be nice to define What is Experiment, and What is Run from the Kubeflow point of view.
WDYT @mprahl ?

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Goals
+
+1. Adopt MLflow as the primary experiment tracking integration for Kubeflow users.
+1. Donate and maintain Kubernetes-native MLflow plugins in Kubeflow so that MLflow integrates cleanly with namespaces,
+   Profiles, and Kubernetes RBAC.
+1. Offer a supported MLflow deployment story for Kubeflow operators, including a rebuilt MLflow image with the donated
+   plugins preinstalled.
+1. Support a shared multi-tenant MLflow deployment in which profile namespaces act as natural MLflow workspace
+   boundaries.
```
andreyvelich: Can we have one of the goals to ensure that this integration is composable? Which should allow users that don't deploy all Kubeflow tools to benefit from it, similar as Kubeflow SDK?

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+1. Improve Kubeflow's relevance for GenAI workflows by aligning with an experiment tracking ecosystem that is investing
+   in tracing and agent observability.
+1. Provide a path for Kubeflow components to integrate with the shared MLflow deployment over time, with
+   [KEP-12862](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md) serving
+   as one concrete example already underway.
```
andreyvelich: Does this have dependency on Profile Controller or other common components for multi-tenant deployment of Kubeflow Platform?

proposals/897-experiment-tracking/README.md
andreyvelich: +1 on this.
Shall we discuss maintainers?

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+that model:
+
+- a profile namespace becomes the default MLflow workspace boundary
+- Kubernetes RBAC remains the source of truth for access decisions
+- a shared MLflow deployment can serve many namespaces without abandoning isolation
+- component UIs and SDKs should consistently map the active Kubeflow Profile or namespace to the default MLflow
+  workspace
+
+This is a better fit for Kubeflow than a generic MLflow deployment that knows nothing about namespaces, service account
+tokens, or Profile-managed multi-user clusters.
```
andreyvelich: Does it require to install profile controller to use it?
What if users have their own identity providers on their k8s clusters?

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+Kubeflow should not require every operator to build a custom MLflow image or invent their own deployment manifests.
+Instead, the community should provide a supported distribution story with the following pieces:
+
+1. A Kubeflow-published MLflow container image rebuilt from upstream MLflow and shipped with the donated Kubernetes
+   plugins preinstalled.
+1. A supported Helm-based deployment path. Kubeflow should prefer contributing the missing requirements upstream to
+   [mlflow/mlflow#21973](https://github.com/mlflow/mlflow/pull/21973). If the upstream chart cannot accommodate the
+   Kubernetes plugins, workspace configuration, and multi-tenant auth requirements in time for Phase 1, Kubeflow should
+   maintain its own chart or manifests until the upstream path is sufficient.
```
andreyvelich: So this can be used as add-on to existing deployment, right? For example, if I want to integrate experiment tracking with Trainer, I should install this:
```
helm install kubeflow-trainer oci://ghcr.io/kubeflow/charts/kubeflow-trainer --version 2.2.0
helm install kubeflow-mlflow oci://ghcr.io/kubeflow/charts/kubeflow-mlflow --version 0.1.0
```

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+1. A request originates from a notebook, pipeline workload, training job, or interactive user session running in a
+   namespace associated with a Kubeflow Profile.
+1. The caller authenticates to MLflow using either projected Kubernetes service account credentials for workloads or the
+   existing Kubeflow web identity path for interactive users.
+1. The authorization plugin validates that identity and checks Kubernetes RBAC for the requested MLflow operation.
+1. If authorized, MLflow serves the requested experiments, runs, metrics, and artifacts within the deployment's
+   namespace-aware tenancy model.
+
+Kubeflow components should reuse this shared authentication model rather than defining their own connection and identity
+flow.
```
andreyvelich, line 317: So looking at Kubeflow projects, we have these:

1. Kubeflow Pipelines - access via KFP API server
2. Kubeflow Spark Operator, Trainer, and Optimizer - access via Kubernetes API server

How both of these models, can reuse shared auth model?

andreyvelich (commented):
Thanks a lot for this @mprahl, overall looks awesome.
Kubeflow users wanted to have this capability since 2019 🚀 

cc @kubeflow/kubeflow-trainer-team @akshaychitneni @nabuskey @kubeflow/wg-data-leads @kubeflow/kubeflow-sdk-team @kubeflow/kubeflow-kale-team @kubeflow/wg-notebooks-leads appreciate your review too!

proposals/897-experiment-tracking/README.md
juliusvonkohout: The access yes and we should embed the mlflow UI directly into the dashboard as iframe.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+1. Expand Kubeflow Model Registry to become the experiment tracking backend for Kubeflow.
+1. Design a brand-new Kubeflow-native experiment tracking service with its own API, storage model, and UI.
+1. Rework Pipelines' or Katib's current experiment model as part of this proposal.
+
+## Proposal
+
+### MLflow-First Vision
+
+Kubeflow should package, integrate, and support MLflow directly rather than building a parallel experiment tracking
```
juliusvonkohout, line 128: I actually have such a setup myself. So we can create mlflow secrets per namespace

juliusvonkohout: I wil also try to use a common  namespace name such as mlfor for running the shared server. One namespace per component is in general a good idea. Kubeflow-pipelines, kubeflow-mlflow, kubeflow-katib, kubeflow-trainer etc. could make a lot of sense.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Deployment and Distribution
+
+Kubeflow should not require every operator to build a custom MLflow image or invent their own deployment manifests.
+Instead, the community should provide a supported distribution story with the following pieces:
+
+1. A Kubeflow-published MLflow container image rebuilt from upstream MLflow and shipped with the donated Kubernetes
+   plugins preinstalled.
+1. A supported Helm-based deployment path. Kubeflow should prefer contributing the missing requirements upstream to
+   [mlflow/mlflow#21973](https://github.com/mlflow/mlflow/pull/21973). If the upstream chart cannot accommodate the
```
juliusvonkohout: I would like to rely on upstream mlflow as much as possible and reduce our maintenance burden

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+   Kubernetes plugins, workspace configuration, and multi-tenant auth requirements in time for Phase 1, Kubeflow should
+   maintain its own chart or manifests until the upstream path is sufficient.
+1. Kubeflow-specific documentation and examples covering the supported MLflow deployment pattern, multi-user
+   authentication, Profiles integration, and any required Kubeflow packaging or configuration.
+1. Version guidance that tells operators which MLflow version, plugin revision, and chart configuration are supported by
+   each Kubeflow release. This proposal depends on MLflow `>= 3.10`, since MLflow 3.10 includes workspace support, which
+   is a core dependency for the multi-tenant design. As MLflow 3.11 lands, Kubeflow should also adopt its built-in
```
juliusvonkohout: Then let's wait for and start with 3.11

proposals/897-experiment-tracking/README.md
juliusvonkohout: All components can directly talk to mlflow with for example the service account as bearer token.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+1. Provide a path for Kubeflow components to integrate with the shared MLflow deployment over time, with
+   [KEP-12862](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md) serving
+   as one concrete example already underway.
+1. Provide a clear UI strategy in which Kubeflow either embeds MLflow or links users out to the MLflow UI.
+
+### Non-Goals
+
+1. Expand Kubeflow Model Registry to become the experiment tracking backend for Kubeflow.
+1. Design a brand-new Kubeflow-native experiment tracking service with its own API, storage model, and UI.
+1. Rework Pipelines' or Katib's current experiment model as part of this proposal.
```
HumairAK: Agree that this should not block the effort, but I do think it's a risk that we take on of confusing UX if multiple notions of experiments show up in kubeflow UI. Per the last KFP community call we learned that Katib looks like it's already renaming experiments. I think KFP should do the same. 

I'm not sure if and where we want that in this KEP, but at the least we should point it out, and nudge KFP to consider doing this for next major release.

proposals/897-experiment-tracking/README.md
HumairAK: mlflow 3.11 is now released, so I think we can just go with 3.11 as the min requirement

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,354 @@
...
+[KEP-12862](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md).
+
+### UI Strategy
+
+The UI strategy for MLflow integration should be explicit and limited to two supported modes:
+
+1. **Launch-out link**: Kubeflow surfaces links such as "Open in MLflow" from notebooks, training jobs, pipelines, or
+   dashboard pages. This is the simpler near-term path and uses the MLflow UI directly.
+1. **Embedded MLflow experience**: Kubeflow embeds the MLflow UI, or selected MLflow views, inside the Kubeflow
+   dashboard for a more cohesive experience.
```
HumairAK: I think we should just move embedded suggestion out of scope for now, maybe move it to a "Future/Follow up goals", it doesn't seem like it's covered in the two phases, but multiple times it's brought up in this doc as something we might pursue.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Kubeflow Component Integration Pattern". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Kubeflow Component Integration Pattern". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Goals". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Donated Kubernetes Plugins". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Kubeflow Profiles and Multi-tenancy". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Deployment and Distribution". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Authentication and Authorization". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Deployment and Distribution". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Deployment and Distribution". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Deployment and Distribution". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "Kubeflow Component Integration Pattern". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in section "UI Strategy". Please take a look.

proposals/897-experiment-tracking/README.md
mprahl: Addressed in sections "Kubeflow Profiles and Multi-tenancy" and "UI Strategy". Please take a look.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+The shared MLflow data model in this KEP is intentionally simple: an MLflow experiment is the cross-component container
+for related work, and an MLflow run is the record of a single execution or logging session within that experiment.
+[KEP-12862](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md) applies
+that model to Kubeflow Pipelines by creating one parent MLflow run per KFP pipeline run and nested MLflow runs for
+individual tasks and loop iterations.
+
+To avoid overloading the word "experiment" inside Kubeflow, follow-up component work should align user-facing
+terminology with that shared model. In particular, this KEP sets the direction that KFP's current Experiment grouping
+should be renamed to Run Group, while Katib or Kubeflow Optimizer follow-up work should avoid introducing another
```
mprahl: @droctothorpe @zazulam @HumairAK could you please provide your opinion on renaming KFP experiments to "Run Groups"?

andreyvelich: You can also sat that Katib will use OptimizationJob instead of Experiment API as we agreed in this doc: https://docs.google.com/document/d/1Y8IJ-UdZ7VCEAlax_xEFbbqEi7EB6SfIX4D7ua-xn4M/edit

cc @akshaychitneni @aniket2405 @Krishna-kg732

proposals/897-experiment-tracking/README.md
andreyvelich: What if users have their own way to manage namespaces? Is there a way for us to install MLFlow workspace into existing namespace without using the Profiler controller?
I would hesitate to introduce hard dependency on it.

@thesuperzapper @juliusvonkohout @astefanutti Any thoughts on this?

proposals/897-experiment-tracking/README.md
```diff
@@ -133,68 +141,84 @@ The donated work provides two important capabilities:
...
+Kubeflow components that choose to integrate with MLflow should build on the shared deployment, auth, workspace, and
+terminology model described in this KEP. This platform contract is intentionally standardized here so follow-up
+component KEPs do not each need to re-solve tenancy, credentials, or UI hand-off independently.
 
-### UI Strategy
+The shared MLflow data model in this KEP is intentionally simple: an MLflow experiment is the cross-component container
+for related work, and an MLflow run is the record of a single execution or logging session within that experiment.
+[KEP-12862](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md) applies
+that model to Kubeflow Pipelines by creating one parent MLflow run per KFP pipeline run and nested MLflow runs for
+individual tasks and loop iterations.
```
andreyvelich, line 219: Can we map MLFlow Experiment, MLFlow Run, and nested Runs across all Kubeflow tools?
We discussed with @kramaranya, how we can map this model to OptimizationJobs too.

My initial thoughts:
1. Kubeflow Pipelines: Experiment <-> Pipeline, Run <-> PipelineRun, nested Run <-> PipelineRun' tasks.
2. Kubeflow Trainer: Experiment and Run <-> TrainJob
3. Kubeflow Optimizer: Experiment and Run <-> OptimizationJob, nested Run <-> OptimizationJob' Trial
4. Kubeflow Spark: Experiment and Run <-> SparkApplication.

@astefanutti @kramaranya @Fiona-Waters @tenzen-y @akshaychitneni @Shekharrajak @tariq-hasan @nabuskey @vara-bonthu @ChenYi015 @chensun  any thoughts?

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+1. A Kubeflow-published MLflow container image rebuilt from upstream MLflow and shipped with the donated Kubernetes
+   plugins preinstalled.
+1. A supported Helm-based deployment path. Kubeflow should prefer contributing the missing requirements upstream to
+   [mlflow/mlflow#21973](https://github.com/mlflow/mlflow/pull/21973), where review feedback is already capturing several
+   Kubeflow-specific needs. If the upstream chart cannot accommodate Kubeflow's plugin, deployment, and authentication
+   requirements in time for Phase 1, Kubeflow should maintain its own chart or manifests only until the upstream path is
+   sufficient.
+1. Default Helm values that deploy MLflow into a dedicated namespace such as `kubeflow-mlflow`, separate from
+   component namespaces.
```
andreyvelich: Can we just install the MLFlow control plane into system namespace?

```suggestion
1. Default Helm values that deploy MLflow into a system namespace: `kubeflow-system`
```

proposals/897-experiment-tracking/README.md
andreyvelich: Why do we need to have namespace per component? Can we always install the Kubeflow control plane component into `kubeflow-system` namespace as we do for Trainer?
Ref: https://github.com/kubeflow/manifests/issues/2937

proposals/897-experiment-tracking/README.md
juliusvonkohout: If embedding is possible I would definitely try to do it :-)

proposals/897-experiment-tracking/README.md
mprahl: @andreyvelich yes, for sure. There is no direct connection to Kubeflow profiles. The MLflow workspaces are 1x1 mappings to Kubernetes namespaces. Here is the doc for that specific part of the plugin:
https://github.com/opendatahub-io/mlflow-kubernetes-plugins/blob/main/docs/workspace-provider.md

For the Kubeflow profile case, we would likely just set `MLFLOW_K8S_WORKSPACE_LABEL_SELECTOR` env var to match the labels set for Kubeflow profile namespaces.

DaoDaoNoCode:
@mprahl Hello, our team is interested in contributing to the community and would like to work on this! cc @bobbravo2

mprahl:
> @mprahl Hello, our team is interested in contributing to the community and would like to work on this! cc @bobbravo2

Thanks! Specifically, this team is willing to contribute to the UI embedded work. :smile:

proposals/897-experiment-tracking/README.md
mprahl: In light of @DaoDaoNoCode volunteering his team to work on this, I'm changing the KEP to include iframe embedding as the initial deliverable and a future goal to use React module federation for a better UX pending upstream MLflow contributions.

proposals/897-experiment-tracking/README.md
mprahl: Added a suggested mapping in the KEP in the latest commit based on this comment!

proposals/897-experiment-tracking/README.md
mprahl: Addressed in the latest commit.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,474 @@
...
+1. The best integrated UI experience may depend on upstream MLflow changes that Kubeflow does not control outright.
+1. Some users may prefer a more deeply Kubeflow-specific experiment tracking experience than an MLflow-first design
+   provides.
+
+## Alternatives
+
+### Expand Model Registry
+
+Kubeflow could continue with the earlier direction of enhancing Model Registry to become the experiment tracking backend
+for the platform.
```
jonburdo, line 490: What is the role of [kubeflow model registry](https://github.com/kubeflow/model-registry) in the kubeflow ecosystem, given the direction this KEP? This may warrant a KEP of it's own, but I want to mention it here, and hear and high-level thoughts.

For example, does kf model registry continue to serve model lifecycle needs separately from mlflow's experiment tracking and model registry features? Or does mlflow's own model registry eventually replace kf model registry? Maybe the tooling/infra around kubeflow model registry and kubeflow sdk serve to integrate various components, extending mlflow's capability and usefulness within kubeflow.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,474 @@
...
+terminology with that shared model. In particular, this KEP sets the direction that KFP's current Experiment grouping
+should be renamed to Run Group, while Katib or Kubeflow Optimizer follow-up work should rename to `OptimizationJob`
+terminology, as discussed in the
+[Kubeflow Optimization API design doc](https://docs.google.com/document/d/1Y8IJ-UdZ7VCEAlax_xEFbbqEi7EB6SfIX4D7ua-xn4M/edit),
+instead of introducing another conflicting "experiment" concept.
+
+### UI Strategy
+
+The initial UI deliverable should be an embedded iframe experience in which Kubeflow embeds selected MLflow views
+inside Central Dashboard or other Kubeflow UIs.
```
DaoDaoNoCode: Worth noting a limitation here: with a plain upstream MLflow iframe, you get the full MLflow shell (header, sidebar, Model Registry, Prompts, etc). Cross-origin iframes don't allow CSS injection or DOM manipulation from the parent, so there's no way to strip those elements from the Kubeflow side. "Selected MLflow views" in practice means the full UI loaded at a specific URL, not a surgically embedded component. Users could freely navigate away from experiment tracking to unrelated MLflow pages from within the Kubeflow dashboard.                                                                                 
This is a known Phase 1 limitation of the iframe approach and should be documented as such unless we contribute an embed/kiosk mode to upstream MLflow. However, with the plan of supporting module federation to upstream MLflow in the future, I think it might not be worth the effort.
Just want to state the limitation here and know your thoughts. Thanks! @mprahl

proposals/897-experiment-tracking/README.md
mprahl: I think coexistence is the most pragmatic near-term path.

My current view is that MLflow can cover the experiment tracking experience and optionally the model registry experience if admin wants to, while Kubeflow Model Registry can continue to cover registry-oriented and deployment-oriented capabilities. I'd rather get community feedback on how Kubeflow is deployed with MLflow before making any further decisions unless the Kubeflow Model Registry working group has a strong opinion.

For now, I could see us extracting pieces like the catalog experience and the KServe storage adapter into a more backend-agnostic repo so they can work with either MLflow or Kubeflow Model Registry. That would reduce coupling and let us evaluate, based on real usage, whether those capabilities should remain shared, move more toward MLflow, or stay primarily aligned with Model Registry.

@kubeflow/wg-data-leads do you have thoughts?

proposals/897-experiment-tracking/README.md
mprahl: Thanks @DaoDaoNoCode I can add this as a limitation. We can also consider "small" patches applied in the container image build when we redistribute MLflow with our plugins preinstalled.

proposals/897-experiment-tracking/README.md
tarilabs: > extracting pieces like the catalog experience and the KServe storage adapter into a more backend-agnostic repo so they can work with either MLflow or Kubeflow Model Registry

+1 , Model Catalog, MCP Catalog and in general Catalog capabilities are already isolated, so that part in my view is already sort of covered, although it will help the rename currently in action following:

- https://github.com/kubeflow/community/pull/907#pullrequestreview-3608586640
- https://github.com/kubeflow/model-registry/pull/2239

What's not yet in-action in my view are:

1.
extracting the capabilities of:
- Storage (especially OCI ModelCar)
- Signature

into self-sufficient capabilities possibly in KF SDK would be nice,

2.
and making the async-upload job work not only using KF MR as a metadata store, but also MLflow as metadata store (of the Models)

tarilabs: One part that need clarification from the authors of this KEP @mprahl (imho) is once MLflow is integrated as model metadata storage in Kubeflow, how the deployment tracking would work (in MLflow)?

That would allow to assess how to best proceed with the [Isvc reconciler](https://github.com/kubeflow/model-registry/blob/main/cmd/controller/main.go) capabilities of KF MR; that needs an indication how to remap metadata that are currently mapped to KF MR on top of MLflow

rareddy: >whether those capabilities should remain shared, move more toward MLflow, or stay primarily aligned with Model Registry.

@mprahl with the two MR solutions, wondering if you have any suggestions to pull model metadata from mlflow to show in KF MR for better user experience. If there is plugin we can write may be worth while IMO.

mprahl: @tarilabs @rareddy Good questions. I think this is out of scope for this KEP. The intent here is to define MLflow as the first-class experiment tracking integration for Kubeflow, including the shared platform contract around tenancy, auth, deployment, UI hand-off, and alignment with MLflow's GenAI direction.

My view is that Kubeflow Model Registry can continue to cover model registry and model deployment capabilities. This KEP does not currently propose deeper metadata or deployment-tracking integration between MLflow and Kubeflow Model Registry. If, after adoption, we see a strong need for that, I think it would be reasonable optional follow-up work for the Kubeflow Data WG to pursue. I'm happy to help and provide guidance there too.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,487 @@
...
+   control remains the source of truth. It supports direct `SelfSubjectAccessReview` checks against caller tokens and a
+   trusted-proxy `SubjectAccessReview` mode that consumes forwarded user and group headers.
+
+By bringing these capabilities into Kubeflow, the community can evolve them in the open alongside Profiles, dashboard
+integration, and component-specific adoption work.
+
+The current repository is Apache-2.0 licensed, which is compatible with Kubeflow. Before Phase 1 is considered complete,
+Kubeflow and OpenDataHub maintainers should agree on transferring the repository to Kubeflow community ownership, with
+[`mprahl`](https://github.com/mprahl), [`HumairAK`](https://github.com/HumairAK), and any additional volunteers serving
+as the initial maintainer group with clear release responsibilities.
```
andreyvelich: Shall we say that WG Pipelines initially own this code?
We can add this repo to the WG assets: https://github.com/kubeflow/community/blob/master/wgs.yaml#L455-L465

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,487 @@
...
+[KEP-12862](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md) applies
+that model to Kubeflow Pipelines by creating one parent MLflow run per KFP pipeline run and nested MLflow runs for
+individual tasks and loop iterations.
+
+The intended cross-component mapping is:
+
+- MLflow experiment: the shared grouping for related work across Kubeflow tools
+- Kubeflow Pipelines pipeline run: one parent MLflow run, with nested MLflow runs for component tasks and loop
+iterations
+- TrainJob or SparkApplication execution: one MLflow run for that execution
```
andreyvelich, line 226: @mprahl @kramaranya I am also curious how we can map the MLFlow Experiment concept when TrainJob or OptimizationJob is submitted via KFP?
Not a blocker, we can discuss it later.

proposals/897-experiment-tracking/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+Kubeflow should use the same trusted-ingress pattern for machine-to-machine traffic as well. Rather than sending raw
+bearer tokens directly to MLflow, trusted ingress should validate supported JWTs for both browser and machine-to-machine
+requests, derive `kubeflow-userid` and `kubeflow-groups` from validated claims, and forward only those normalized
+headers to MLflow.
+
+A concrete example of that ingress pattern looks like:
+
+```yaml
+apiVersion: security.istio.io/v1beta1
```
andreyvelich, line 359: Do we have dependency on Istio in that case? I remember we talked before that we would like to integrate with Gateway API moving forward.

andreyvelich: cc @juliusvonkohout @thesuperzapper

.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
```
andreyvelich: Remove this?

andreyvelich (commented):
Thanks for the updates @mprahl!
Overall, +1 on this to move forward. Left a few small comments.

juliusvonkohout:
As member of the KSC I vote in general in favor. Technical details and open question are not concerning enough for me to wait with the vote. We will find a way to integrate this at the platform level. I can help with the maintenance as well, since i anyway need to deal with the integration into Kubeflow platform.

andreyvelich (commented):
+1 for this, just left a few small comments @mprahl.

chasecadet:
Looks good! Building a deep integration with MLflow for our go-to ML tracking and supporting a Helm chart would benefit the e2e story.  I see this as a plugin or integration. Let's call out that Kubeflow (the tools/components/) integrates with MLflow, and that MLflow is not a Kubeflow project. It's a dependency, and we are essentially saying "MLflow won" the OSS registry war here, and we want to provide that functionality to our community. Excited to see this in action and help folks build, deploy, and serve models in a more deterministic manner wherever they see fit.  I vote yes!

proposals/897-experiment-tracking/README.md
mprahl: @astefanutti do you have thoughts?

proposals/897-experiment-tracking/README.md
mprahl: @andreyvelich I'm fine with Gateway API as well as long as it supports Kubernetes SA tokens too. That is actually how MLflow is deployed in Red Hat OpenShift AI.

mprahl:
Thanks for the reviews! I address the comments in the last commit.

proposals/897-experiment-tracking/README.md
andreyvelich: Yes, I would prefer us to use this integration. That should help users to choose what they want.

andreyvelich:
Thanks for this effort @mprahl!
Really excited to see this moving forward – the community has been waiting for this integration for the past 6 years!

/lgtm
/approve
/hold in case others have any followup comments.
/assign @franciscojavierarceo @thesuperzapper @juliusvonkohout @chasecadet

mprahl:
@franciscojavierarceo @thesuperzapper @juliusvonkohout @chasecadet any objections to the KEP merging now?

andreyvelich:
Yes, we should be good to move this forward 🚀 
/hold cancel

PR #837, opened 2025-03-18
Adjust the platform/manifests charter according to the last 5 years and the state of 2025

juliusvonkohout:
@kimwnasptd for lgtm since these changes are usually and historically done by the manifests WG leads (@kimwnasptd and @juliusvonkohout)

CC @kubeflow/kubeflow-steering-committee

https://github.com/kubeflow/community/pull/832#issuecomment-2731060522 and https://github.com/kubeflow/community/issues/834#issuecomment-2733033791 also provides a lot of context and argumentation.

"
# WG Manifests Charter

This charter adheres to the conventions, roles and organization management
outlined in [wg-governance].

## Scope

We simply (automatically) synchronize the application and dependencies manifests to then elaborately combine (configure)them for full platform experience.
Providing a consistent and tested end-to-end multi-tenant experience is the most important task of the platform/manifests WG.
To achieve this we maintain an extensive testing suite that covers most basic scenarios users would expect from a Platform for ML orchestration.
We also provide the documentation regarding, but not limited to installation, extension, security and architecture to enable users to run their own ML Platform on Kubernetes.
Users may choose to derive from platform/manifests to create so called distributions, which are opinionated to satisfy individual requirements.
Users may also choose to install individual components without the benefits of the platform.

#### Code, Binaries and Services

- Enable users / distributions to install, extend and maintain Kubeflow as a end-to-end multi-tenant platform for multiple users
- This includes dependencies, security efforts and exemplary integration with popular tools and frameworks.
- Users can also install individual components without the benefits of the platform, but then they could also just directly fetch them from the WG releases.
- Synchronize the manifests between working groups and make sure via integration tests that the components work end-to-end together as multi-tenant platform
- Release tested releases of the Kubeflow platform for downstream consumption
- We try to be compatible with the popular Kubernetes clusters (Kind, Rancher, AKS, EKS, GKE, ...)
- We provide hints and experimental examples how a user / distribution could integrate non-default external authentication (e.g. companies Identity Provider) and popular non-default services on his own
- We in general document the installation of Kubeflow as a platform and / or  individual components including common problems and architectural overviews.
- There is the evolving and not exhaustive list of dependencies for a proper multi-tenant platform installation: Istio, KNative, Dex, Oauth2-proxy, Cert-Manager, ...
- There is the evolving and not exhaustive list of applications:  KFP, Trainer, Dashboard, Workspaces / Noteboks, Kserve, Spark, ...

### With Application Owners

[...]

PR #896, opened 2025-08-12
Update wgs.yaml

juliusvonkohout:
@andreyvelich @kimwnasptd 
based on https://github.com/kubeflow/community/pull/871#discussion_r2266827491

### DIFF
--- a/wgs.yaml
+++ b/wgs.yaml
@@ -193,13 +193,13 @@ workinggroups:
   - name: kfctl
     owners:
     - https://raw.githubusercontent.com/kubeflow/kfctl/master/OWNERS
-- dir: wg-manifests
-  name: Manifests
+- dir: wg-platform
+  name: Platform & Security
   mission_statement: >
-    WG Manifests maintains a centralized repository of manifests.
+    WG Platform maintains a centralized repository for the Kubeflow AI reference platform and security topics.
 
   charter_link: charter.md
-  label: area/wg-manifests
+  label: area/wg-platform
   leadership:
     chairs:
     - github: juliusvonkohout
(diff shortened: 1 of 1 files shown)


andreyvelich (commented):
But we should have a charter/vote to establish this working group defined, as we started in: https://github.com/kubeflow/community/pull/837
Since CNCF will review our governance structure for graduation, we should keep our working group names consistent everywhere.

cc @kubeflow/kubeflow-steering-committee

andreyvelich:
>It is more renaming the existing one. So should we vote here ? 

We can definitely discuss it with KSC, but I suggest that we rename/create/change WGs after we graduate with CNCF.
We need to make sure that our WG names are consistent, since CNCF will review it: https://www.kubeflow.org/docs/about/community/#kubeflow-working-groups

As of now, we have 6 WGs + WG ML Experience + WG Data.

wgs.yaml
```diff
@@ -193,13 +193,13 @@ workinggroups:
   - name: kfctl
     owners:
     - https://raw.githubusercontent.com/kubeflow/kfctl/master/OWNERS
-- dir: wg-manifests
-  name: Manifests
+- dir: wg-platform
+  name: Platform & Security
   mission_statement: >
-    WG Manifests maintains a centralized repository of manifests.
+    WG Platform maintains a centralized repository for the Kubeflow AI reference platform and security topics.
```
chasecadet, line 199: My vote is we stick to technical purity and just call it the 'Kubeflow Platform.' for the team description. 

    'AI' is redundant buzzword-chasing: Calling it the 'Kubeflow AI platform' is like saying 'ATM machine.' Kubeflow is intrinsically an ML platform, and we shouldn't sacrifice accuracy just to chase a marketing trend. If folks don't know it's for AI, they need to do a quick read on what Kubeflow is first anyway.

    'Reference' is too clunky: 'Kubeflow reference platform' is us trying to fit too much data into one sentence. Yes, Kubeflow is an ecosystem of tools, but we've integrated them into a platform for a cohesive story.

Think of it like Vertex. Nobody calls it the 'Vertex reference platform.' Users can bolt on whatever integrations they want to enhance it, but at its core, it remains Vertex. If someone wants to use our setup as a baseline 'Kubeflow core' and build on top of it or around it, they absolutely can.

tarekabouzeid: I think `Kubeflow platform` is more inclusive for both ML and AI.

juliusvonkohout: Then lets put this to a vote in the KSC

christian-heusel:
Given the newly formed Kubeflow Distribution Committee (which superseeds the WG) I think this can be closed @juliusvonkohout

andreyvelich (commented):
Yes, we have KDC now
/close

Issue #924, opened 2025-11-25
KEP-924: Filesystem-Based Artifact Storage for Kubeflow Pipelines


hbelmiro:
## Description

This KEP proposes adding PVC-based artifact storage as an alternative to S3/GCS/MinIO.

**Problem:** KFP requires object storage (S3/GCS/MinIO), which adds complexity for local/air-gapped deployments.

**Solution:** New `kfp-artifacts://` URI scheme that routes artifacts through an artifact server API, with storage on Kubernetes PVCs.

hbelmiro:
Closed in favor of https://github.com/kubeflow/pipelines/issues/12513.

PR #927, opened 2025-12-01
Add KEP-924: Introduce PVC-based artifact storage for Kubeflow Pipelines

hbelmiro:
Resolves: https://github.com/kubeflow/community/issues/924

### DIFF
--- a/proposals/924-pvc-artifact-storage/README.md
+++ b/proposals/924-pvc-artifact-storage/README.md
@@ -1,0 +1,1650 @@
+# KEP-924: Filesystem-Based Artifact Storage for Kubeflow Pipelines
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Architecture at a Glance](#architecture-at-a-glance)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Overview](#overview)
+  - [User Stories](#user-stories)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Configuration Approach](#configuration-approach)
+  - [Storage Backend Selection](#storage-backend-selection)
+    - [URI, API Endpoint, and Filesystem Path Relationship](#uri-api-endpoint-and-filesystem-path-relationship)
+  - [Artifact Server Architecture](#artifact-server-architecture)
+    - [Backend Responsibilities](#backend-responsibilities)
+    - [Request Routing](#request-routing)
+    - [UI Integration](#ui-integration)
+  - [PVC Management](#pvc-management)
+    - [PVC Creation and Configuration](#pvc-creation-and-configuration)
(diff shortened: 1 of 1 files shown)


proposals/924-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1650 @@
...
+│  Artifact  │ │  Artifact  │ │  Artifact  │ │     │
+│  Server    │ │  Server    │ │  Server    │ │ ... │
+│  (ns1)     │ │  (ns2)     │ │  (ns3)     │ │     │
+└─────┬──────┘ └─────┬──────┘ └─────┬──────┘ └──┬──┘
+      │ mounts       │ mounts       │ mounts    │
+      ▼              ▼              ▼           ▼
+┌────────────┐ ┌────────────┐ ┌────────────┐ ┌─────┐
+│  PVC (ns1) │ │  PVC (ns2) │ │  PVC (ns3) │ │ ... │
+│ /artifacts │ │ /artifacts │ │ /artifacts │ │     │
+└────────────┘ └────────────┘ └────────────┘ └─────┘
```
juliusvonkohout, line 519: That is now the execption and we go directly to seaweedfs with per namespace credentials instead https://github.com/kubeflow/manifests/releases/tag/v1.11.0-rc.1 So we have empty zero overhead namespaces by default.

hbelmiro: @juliusvonkohout sorry, I'm not following.
If users need complete PVC isolation, we need one artifacts server in each namespace because the launcher wouldn't be able to access the PVC directly in a different namespace.
That said, central mode is the default in this KEP (single artifact server + PVC for all namespaces). Namespace-local mode is optional for environments with strict isolation requirements.

juliusvonkohout: Here are some thoughts, but I still need to read the full document.

Tthat is very far away from V1 artifact passing via PVC and would violate zero overhead namespaces. Just imagine a cluster with 1000 namespaces for scalability. Then you add 1000 permanent pods, so massive overhead. In V1 we just accepted that the artifacts will not be in the UI instead. PVC for all namespace also sounds scary. Why should we offer such a security nightmare in the first place? That would break the namespace isolation contract.

Maybe I am missing something and I very much admire that you spend the time for the KEP, but there might be some fundamental problems as mentioned above regarding security and scalability that we did not have in the kfp V1 Implementation.

But as I said some of my statements could be wrong and that is just my initial assessment without checking it thoroughly.

hbelmiro: Got it. Those are really valid concerns for using it in production. But the idea of using PVCs is to make it easier for people who want to try KFP without having to worry about deploying and configuring an object storage. That means development (and mostly local) environments.
I didn't want to limit the usage only to development as it can be used on simple production environments.
We need to add clear documentation about this telling that the usage in production is not encouraged although not forbidden. Your concern is a good example that we can use to discourage using it in production.

mprahl:
@hbelmiro could you reopen this KEP in the KFP repo since this is KFP specific and not Kubeflow wide?

hbelmiro:
Done!
- https://github.com/kubeflow/pipelines/pull/12515

PR #929, opened 2025-12-05
Update adopters list in ADOPTERS.md

ma-armenta:
Added AT&T as an adopter and updated the list of organizations using Kubeflow.


ma-armenta:
I was just wondering that @andreyvelich, thanks!

Will do!

PR #937, opened 2026-01-28
Add KEP-936: Introduce Kubeflow-MCP proposal for AI-Powered Training …

### LINKED ISSUE #936: KEP: Kubeflow MCP Server - AI-Powered Training Interface
## Description

This KEP proposes adding a **Model Context Protocol (MCP) Server** to the Kubeflow SDK.

**Problem:** Using Kubeflow Training requires Python SDK expertise + Kubernetes knowledge. Users must understand `TrainerClient`, `BuiltinTrainer`, `CustomTrainer`, PVCs, RBAC, and multiple tools.

**Solution:** MCP server wraps the existing SDK, enabling AI agents (Claude, Cursor, VS Code Copilot) to interact with training through natural language—no code duplication.

**Related:** [kubeflow/sdk#238](https://github.com/kubeflow/sdk/issues/238)

abhijeet-dhumal:
Resolves #936  

## Summary

Introduces **KEP-936: Kubeflow MCP Server** - an AI-powered interface enabling LLM agents to interact with Kubeflow through the Model Context Protocol (MCP).

## Key Features

| Feature | Description |
|---------|-------------|
| **Dedicated Training Tools** | `fine_tune()`, `run_custom_training()`, `run_container_training()` mapping to SDK trainer types |
| **Modular Architecture** | `--clients trainer,optimizer,hub` for selective tool loading |
| **Persona-Based Access** | `readonly`, `data-scientist`, `ml-engineer`, `platform-admin` |
| **SDK Integration** | Wraps `TrainerClient`, `OptimizerClient`, `ModelRegistryClient` |

## Implementation Plan

| Phase | Scope | Tools |
|-------|-------|-------|
| 1. Core MCP Server | Trainer tools + mcp-tef validation | 16 |
| 2. Resource Management | Pre-flight checks, Mellea exploration | 17 |
| 3. Multi-Tenancy | Istio/impersonation, AGNTCY Identity | 17 |
| 4. Job Lifecycle | Suspend/resume, checkpoints | 19 |
| 5. Optimizer & Hub | Katib + Model Registry modules | 33 |
| 6. Future Clients | Pipelines, Spark (when SDK available) | TBD |

## Related

- Resolves #936
- SDK implementation tracking: kubeflow/sdk#238
- Trainer extensibility: kubeflow/trainer#2839

/cc @andreyvelich @astefanutti @franciscojavierarceo @juliusvonkohout

https://github.com/user-attachments/assets/49485bda-20d5-4d27-a90d-d10409e89d4a


proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,608 @@
...
+
+| Tool | SDK Mapping | Description |
+|------|-------------|-------------|
+| `delete_job(job_id)` | `TrainerClient.delete_job()` | Delete training job |
+| `suspend_job(job_id)` | K8s API patch | Pause job, free GPUs |
+| `resume_job(job_id)` | K8s API patch | Resume suspended job |
+
+### Trainer Selection Logic
+
+![Trainer Selection](trainer-selection.png)
```
astefanutti: I think it would be useful to link to kubeflow/trainer#2839 so it's extensible and support the future trainers that will be added.

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,608 @@
...
+| Risk | Mitigation |
+|------|------------|
+| **SDK Breaking Changes** | Pin SDK version, adapter pattern isolates changes |
+| **LLM Hallucination** | Pre-flight checks validate all inputs before execution |
+| **Resource Exhaustion** | Automatic resource estimation and user confirmation |
+| **Multi-tenancy** | Namespace isolation, uses user's kubeconfig RBAC |
+| **Unauthorized Access** | Policy layer enforces RBAC at tool level |
+| **Scope Creep** | Clear delegation to `kubernetes-mcp-server` for generic K8s ops |
+
+## Design Details
```
astefanutti: A section that covers the security aspects would be very valuable.

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,608 @@
...
+- Documentation complete
+
+
+## Drawbacks
+
+1. **Additional Dependency**: Users must install both SDK and MCP server
+2. **Maintenance Overhead**: MCP layer must track SDK changes
+3. **Abstraction Layer**: Natural language may hide complexity users need to understand
+
+## Alternatives
```
astefanutti, line 520: As skills are becoming popular, a dedicated alternative section could provide context on how it compares with the proposal.

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,608 @@
...
+## References
+
+### Core References
+- [Model Context Protocol Specification](https://modelcontextprotocol.io/)
+- [Kubeflow SDK Repository](https://github.com/kubeflow/sdk)
+- [Kubeflow Training Operator](https://github.com/kubeflow/trainer)
+- [KEP-2170: Kubeflow Trainer V2 API](https://github.com/kubeflow/trainer/blob/master/docs/proposals/2170-kubeflow-trainer-v2/README.md)
+
+### Related Issues
+- [#936: KEP Tracking Issue](https://github.com/kubeflow/community/issues/936) - This proposal
```
astefanutti: https://github.com/kubeflow/model-registry/pull/2029 would be added

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,608 @@
...
+| `get_training_job(job_id)` | `TrainerClient.get_job()` | Job details |
+| `list_training_runtimes()` | `TrainerClient.list_runtimes()` | Available ClusterTrainingRuntimes |
+| `get_runtime_spec(name)` | K8s API | Full runtime YAML |
+| `get_job_spec(job_id)` | K8s API | Full TrainJob YAML |
+
+#### Planning Tools (Validation)
+
+| Tool | Purpose | Returns |
+|------|---------|---------|
+| `estimate_resources(model, peft_method)` | GPU/memory requirements | `{gpu_memory, recommended_gpus, feasible}` |
```
astefanutti: Resource estimation may need to call different tools depending on the trainers.

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Thanks a lot @astefanutti for reviewing 🙌 
Awesome suggestion, I have added references to KEP-2839 accordingly ✅

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Added a dedicated "Security Considerations" section ✅

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Great catch 🙌 
I have added now "Alternative 4: Hugging Face Skills" with a detailed comparative analysis covering architecture, protocol, integration, and execution model differences. Also noted how Skills and MCP can be complementary, please review !

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Added "Note on Resource Estimation" with a table showing trainer-specific estimation methods (BuiltinTrainer uses model params, CustomTrainer uses heuristics, CustomTrainerContainer defers to user). Also noted extensibility for KEP-2839 backends. ✅

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,799 @@
...
+![Security Architecture](security.png)
+
+### Authentication
+
+The MCP server should inherit authentication from the underlying Kubernetes context:
+
+| Method | Description | Use Case |
+|--------|-------------|----------|
+| **Kubeconfig** | Uses `~/.kube/config` or `KUBECONFIG` env var | Local development, CI/CD |
+| **ServiceAccount Token** | Mounted at `/var/run/secrets/kubernetes.io/serviceaccount/token` | In-cluster deployment |
```
astefanutti: In the case of in-cluster deployment, should the MCP server rather impersonate users?

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,799 @@
...
+
+### Notes/Constraints/Caveats
+
+#### Multi-MCP Ecosystem
+
+Kubeflow MCP should be designed to work alongside other MCP servers, with clear scope boundaries:
+
+![Multi-MCP Ecosystem](multi-mcp.png)
+
+**Design Principle:** No overlap. `kubeflow-mcp` should handle training; `kubernetes-mcp-server` should handle generic K8s operations.
```
astefanutti: `kubeflow-mcp` would eventually handle more than training.

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,799 @@
...
+  "result": "success",
+  "job_id": "ft-qwen-abc123"
+}
+```
+
+**Note:** Sensitive parameters (tokens, credentials) should be **redacted** from logs.
+
+## Design Details
+
+### MCP Tool Inventory
```
astefanutti: Given the large number of tools, do you have an estimate of the size that it takes in the context based on your prototype?

Is there any concern w.r.t. to scaling the number of tools for all the Kubeflow components?

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Ah good point - I hadn't fully thought through the multi-user in-cluster scenario.

Please correct me here, So the flow would be:
1. MCP server runs with a ServiceAccount that has `impersonate` permissions
2. AI agent authenticates user and passes identity to MCP server
3. MCP server can use K8s impersonation (`--as` / `Impersonate-User` header) for API calls

This way RBAC is enforced per-user even with a shared MCP deployment.

Does this align with how you'd expect it to work? I'm curious if there's a standard pattern for this - perhaps similar to how the Notebooks controller handles user identity? 🤔

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Ah right - I was thinking the principle should be: kubeflow-mcp owns Kubeflow-specific CRDs (TrainJob, Experiment, ModelVersion, etc.), while `kubernetes-mcp-server` handles generic K8s resources (PVCs, ConfigMaps, Secrets).

So the row should probably say "Kubeflow CRDs" rather than "TrainJob CRDs"?

Or do you think there's value in having separate MCP servers per Kubeflow component (trainer-mcp, katib-mcp, model-registry-mcp)?
I see Model Registry is building its own MCP server (model-registry#2029). This raises a design question  🤔

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: From the prototype, ~24 tools with docstrings comes to roughly 8-10K tokens.

My intuition is that this is manageable for current MCP clients, but I'm curious about your perspective. A few questions:
1. Do you know if there's a recommended upper bound for MCP tool count? I haven't found guidance on this in the spec.
2. For full Kubeflow coverage (Trainer + Katib + Model Registry + Pipelines), we might hit ~40-50 tools. Would lazy loading (only expose tools for installed components) be a reasonable mitigation?
3. The other approach is policy-based filtering - a `readonly` user sees only ~7 discovery tools, not all 50. Does this feel like the right direction?
4. And the discussion from above thread, should we plan separate MCP servers for each kubeflow component ? I think it's good to keep everything under 1 roof as a Unified kubeflow-mcp simialr to kubeflow-sdk which can wrap all SDK clients, single install.. but I would be happy to get more discussion or inputs on this aspect 💭 

Happy to add a "Scalability Considerations" section if you think it's worth calling out explicitly.

cc: @andreyvelich @jaiakash @dhanishaphadate

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: I leaned toward unified since it mirrors the unified SDK structure, but curious what you think? Separate servers might give component teams more autonomy.

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: @astefanutti  @andreyvelich 
Please correct me here but what if we use Istio/OAuth2Proxy for this case  to inject user identity ? Kubeflow already uses Istio + OIDC right?

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: To address security constraints, I have added dedicated section with : 
- Authentication: Kubeconfig, ServiceAccount, ServiceAccount + Impersonation, OIDC
- Authorization: K8s RBAC flow, required verbs per resource
- Multi-User In-Cluster: Istio integration with x-forwarded-user header (see identity-flow diagram)
- Secret Management: HF tokens, S3 credentials handling
- Multi-Tenancy: Namespace isolation, policy enforcement, ResourceQuota validation
- Audit Logging: Structured JSON logs with redaction

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: To address this constraint, updated Authentication section with three deployment modes:

- ServiceAccount Token : Single-user in-cluster
-ServiceAccount + Impersonation : Multi-user in-cluster
- OIDC : Enterprise SSO

For multi-user, the flow uses Kubeflow's existing Istio layer:
1. User authenticates via Kubeflow dashboard
2. Istio Gateway validates JWT, adds x-forwarded-user header
3. MCP server reads header, impersonates user for K8s API calls

Also added RBAC example for the impersonator ClusterRole. 
This now aligns with how Kubeflow Notebooks handles user identity. ✅

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Added "Tool Scalability" section with detailed analysis:

Token estimates:
- Phase 1 (trainer only): 16 tools, nearly ~5.5K tokens
- Full stack (trainer + optimizer + hub): 30 tools, ~11K tokens

Scalability strategies:
- Modular Client Loading (--clients trainer)  : 40-70%
- Persona Filtering (--persona data-scientist) : 50-70%
- Combined : 70-85%

Research shows LLM accuracy degrades beyond 20-25 tools : 
1. With modular loading + personas, we stay within optimal range.
2. Also tracking [MCP Lazy Tool Hydration](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1978) proposal which could reduce overhead by 90%+ via deferred schema loading 
3. Added mcp-tef validation in each implementation phase to ensure tool descriptions remain distinguishable as we scale.

abhijeet-dhumal: Pleas review if this approach seems correct? 
cc: @astefanutti @andreyvelich

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,419 @@
...
+- [Implementation Plan](#implementation-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References](#references)
+
+---
+
+## Summary
+
+This KEP proposes a **Model Context Protocol (MCP) Server** for the Kubeflow SDK that enables AI agents to interact with Kubeflow Training resources through natural language. The MCP server wraps the existing Kubeflow SDK (`TrainerClient`, `BuiltinTrainer`, `CustomTrainer`) without duplicating code.
```
astefanutti:
```suggestion
This KEP proposes a **Model Context Protocol (MCP) Server** for the Kubeflow SDK that enables AI agents to interact with Kubeflow Trainer resources through natural language. The MCP server wraps the existing Kubeflow SDK (`TrainerClient`, `BuiltinTrainer`, `CustomTrainer`) without duplicating code.
```

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+**Core Principle:** The MCP server is a *complementary interface*, not a replacement. It wraps the SDK, enabling natural language workflows while preserving full programmatic access.
+
+### Before vs After
+
+![Before vs After MCP](assets/before-after.png)
+
+## Motivation
+
+Kubeflow Training provides powerful distributed training capabilities, but requires Python SDK knowledge, Kubernetes expertise, and manual validation. The [Model Context Protocol (MCP)](https://modelcontextprotocol.io/) provides a standard way for AI agents to interact with external systems, supported by modern AI-powered IDEs (Claude Code, Cursor, VS Code with Copilot).
```
astefanutti:
```suggestion
Kubeflow Trainer provides powerful distributed training capabilities, but requires Python SDK knowledge, Kubernetes expertise, and manual validation. The [Model Context Protocol (MCP)](https://modelcontextprotocol.io/) provides a standard way for AI agents to interact with external systems, supported by modern AI-powered IDEs (Claude Code, Cursor, VS Code with Copilot).
```

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+![Policies](assets/policies.png)
+
+| Persona | Discovery | Training | Lifecycle | Runtimes |
+|---------|-----------|----------|-----------|----------|
+| `readonly` | Yes | No | No | No |
+| `data-scientist` | Yes | Yes | Yes (own jobs) | No |
+| `ml-engineer` | Yes | Yes | Yes | Yes |
+
+### Package Structure
```
astefanutti: Maybe we can keep it in the design doc.

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,419 @@
...
+| **Build Reasoning into Tools** | Tools provide DATA + ACTIONS; LLMs provide REASONING |
+| **Custom Protocol** | MCP is an open standard with multi-vendor support |
+| **HuggingFace Skills** | Instruction-based; lacks execution, RBAC, namespace isolation |
+
+## References
+
+### Core
+- [Model Context Protocol Specification](https://modelcontextprotocol.io/)
+- [Kubeflow SDK Repository](https://github.com/kubeflow/sdk)
+- [Kubeflow Training Operator](https://github.com/kubeflow/trainer)
```
astefanutti:
```suggestion
- [Kubeflow Trainer](https://github.com/kubeflow/trainer)
```

astefanutti (commented):
I left some small comments, otherwise looks good.

astefanutti (commented):
Thanks @abhijeet-dhumal!

/lgtm

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+|--------|-------|----------------|
+| **optimizer** | 8 | `OptimizerClient` |
+| **hub** | 6 | `ModelRegistryClient` |
+
+- Optional `--mode dynamic` for 33+ tools ([100x token reduction](https://www.speakeasy.com/blog/100x-token-reduction-dynamic-toolsets) via semantic search)
+
+### Phase 6: Future Modules
+- `pipelines/` - `PipelinesClient`
+- `spark/` - `SparkClient`
+- `feast/` - `FeastClient` ([#239](https://github.com/kubeflow/sdk/issues/239))
```
kramaranya, line 512: Feast already supports MCP - https://github.com/feast-dev/feast/issues/5404
Did you have a chance to investigate how this would work with Kubeflow MCP?

proposals/936-kubeflow-mcp-server/README.md
andreyvelich: IIUC, the goal of https://github.com/kubeflow/model-registry/pull/2029 is to provide searchable catalog of MCP servers that users can install via Hub UI. @ederign @kubeflow/kubeflow-hub-team please correct me if I miss-understood it.


>Or do you think there's value in having separate MCP servers per Kubeflow component (trainer-mcp, katib-mcp, model-registry-mcp)?

This is good question, we should talk more about it. Since each project has a different scope and set of requirements, a broad range of tools is often necessary to support end-to-end LLMOps. However, exposing too many tools directly to an agent can quickly overwhelm its context window, leading to degraded performance and less reliable reasoning

As a workaround, we can think of introducing special Agent Skills: https://agentskills.io/home, which expose sub-set of MCP tools agent can talk to. These skills can act as curated capability bundles, exposing only a relevant subset of MCP tools to the agent

Similar to this diagram: https://github.com/anthropics/claude-agent-sdk-python/issues/544


Or we can design sub-agents which can be Trainer or Spark experts.

proposals/936-kubeflow-mcp-server/README.md
andreyvelich: I think, the long term approach would be to build specialized agent personas to reduce context.
For example, if I am only interested in PyTorch fine-tuning, I don't need Spark tools.

andreyvelich:
> Thanks for the proposal :) For the mcp server, I’d like to better understand the motivation for choosing the **SDK-wrapping architecture** instead of direct API interaction.
> 
> If the kubeflow-mcp server consider another way to interact with k8s or kubeflow controller, then I think it doesn’t have to be coupled with sdk wg?

Good question, @anencore94. I think, building on top of Kubeflow SDK has a few advantages:

- As as @szaher proposed in https://github.com/kubeflow/sdk/issues/281, the SDK can provide a consistent auth layer across Kubeflow APIs. Since not all resources are accessible via `kubectl` (e.g., Trainer/Optimizer/Spark use CRs, while KFP and Model Registry have separate API servers), a unified interface simplifies client authorization and agent interactions.
- First-class [local execution](https://github.com/kubeflow/sdk/tree/main/docs/proposals/2-trainer-local-execution), cc @Fiona-Waters. The SDK enables local prototyping before submitting jobs to a Kubernetes cluster. Exposing this to agents improves iteration speed and reduces cluster dependency.
- [Builtin](https://www.kubeflow.org/docs/components/trainer/user-guides/builtin-trainer/torchtune/) and specialized Trainers – https://github.com/kubeflow/sdk/pull/308, we’re improving the TrainJob experience within the SDK. Because this workflow is tightly coupled with Python orchestration, it is harder to replicate cleanly through TrainJob APIs alone.

Does it make sense?

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+- [Implementation Plan](#implementation-plan)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [References](#references)
+
+---
+
+## Summary
+
+This KEP proposes a **Model Context Protocol (MCP) Server** for the Kubeflow SDK that enables AI agents to interact with Kubeflow Trainer resources through natural language. The MCP server wraps the existing Kubeflow SDK (`TrainerClient`, `BuiltinTrainer`, `CustomTrainer`) without duplicating code.
```
andreyvelich: Can we explain that we will start with TrainerClient, but our long-term goal is to support end-to-end AI-Powered LLMOps?

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+
+### Alignment with Unified Kubeflow SDK
+
+The MCP server evolves alongside the unified Kubeflow SDK (`kubeflow/sdk`):
+
+| SDK Client | Component | MCP Integration |
+|------------|-----------|-----------------|
+| `TrainerClient` | Kubeflow Trainer | Phase 1 (this proposal) |
+| `OptimizerClient` | Kubeflow Katib | Phase 5 |
+| `ModelRegistryClient` | Model Registry | Phase 5 |
```
andreyvelich, line 106: Add SparkClient.
cc @Shekharrajak @tariq-hasan

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+| `PipelinesClient` | Kubeflow Pipelines | Future |
+
+![Unified SDK Architecture](assets/unified-sdk.png)
```
andreyvelich, line 110: It should be Trainer instead of Training Operator in this diagram.

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+
+**Story 1: Data Scientist Fine-Tuning**
+
+```
+User: "Fine-tune Qwen/Qwen2.5-7B-Instruct on tatsu-lab/alpaca using LoRA"
+
+AI Agent (using MCP tools):
+1. get_cluster_resources() - "4x A100 80GB available"
+2. estimate_resources("Qwen/Qwen2.5-7B-Instruct", "lora") - "24GB needed"
+3. fine_tune(model="...", dataset="...", peft_method="lora", confirmed=True)
```
andreyvelich, line 153: Why do we need to expose all of those tools via MCP if users just want to fine-tune LLM?
Shall we wrap all orchestration logic under the single tool?
Can we explore how HF implement this, do they just offer set of Skills to fine-tune models?
https://huggingface.co/blog/hf-skills-training

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+                                  confirmed=True)
+```
+
+## Design Details
+
+### MCP Tool Inventory
+
+![Tool Layers](assets/tool-layers.png)
+
+Tools are organized in layers aligned with SDK structure (16 tools in Phase 1):
```
andreyvelich, line 196: Shall we reduce number of tools we expose in the first version to not overload agent context?
For example, we don't really need to have `get` calls, since we have `list`.

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+| **Planning** | `estimate_resources()` | Memory estimation |
+| **Training** | `fine_tune()`, `run_custom_training()`, `run_container_training()` | Job submission |
+| **Discovery** | `list_training_jobs()`, `get_training_job()`, `list_runtimes()`, `get_runtime()`, `get_runtime_packages()` | Resource lookup |
+| **Monitoring** | `get_training_logs()`, `get_training_events()`, `wait_for_training()` | Job monitoring |
+| **Lifecycle** | `delete_training_job()`, `suspend_training_job()`, `resume_training_job()` | Job management |
+
+**Dedicated Training Tools:**
+- `fine_tune()` maps to `BuiltinTrainer` (zero-code LLM fine-tuning with TorchTune)
+- `run_custom_training()` maps to `CustomTrainer` (user-provided Python code)
+- `run_container_training()` maps to `CustomTrainerContainer` (pre-built container)
```
andreyvelich, line 210: We need to think on this a bit more. For example, users can ask agent to create PyTorch script to fine-tune  LLM, and trigger TrainJob. That will require usage of `CustomTrainer()`

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+| **Pod debugging (exec, logs)** | Delegates | Owns |
+
+### Tool Scalability
+
+Research shows LLM accuracy degrades beyond 20-25 tools ([ToolScope](https://arxiv.org/abs/2510.20036)). The modular architecture addresses this:
+
+| Strategy | Mechanism |
+|----------|-----------|
+| **Modular Client Loading** | `--clients trainer` loads only trainer tools |
+| **Persona Filtering** | `--persona data-scientist` hides admin tools |
```
andreyvelich, line 242: Can you explain how are we going to do persona filtering?

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+| `readonly` | Yes | No | No | No |
+| `data-scientist` | Yes | Yes | Yes (own jobs) | No |
+| `ml-engineer` | Yes | Yes | Yes | Yes |
+
+### CLI Usage
+
+```bash
+kubeflow-mcp serve --clients trainer --persona data-scientist
+kubeflow-mcp clients list
+```
```
andreyvelich, line 296: Do we expect this MCP Servers to serve locally? How users can deploy it on the server?

proposals/936-kubeflow-mcp-server/README.md
```diff
@@ -0,0 +1,402 @@
...
+
+### Before vs After
+
+![Before vs After MCP](assets/before-after.png)
+
+## Motivation
+
+Kubeflow Trainer provides powerful distributed training capabilities, but requires Python SDK knowledge, Kubernetes expertise, and manual validation. The [Model Context Protocol (MCP)](https://modelcontextprotocol.io/) provides a standard way for AI agents to interact with external systems, supported by modern AI-powered IDEs (Claude Code, Cursor, VS Code with Copilot).
+
+### Goals
```
andreyvelich, line 79: Do we also want to build Claude Plugin for integrations: https://code.claude.com/docs/en/plugins#plugin-structure-overview ?
For example: https://github.com/obra/superpowers/tree/main

andreyvelich (commented):
Thanks @abhijeet-dhumal!
I left a few comments, overall looks great.

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Hi @kramaranya , Actually I saw this before but didn't get much time.. 
Let me take a look thoroughly 👀

abhijeet-dhumal: I tried investigating both Feast MCP and Model Registry MCP Catalog:

- Feast MCP 1 exposes get_online_features for feature retrieval - completely different domain. They serve features for training data prep, we execute training. I think of it like a complementary, no overlap.

- I have added feast-mcp column just now to Multi-MCP Ecosystem table and coordination notes in the proposal.

For future: I think if an LLMOps workflow needs both feature prep and training, the AI agent would call Feast MCP for features, then kubeflow-mcp for training - each server owns its domain.

abhijeet-dhumal: Let me know if answers the question, happy to discuss it further.. Thanks !

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Done, I have made updates accordingly!

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Great pointers! I looked into Agent Skills and the sub-agent pattern. Here's my take:
-  Agent Skills - These are essentially curated tool bundles. Our modular client loading (--clients trainer) and persona filtering (--persona data-scientist) achieve similar scoping at the MCP level. The difference is we filter at server startup rather than requiring an intermediary skill layer.

- Sub-agents - Interesting pattern for complex multi-step workflows. For Phase 1, we're keeping tools granular so the primary agent (Claude/Cursor) can orchestrate. If workflows get complex enough to need dedicated "Trainer expert" sub-agents, i think that's more of a client-side concern - the MCP server just needs to expose the right tool subsets, which we support via --clients.

That said, I added a note in Design Decisions about Speakeasy's dynamic toolsets pattern for Phase 5+ (33+ tools) - meta-tools like "search_tools", "describe_tool", "execute_tool" that let the LLM discover capabilities without loading all schemas upfront.

abhijeet-dhumal: Open to discussing further if you think we should explicitly support Agent Skills integration!

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Exactly what I have added in proposal ! The proposal includes:

| Persona | Tools | Use Case |
|---------|-------|----------|
| readonly | 7 | Monitoring |
| data-scientist | 12 | Fine-tuning |
| ml-engineer | 16 | Full training |

Combined with --clients trainer (only load trainer tools), a PyTorch fine-tuning user never sees Spark tools. The 70-85% reduction keeps us in the optimal 10-15 tool range per research.

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Added! Summary now reads: "Phase 1 focuses on TrainerClient for distributed training and LLM fine-tuning; the long-term goal is AI-powered end-to-end LLMOps covering training, hyperparameter optimization, model registry, and pipelines." ✅

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: There are many reasons for following this analogy, I have tried listing them in a "Design Decisions" section explaining this - Please review !

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Fair point on get vs list. The reason for separate tools:
- list_training_jobs() returns summaries (name, status, created_at) for browsing
- get_training_job(name) returns full spec (config, resources, conditions) for inspection
- Combined list with full details would bloat responses when users just want "show me running jobs."

That said, happy to revisit if prototype shows agents don't need both. We can always start with fewer tools and add based on real usage. Current Phase 1 has 16 - with persona filtering (--persona data-scientist) that drops to 12

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Exactly - just now I added "Story 4: Agent-Generated Training Code" to capture this.. please review!

The func_code serialization + AST security checks in DESIGN.md handle the code-as-string transport safely.

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Persona filtering works at server startup via CLI flag:
```
kubeflow-mcp --persona data-scientist
```
The server can register only tools matching that persona's scope:
| Persona | Tools | What's hidden |
|---------|-------|---------------|
| readonly | 7 | All write operations |
| data-scientist | 12 | Container tools, admin ops |
| ml-engineer | 16 | Nothing hidden |

Implementation is a simple decorator filter at registration time - tools declare their required persona level, and we skip registration if current persona is lower. Please check DESIGN.md for the pattern.

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Both! I added a Deployment Modes table to clarify:

| Mode | Location | Auth | Transport |
|------|----------|------|-----------|
| Local | User's laptop | Kubeconfig | stdio |
| In-cluster | K8s Pod | ServiceAccount + Impersonation | StreamableHTTP |
| Gateway | Behind MCP gateway | OAuth/OIDC | StreamableHTTP |

Local is simplest (like kubectl - uses your kubeconfig). In-cluster needs impersonation for multi-user RBAC. Gateway mode for enterprise with existing auth infra.

For in-cluster, deployment would be a standard K8s Deployment + Service, similar to how Kubeflow dashboard is deployed.

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Interesting idea! Looking at the Claude Plugin structure, plugins are essentially MCP servers with additional metadata (manifest, permissions).
Since kubeflow-mcp is an MCP server, making it a Claude Plugin would be minimal extra work - mainly adding a manifest.json with capabilities declaration.

abhijeet-dhumal: I'd suggest: Phase 1 focuses on core MCP server, then packaging as Claude Plugin can be a Phase 2 enhancement once the tool surface is stable. Seems like a good plan ?
Please correct me if wrong and Happy to discuss it further :)

proposals/936-kubeflow-mcp-server/README.md
Sanskarzz: Great discussion! I'd like to offer a different angle on the tool-count concern.

Rather than hard-capping the number of tools through personas or modular loading, I think we should be careful not to artificially constrain the API surface of kubeflow-mcp, especially as the SDK grows and new components get added. Limiting tools solves today's problem but creates friction tomorrow.

Instead, we can delegate the context-optimization problem to the MCP layer itself. There's an open-source project called [mcp-optimizer](https://github.com/StacklokLabs/mcp-optimizer) that tackles exactly this, it dynamically prunes and prioritizes tools based on the current task/query, so the LLM only sees the relevant subset at inference time, without us having to bake filtering logic into the server.

I think there are two paths worth considering:
1. **Integrate mcp-optimizer-style logic directly into kubeflow-mcp** — so the server itself handles intelligent tool surfacing.
2. **Provide official documentation/guidance** on how to use mcp-optimizer (or similar middleware) alongside kubeflow-mcp for users who hit context limits.

This way, the server stays complete and unopinionated, and the optimization lives at the right layer. Happy to explore either direction.

proposals/936-kubeflow-mcp-server/README.md
andreyvelich: Sounds good, any thoughts on how local agents (e.g. Claude) can configure MCP server with the desired scope (e.g. `--clients trainer`) ?
Are we going to provide this argument as an entrypoint to start the FastAPI server?

proposals/936-kubeflow-mcp-server/README.md
andreyvelich: That's interesting, thanks for sharing @Sanskarzz!
What are the main benefits of delegating tool selection to an `mcp-optimizer`, instead of managing access through multiple sub-agents that statically define Skills that have access to MCP tools by defining them in `allowed-tools` section?

proposals/936-kubeflow-mcp-server/README.md
andreyvelich: Are we planning to change the `list_jobs()` API in the MCP Server?
As you can see, currently it returns `list[TrainJobs]`
https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L146

proposals/936-kubeflow-mcp-server/README.md
andreyvelich: Sounds good, let's add it to the plan.

sjiang83:
> Hi folks — just wanted to chime in here with a GSoC applicant's perspective, specifically re: what added value kubeflow-mcp provides beyond an OpenAPI→MCP generator, and how we might scope Phase 1.
> 
> My two cents on auto-generating tools: while exposing K8s endpoints is easy, the durable value of kubeflow-mcp really lives in workflow-level, SDK-backed abstractions. Agents generally struggle to stitch together raw, low-level APIs. By wrapping the SDK, we can keep a unified UX across clients (consistent auth surface, BuiltinTrainer, local iteration paths) and inject agent-specific guardrails like pre-flight checks, explicit confirmations, and error summaries. A generic 1:1 generator won’t give the LLM that kind of reliability.
> 
> Regarding the list_jobs() payload size discussion — I agree we shouldn't change the underlying SDK APIs just to accommodate LLM context windows. It makes more sense to adapt this right at the MCP tool layer:
> 
> * list_training_jobs() can return a lean summary (name, status, namespace, created_at)
> * if the agent needs details, it can call get_training_job(name) for the full spec
> 
> For tool count, keeping Phase 1 strictly statically scoped (e.g., --clients trainer + persona) feels like the safest bet: predictable and easy to validate. We can revisit dynamic pruning (e.g., mcp-optimizer-style middleware) or sub-agent routing later (Phase 5+) when we actually hit the 30+ tool scale — let’s not over-engineer it upfront.
> 
> Lastly, on script_code security: I think we just need to be super upfront about limitations. AST checks are great as a best-effort guardrail, but in Python there will be bypass vectors (e.g., importlib indirection). If untrusted code is part of the threat model, we should explicitly recommend run_container_training() as the safer path.

Just catching up on the recent thread, and I wanted to propose a practical boundary for Phase 1 based on the discussion.

I think it makes complete sense for kubeflow-mcp to leverage the SDK when dealing with high-level abstractions, like BuiltinTrainer, zero-code flows, unified auth, and local iteration paths. At the same time, we don't have to force everything through the SDK. If certain tools are better handled via direct Kubernetes access or by integrating kubernetes-mcp-server as a plugin, we should absolutely take that route.

How do the maintainers feel about scoping Phase 1 like this?

In Scope (Phase 1): Focusing on Trainer tools using SDK wrappers. We should also heavily prioritize making things easy for AI agents—think pre-flight checks, explicit confirmations for risky actions, summarized outputs, and clear error reporting.

Out of Scope (Phase 1): We should avoid reinventing generic Kubernetes operations like exec, logs, and debug commands inside kubeflow-mcp. We can just delegate those to the existing K8s MCP server where it fits.

Future: Once the core foundation is stable, we can start exploring skills-based orchestration and toolsets that admins can configure dynamically.

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: No SDK API changes needed. I'm planning it like MCP tool layer can handle serialization:
- list_training_jobs() calls TrainerClient.list_jobs(), extracts summary fields (name, status, namespace, created_at)
- get_training_job(name) returns full TrainJob spec

SDK will return list[TrainJob], we will transform at the MCP layer to keep agent context lean, WDYT ?

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Yes, via entrypoint args. 
Example for Claude Desktop (~/.claude/claude_desktop_config.json):
```
{
   "mcpServers": {
     "kubeflow": {
       "command": "kubeflow-mcp",
       "args": ["serve", "--clients", "trainer", "--persona", "data-scientist"]
     }
   }
 }
```

proposals/936-kubeflow-mcp-server/README.md
abhijeet-dhumal: Great point @Sanskarzz! I agree we shouldn't artificially constrain the API surface.
The way I see it, we have two complementary strategies:

1. Static filtering (--clients, --persona) - deterministic, predictable, zero overhead. Good for: "I'm a data scientist, I only need training tools." Admin sets it once at deployment.
2. Dynamic pruning (mcp-optimizer style) - intelligent, query-aware. Good for: "I have 50 tools loaded but this query only needs 5." Adds latency but handles the long tail.

For Phase 1-4, static filtering keeps us in the 10-16 tool range which works well. For Phase 5+ (33+ tools), I think you're right that dynamic approaches become necessary.
I'd suggest:
1. Phase 1-4: Ship with --clients + --persona (already in proposal)
2. Phase 5: Add --mode dynamic option that enables semantic tool discovery (the Speakeasy pattern I mentioned)
3. Document mcp-optimizer as a recommended middleware for users who want external optimization

This way the server stays complete and unopinionated (as you said), and optimization can happen at whichever layer makes sense for the deployment.
Does this approach work? Happy to add explicit guidance on mcp-optimizer integration to the docs.

thesuperzapper:
I am ok with creation of new repo with the this project with the [WG ML Experience](https://github.com/kubeflow/community/tree/master/wg-ml-experience) leads as the root OWNERS, as long as we clearly mark it as experimental in the readme, and say that its not intended for production usage until we make further statements.

We should defer the acceptance of this KEP (which talks about implementation details of the MCP server) to the WG ML Experience leads to have final vote on as they will own that new repo.

proposals/936-kubeflow-mcp-server/DESIGN.md
```diff
@@ -0,0 +1,824 @@
...
+```python
+from kubernetes.client.exceptions import ApiException
+from kubeflow.trainer import (
+    TrainerClient, BuiltinTrainer, CustomTrainer, CustomTrainerContainer,
+    TorchTuneConfig, LoraConfig, Initializer,
+    HuggingFaceModelInitializer, HuggingFaceDatasetInitializer,
+)
+
+@mcp.tool()
+def fine_tune(
```
andreyvelich, line 44: I think, we should consolidate all Create TrainJob calls to a single MCP tool which might allow agents to use CustomTrainer or BuiltinTrainer. That will reduce context.
Not a blocker, but we can start with some simple flows and interate.

cc @kubeflow/kubeflow-sdk-team

andreyvelich (commented):
Thanks @abhijeet-dhumal, overall lgtm, let's continue the design discussion in the future meetings, but we can at least unblock the initial work.
/lgtm
/assign @chasecadet @thesuperzapper @franciscojavierarceo @juliusvonkohout

proposals/936-kubeflow-mcp-server/DESIGN.md
abhijeet-dhumal: Thanks @andreyvelich! The separation was intentional !
A unified tool would need a complex union schema with conditional required fields, which LLMs struggle to navigate correctly.
I have already included this concern in design decision section and described reasoning in detail : https://github.com/kubeflow/community/pull/937/changes#diff-6088d3fe5c7209fb9d22f384545afe417304387ddf7a0c0cb7d962b5d704f746R353

proposals/936-kubeflow-mcp-server/DESIGN.md
```diff
@@ -0,0 +1,824 @@
...
+        with tempfile.NamedTemporaryFile(
+            mode='w', suffix='.py', prefix=f'mcp_train_{script_hash}_', delete=False
+        ) as f:
+            f.write(func_code)
+            temp_path = f.name
+        
+        spec = importlib.util.spec_from_file_location("training_module", temp_path)
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
+        train_func = module.train
```
Sanskarzz, line 141: One thing worth clarifying in the security model here: `spec.loader.exec_module(module)` actually executes any **module-level code** in `func_code` on the **MCP server process itself**, not just inside the K8s pod. The pod only receives the extracted `train` function later. So if `func_code` contains something like:

```python
x = open("/etc/passwd").read()   # runs on MCP server at load time!
def train(**kwargs):
    pass
```

…the `open()` call runs before any K8s pod is created.

The Security Note says "The func_code is executed within K8s pods, not on the MCP server host" — this is accurate for the `train()` function body, but not for module-level code. I think it's worth adding a clarifying sentence like:

> "Module-level statements in `func_code` execute on the MCP server host at load time. Only the `train()` function body is serialized and sent to the K8s worker pod."

The AST checks catch a lot of the obvious cases, but as a defense-in-depth note this distinction matters — especially for a shared in-cluster/gateway deployment

proposals/936-kubeflow-mcp-server/DESIGN.md
```diff
@@ -0,0 +1,824 @@
...
+            model=HuggingFaceModelInitializer(storage_uri=f"hf://{model}"),
+            dataset=HuggingFaceDatasetInitializer(storage_uri=f"hf://{dataset}"),
+        )
+        job_name = client.train(runtime=runtime, trainer=trainer, initializer=initializer)
+        return {"success": True, "job_id": job_name, "trainer_type": "BuiltinTrainer"}
+    except ApiException as e:
+        return {"success": False, "error": f"Kubernetes API error: {e.reason}", "status_code": e.status}
+    except Exception as e:
+        return {"success": False, "error": str(e)}
+
```
Sanskarzz, line 84: Small but potentially important gap: each tool does `client = TrainerClient()` inline with no arguments, which defaults to `KubernetesBackendConfig()` and re-reads kubeconfig every call. 

For the multi-user in-cluster scenario (where `auth.py`'s `create_k8s_client_for_user()` creates an impersonated `kubernetes.client.ApiClient`), there's currently no shown path to pass that into `TrainerClient`. The SDK's constructor signature accepts:

```python
TrainerClient(backend_config: KubernetesBackendConfig | ...)
```

Could we add a design note showing how the impersonated client gets wired in? Something like:

```python
# in each tool, after auth:
backend_config = KubernetesBackendConfig(
    client=create_k8s_client_for_user(user_identity),
    namespace=user_identity["namespace"],
)
client = TrainerClient(backend_config=backend_config)
```

Without this, the per-user namespace isolation and impersonation described in the auth section doesn't actually connect to the tool execution path.

proposals/936-kubeflow-mcp-server/DESIGN.md
```diff
@@ -0,0 +1,824 @@
...
+```python
+def estimate_resources(
+    model: str, 
+    peft_method: str,
+    batch_size: int = 4,
+    sequence_length: int = 2048,
+    quantization: str = "bf16",  # "fp32", "bf16", "fp16", "int8", "int4"
+    num_nodes: int = 1,
+) -> dict:
```
Sanskarzz, line 216: The `get_model_info(model)` call hits HuggingFace Hub API synchronously in the tool hot path. A few concerns:

1. **No timeout** — if HF Hub is slow/unreachable, this blocks indefinitely.
2. **No caching** — same model queried repeatedly on every `estimate_resources()` call.
3. **Air-gapped environments** — enterprise Kubeflow deployments often have no external internet access. The tool would always return `"confidence": "low"` and be essentially useless.

The `user_provided_params` approach was already discussed in the thread (great suggestion from @MansiSingh17, agreed to by @abhijeet-dhumal for Phase 2), but it's not yet reflected in the algorithm code. Could we update the pseudocode snippet to show:

```python
# Prefer user-provided params for private/air-gapped models
if user_provided_params:
    param_count = user_provided_params.get("param_count")
    ...
else:
    model_info = get_model_info(model, timeout=10)  # explicit timeout
```

Even a TODO comment would help set expectations for implementors.

proposals/936-kubeflow-mcp-server/DESIGN.md
```diff
@@ -0,0 +1,824 @@
...
+```python
+@mcp.tool()
+def wait_for_training(
+    job_id: str,
+    timeout: int = 3600,      # Default: 1 hour
+    poll_interval: int = 30,
+) -> dict:
+    """Returns current status on timeout (doesn't cancel job)."""
+```
+
```
Sanskarzz, line 459: The default `timeout=3600` (1 hour) will be problematic for the Streamable HTTP transport. Most reverse proxies (nginx default: 60s, AWS ALB: 60s, GCP: 30 min) will terminate idle connections well before 1 hour. The SDK's own `wait_for_job_status()` defaults to 600s.

One option: keep the long timeout for `stdio` mode (Claude Desktop/Cursor) but document that for HTTP deployments the recommended pattern is polling via `get_training_job()` at intervals, rather than holding a long-lived connection open. Could add a transport note here:

> "For StreamableHTTP deployments, prefer using `get_training_job()` polling over `wait_for_training()` to avoid proxy timeout issues."

Sanskarzz (commented):
Thanks for working through this proposal so thoroughly @abhijeet-dhumal  — really impressive.
I've read through the full proposal, DESIGN.md, and the prior discussion, and I have a few specific comments below that I think are worth addressing before or during implementation.

proposals/936-kubeflow-mcp-server/DESIGN.md
abhijeet-dhumal: You're right! exec_module() does execute module-level code on the MCP server before the train function is extracted and sent to K8s.
The AST checks (Phase 1) catch obvious cases like open(), subprocess,..., but as you note - defense-in-depth matters for shared gateway deployments.
I'll update the security note during implementation to clarify:
- Module-level statements execute on MCP server at load time
- Only the train() function body is serialised to K8s worker pods
Thanks for the thorough review!

proposals/936-kubeflow-mcp-server/DESIGN.md
abhijeet-dhumal: Valid point - here the pseudocode shows TrainerClient() with no args, which doesn't connect to the impersonation path in auth.py. The implementation will wire this properly 👍

abhijeet-dhumal: I'll add this pattern to the implementation docs. The KEP pseudocode is intentionally simplified for readability..

proposals/936-kubeflow-mcp-server/DESIGN.md
abhijeet-dhumal: Good point on transport differences. The 1-hour default makes sense for stdio (Claude Desktop, Cursor) but breaks behind reverse proxies. 
but again .. . The KEP pseudocode is intentionally simplified for readability..
We will document the recommended pattern for each transport mode 👍

proposals/936-kubeflow-mcp-server/DESIGN.md
abhijeet-dhumal: When HF Hub is unreachable, we return "confidence": "low" with a message suggesting user_provided_params. This is already the intent - I'll make sure the implementation pseudocode reflects it.

For enterprise deployments, we could also support a local model registry endpoint as an alternative to HF Hub.

proposals/936-kubeflow-mcp-server/DESIGN.md
abhijeet-dhumal: @Sanskarzz  I have added a security note accordingly !

andreyvelich (commented):
/lgtm
/approve
/hold cancel

PR #945, opened 2026-03-12
KEP-938: Create kubeflow/mcp-server repository

abhijeet-dhumal:
This KEP requests the creation of kubeflow/mcp-server repository to host the Kubeflow MCP (Model Context Protocol) Server.
This is a repo-creation KEP focused on governance and ownership as discussed [here](https://github.com/kubeflow/community/pull/937#issuecomment-4041451434). Technical implementation details are covered in [KEP-936](https://github.com/kubeflow/community/pull/937).

cc: @thesuperzapper @andreyvelich @jaiakash

### DIFF
--- a/proposals/938-mcp-server-repo/README.md
+++ b/proposals/938-mcp-server-repo/README.md
@@ -1,0 +1,92 @@
+# KEP-938: Dedicated Repository for  kubeflow/mcp-server Repository
+
+**Authors:**
+- Abhijeet Dhumal (Red Hat) - [@abhijeet-dhumal](https://github.com/abhijeet-dhumal)
+
+**Status:** Provisional
+
+**Related:**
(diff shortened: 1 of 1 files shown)


thesuperzapper:
While I agree we should do the things proposed in this kep, its unclear if we need to merge it as it was already done as part of https://github.com/kubeflow/community/pull/937 

cc @andreyvelich ?

but either way adding a status notice is a good idea (tip, you can do something like what we did in the [notebooks v2 branch](https://github.com/kubeflow/notebooks/tree/notebooks-v2) which uses github's new warning banners)

andreyvelich:
Yes, as we discussed in Slack, we can keep the ownership section in the original KEP: https://github.com/kubeflow/community/tree/master/proposals/936-kubeflow-mcp-server
/close

PR #953, opened 2026-04-29
Fix Trainer ADOPTERS link to use kubeflow/trainer repo

kfaseela:


andreyvelich (commented):
Thanks for fixing this @kfaseela!
/lgtm
/approve

PR #986, opened 2026-06-16
fix wg-data info

nabuskey:
Updated wgs.yaml and re-generate docs.


nabuskey:
@andreyvelich @vara-bonthu

andreyvelich (commented):
Thanks @nabuskey!
/lgtm
/approve

PR #1001, opened 2026-06-19
chore: Add per-WG `OWNERS` files

christian-heusel:
This aims to add WG Leads as reviewers for their charters so that they
are aware of changes concerning their charter and readme.

This is in line with the process as outlined in `committee-steering/wg-governance.md`:

> For minor updates that only impact issues or areas within the scope of the WG, the WG Chairs _SHOULD_ facilitate the change.

### DIFF
--- a/wg-data/OWNERS
+++ b/wg-data/OWNERS
@@ -1,0 +1,2 @@
+reviewers:
+  - wg-data-leads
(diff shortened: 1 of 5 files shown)


christian-heusel (commented):
For this to really make sense we might also want to look at enabling the prow setting that assigns all reviewers instead of just two (at random)

Issue #950, opened 2026-04-19
Transfer Ownership to KSC 2026


chasecadet:
GitHub issue content:

[ ] -     Update Kubeflow Steering Committee document with the new members and emeritus members.
[ ] -     Archive the current Slack channel (e.g. #archived-ksc-2024) and create the new Slack channel (e.g. kubeflow-steering-committee).
[x] -     Schedule weekly calls with the new members.
[ ] -     Update admins for Kubeflow GitHub org.
[ ] -     Update the kubeflow-steering-committee GitHub team.
[ ] -     Update approvers for the following OWNERS files (e.g the past members should be moved to emeritus_approvers):
[ ] -         kubeflow/kubeflow OWNERS file.
[ ] -         kubeflow/community OWNERS file.
[ ] -         kubeflow/internal-acls OWNERS file.
[ ] -         kubeflow/website OWNERS file.
[ ] -         kubeflow/blog OWNERS file.
[ ] -     Kubeflow GCP projects under kubeflow.org organization for ACLs and DNS management.
[ ] -         Access for kf-admin-cluster GKE cluster in kubeflow-admin GCP project for the GitHub ACLs sync.
[ ] -         Access for kubeflow-dns GCP project for the DNS management.
[ ] -     Access for Kubeflow GKE cluster kf-ci-v1 in kubeflow-ci GCP project (No Organization) where Prow is running.
[ ] -     Kubeflow Google Group.
[ ] -     Update members for KSC Google Group.
[ ] -     Access to Kubeflow 1password account.
[ ] -     Kubeflow social media resources.
[ ] -         Kubeflow LinkedIn
[ ] -         Kubeflow X.
[ ] -         Kubeflow Bluesky.
[ ] -         Kubeflow Community YouTube channel.
[ ] -         Kubeflow YouTube channel.

chasecadet:
@andreyvelich @franciscojavierarceo are you the folks with the power to do these things?

andreyvelich:
@chasecadet You should be able to update a lot of things by yourself. Check this as a reference which OWNER files should be updated: https://github.com/kubeflow/community/issues/812

franciscojavierarceo:
yeah you can look at my previous issue here: https://github.com/kubeflow/community/issues/812

And I linked all of the PRs

kfaseela:
@andreyvelich : Would be nice to see this wrapped up, to ensure that the governance listed in the docs are really enforced in practice.

[...]

PR #1010, opened 2026-07-01
chore: Update the KSC onboarding template

christian-heusel:
- Add the 1password account that was mentioned in https://github.com/kubeflow/community/issues/950
- Add the bit.ly account


committee-steering/README.md
```diff
@@ -75,6 +75,8 @@ GitHub issue content:
       where Prow is running.
 - [ ] Kubeflow [Google Group](https://groups.google.com/g/kubeflow-discuss).
 - [ ] Update members for [KSC Google Group](https://groups.google.com/a/kubeflow.org/g/ksc).
+- [ ] Access to Kubeflow 1password account.
+- [ ] Access to Kubeflow bit.ly account.
```
andreyvelich, line 79: Actually, we can remove it since we are no longer manage Zoom password via 1password.
I am not sure about bit.ly account too.

